PR #1, opened 2016-09-30
feature(models): first model prototype for events

thnkloud9:


vied12:
what about our travis test ?! ;)

vied12:
Seems to be green on master. Well tried !

PR #5, opened 2016-10-10
Add a form to create an event

vied12:


client/controllers/PlanningModalController.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { Planning } from '../components';
+
+PlanningModalController.$inject = ['$element', 'api'];
+export function PlanningModalController($element, api) {
```
vied12: but how can access them from the component ?
Passing them by the props may lead to a bigger mess because I may need to save with the API at different levels and fields.
Regarding redux, I feel this is a little bit oversized for this component. What do you think?

vied12:
Replaced by #10

PR #9, opened 2016-10-18
feature(server): events, locations, and planning models

thnkloud9:
Implementing schema only based on:

IPTC-G2-Implementation_Guide version 2.21
(https://www.iptc.org/std/NewsML-G2/latest/IPTC-G2-Implementation_Guide)
Section 15 (EventsML-G2) - 16 (Editorial Planning)

What is included in this update, and please review for:

- All necessary fields are represented, and using appropriate field types for Planning Items,  which subsequently include coverage sets and coverage items (Locations and Events were added in previous PRs).  
- Any additional internal fields (like guid, unique_name, etc) are present.
- Behave test for basic GETs and POSTs

What is NOT included in this update:

- GUID generate functions (this will come in a separate PR later)
- Detailed behave testing for GETs, POSTs, PUTs, UPDATEs
- Predefined lookup values
- Additional validators for specific field values (like occurance_statuses and organizer_rols, etc)

### DIFF
--- a/server/planning/planning.py
+++ b/server/planning/planning.py
@@ -9,7 +9,7 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 import superdesk
-from superdesk.metadata.utils import item_url
+from superdesk.metadata.utils import generate_guid
 
 not_analyzed = {'type': 'string', 'index': 'not_analyzed'}
 
@@ -71,10 +71,26 @@ planning_schema = {
     # Planning Details
     # NewsML-G2 Event properties See IPTC-G2-Implementation_Guide 16.5.1
     # probably can skip this subsection, although its documented in iptc impl guide this way
-    'planning_details': {
+    'planning_item': {
         'type': 'dict',
         'schema': {
             # TODO: fill in from NewsML-G2 version 2.21 spec
+            'item_meta': {
+                'type': 'dict',
+                'schema': {
+                    'item_class': 'plinat:newscoverage'
+                }
+            },
+            'content_meta': {
+                'type': 'dict',
+                'schema': {
+                }
+            },
+            'news_coverage_set': {
+                'type': 'dict',
+                'schema': {
+                }
+            }
         }
     }  # end planning_details
 }  # end planning_schema
@@ -91,7 +107,6 @@ class PlanningResource(superdesk.Resource):
     resource_methods = ['GET', 'POST']
     item_methods = ['GET', 'PATCH', 'PUT', 'DELETE']
     public_methods = ['GET']
-    item_url = item_url
     privileges = {'POST': 'planning',
                   'PATCH': 'planning',
                   'DELETE': 'planning'}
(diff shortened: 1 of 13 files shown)


server/planning/planning.py
```diff
@@ -69,13 +80,158 @@ class PlanningService(superdesk.Service):
...
+    # NewsML-G2 Event properties See IPTC-G2-Implementation_Guide 16
     # probably can skip this subsection, although its documented in iptc impl guide this way
-    'planning_details': {
+    'planning_item': {
         'type': 'dict',
         'schema': {
-            # TODO: fill in from NewsML-G2 version 2.21 spec
-        }
+            # See IPTC-G2-Implementation_Guide 16.1
+            'item_meta': {
```
petrjasek: imo it's easier to work with flat structure for both item meta and content meta, when you do some filtering etc. so you could do search for `headline:foo` rather than `item_meta.headline:foo` etc.

server/planning/planning.py
```diff
@@ -69,13 +80,158 @@ class PlanningService(superdesk.Service):
...
+            # See IPTC-G2-Implementation_Guide 16.2
+            'content_meta': {
+                'type': 'dict',
+                'schema': {
+                    'qcode': {'type': 'string'},
+                    'name': {'type': 'string'}
+                }
+            },
+            # See IPTC-G2-Implementation_Guide 16.3
+            'news_coverage_set': {
```
petrjasek: imo this should be an entity `coverage` with `planning_item` reference, there will be lists of coverages filtered by users etc to see assignments, if it's embedded it will be harder to work with it imo

server/planning/planning.py
thnkloud9: the way I understand how this works is that these are the values that will be defaulted for any items/content added as coverage to this planning item, right?  so here you think we should copy the newsml related fields we have currently in superdesk-core/superdesk/metadata/item.py, yeah?  where it makes sense, of course.

server/planning/planning.py
thnkloud9: yeah, @vied12 were discussing this yesterday and I thought maybe it was better nested because I couldn't think of a case where we would want the coverage WITHOUT the planning information (especially since the item/content metadata is on the planning item, not the coverage), and thought this way saves a lookup in most cases.  what do you think?

thnkloud9: but, mostly I nested it simply because the IPTC implementation guide had it that way to be completely honest.  I'm easily swayed on this one.

petrjasek: yep I agree there wouldn't be coverages without planning most probably, question is if you would list somewhere coverages from different planning items by some other criteria, like all coverages for a desk or user, you would have to first filter out planning items, and then for each do the filtering of coverages, and then somehow put those together. I would rather do this in db

thnkloud9: yep, gotcha, I'll move out coverages to its own resource.  thanks for the input!

PR #10, opened 2016-10-25
Redux in frontend and first Events views

vied12:


thnkloud9:
Sorry about this, but I simplified the event schema a bit here:  

https://github.com/superdesk/superdesk-planning/pull/9

event_details nesting has been removed

PR #11, opened 2016-10-27
use of thunkMiddleware.withExtraArgument to inject ng services in actions

vied12:
poke @petrjasek : a better way IMO to use ng services in my actions


petrjasek:
:+1: nice

PR #17, opened 2016-11-01
feature(client): add google maps and places api and location autocomplete

thnkloud9:
Still need to

- [x] resolve issue importing angular config (for google apikey)
- [x] add action for saveLocation before saveEvent
- [x] add form validation for location
- [x] add unit tests


ioanpocol:
Superdesk already has a controlled vocabulary with locations/places. I think you have to use that list here.

thnkloud9:
@ioanpocol , this will be for planning component where we do not want to limit location to controlled vocabulary.

ioanpocol:
Please ask Karel about this, just to be sure about your asumptions.

client/styles/index.scss
```diff
@@ -3,6 +3,7 @@
 @import './events-list.scss';
 @import './add-event-container.scss';
 @import './add-event-form.scss';
+@import './react-geosuggest.scss';
```
vied12: you should import this where you are importing the geosuggest module, in `AddGeoSuggestInput.jsx` (as I did with the timepicker)

thnkloud9: this is overrides for our implementation (this css in react-geosuggest/module is only for the example for that component).

PR #18, opened 2016-11-08
Location field: load an object as value

vied12:
Give only the name to <Geosuggest/> and save the

object as the new field value


thnkloud9:
manually including these changes with lint fixes in:

https://github.com/superdesk/superdesk-planning/pull/17

PR #19, opened 2016-11-15
Repeat an event

vied12:
- [x] Repeat an event form and tests
- [x] `events` endpoint modified to create events based on `recurring_rule` and tests
- [ ] Need to take into account time zone on events generation
- [ ] Need to check if `unique_name` doesn't exist before creating an event. Because several events can actually share the same `unique_name` if they are generated from the same recurring event.


karma.conf.js
```diff
@@ -40,7 +40,7 @@ module.exports = function(config) {
         autoWatch: true,
         // start these browsers
         // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-        browsers: ['Chrome', 'Firefox'],
+        browsers: ['Firefox'],
```
vied12: this concerns tests you run locally because Travis already runs them only with Firefox.
I did that in order to save some of my memory. Because usually, after 15min, the command exits with an error related to memory.
Do you prefer to keep it ? I could add a parameter to the command on my side if you prefer.

thnkloud9: no problem for me.

client/components/AddEventForm.jsx
```diff
@@ -60,14 +80,28 @@ export class Component extends React.Component {
...
+                           withTime={true}/>
+                </div>
+                <div>
+                    <label htmlFor="repeat">Repeat ...</label>
+                    <input
+                        name="doesRepeat"
+                        type="checkbox"
+                        value={true}
+                        checked={this.state.doesRepeat}
+                        onChange={this.handleDoesRepeatChange.bind(this)}/>
```
vied12: I noticed that half of React developers bind methods in constructor but I personally prefer to bind them where I use them. Like that renaming is easier, definitions are less ambiguous and constructor keeps its role of "state" setter.
Why do you prefer to have bindings in constructor ?
I would prefer to allow arrow function syntax for class methods

PR #51, opened 2017-01-11
WIP: ICS File Ingest

thnkloud9:
Adds initial server file ingest service and ics file parser

parsing of recurring_rules not complete, missing ex_rule and ex_date processing.


server/planning/__init__.py
```diff
@@ -16,6 +16,10 @@
 from .coverage import CoverageResource, CoverageService
 from .locations import LocationsResource, LocationsService
 
+# import feed parsers and services
+from .feed_parsers.ics_2_0 import IcsTwoFeedParser  # NOQA
+from .feeding_services.event_file_service import EventFileFeedingService  # NOQA
```
vied12: I know this is an ordinary thing in SD, but I would avoid to use `# NOQA` there, and would register them directly here, in `init_app`. Just saying, I will do it myself if you agree

thnkloud9: yeah, thats cool.

PR #56, opened 2017-01-17
__init__.py added to planning.feeding_services

vied12:
it seems that `planning.feeding_services` is not a module in the test instance.
Let's see if a `__init__.py` file helps there


vied12:
It helps... WHY?!

thnkloud9:
ha, nice

PR #72, opened 2017-02-15
WIP: Advanced search

thnkloud9:
basic functionality and style. still no tests though.


thnkloud9:
getting a complexity error:

/Users/marklewis/Sites/superdesk-planning/client/reducers/planning.js
  14:18  error  Function 'anonymous' has a complexity of 12  complexity

which does not make sense to me, its just a reducer function.

client/actions/planning.js
```diff
@@ -249,6 +249,14 @@ const closePlanningEditor = () => (
     { type: 'CLOSE_PLANNING_EDITOR' }
 )
 
+const openAdvancedSearch = () => (
```
vied12: This is the event's search here, isn't it ? I would move this to the event actions module then.

client/containers/AdvancedSearchPanelContainer/style.scss
```diff
@@ -0,0 +1,25 @@
...
+    width: 0;
+    background-color: white;
+
+    .subnav {
+        .close { padding: 12px 12px 0 0; }
+        background-color: white;
+    }
+}
+
+.Planning--advanced-search-view { // when the planning editor is open
```
vied12: the comment could be adapted :)

client/containers/EventsListPanelContainer/index.js
```diff
@@ -39,14 +40,30 @@ class EventsListPanel extends React.Component {
         this.setState({ searchInputValue: event.target.value })
     }
 
+    toggleAdvancedSearch() {
```
vied12: Nothing wrong with this. I just share how I would have done this.
1. Add a `toggleAdvSearch` in `mapStateToProps`
2. Implement this toggle in it, without touching the component state but by using the value directly from the state
3. Call this `toggleAdvSearch` property on the icon `onClick`

This is not better, it's just another way that prevents to create a component state.

client/styles/index.scss
```diff
@@ -24,7 +24,9 @@ $EVENTS_LIST_WIDTH: 500px;
 }
 .Planning__events-list {
     position: relative;
-    width: 0;
+    display: flex;
+    flex-grow: 2;
+    /* width: 0; */
```
vied12: missing deleted comment

client/reducers/planning.js
```diff
@@ -43,6 +43,10 @@ const planning = (state={}, action) => {
             }
         case 'CLOSE_PLANNING_EDITOR':
             return { ...state, editorOpened: false, currentPlanningId: null }
+        case 'OPEN_ADVANCED_SEARCH':
```
vied12: This fails because the ESLint code complexity doesn't like `switch ... case`.
You can either
. move the `RECEIVE_PLANNINGS` case implementation in a function
. increment the tolerated code complexity index
. convert the `switch case` with `if` statements

vied12:
replaced by #87

PR #95, opened 2017-03-09
[SDESK-88] Applying privilege while registering superdesk-planning item and activities

nrvikas:
Depends on https://github.com/superdesk/superdesk-client-core/pull/1406


vied12:
Hi
you should rebase your branch with master to make the travis test pass.
We got a problem with an update of angular-mocks. An older version has been set.

PR #102, opened 2017-03-14
[SDESK-81] Add external links to events and display the link-page's  title to an event

nrvikas:


client/actions/events.js
```diff
@@ -56,6 +56,13 @@ export function uploadFilesAndSaveEvent(newEvent) {
     newEvent = cloneDeep(newEvent) || {}
     return (dispatch) => (
         Promise.resolve((() => {
+            // Validate event links
+            if (newEvent.links && newEvent.links.length > 0) {
+                newEvent.links = newEvent.links.filter(
+                    (l) => (l && l !== 'Paste link')
```
vied12: Why not using a `placeholder` attribute for this purpose? then you wouldn't have to make this check.
Also this would risky to use an hard coded string here. If the text is changed in the other file, this instruction won't be relevant anymore.

client/components/fields/LinkField.jsx
```diff
@@ -0,0 +1,81 @@
...
+            this.setState({title: null})
+            return
+        }
+
+        if (!this.props.iframelyKey) {
+            this.setState({title: this.errorTitle})
+            return
+        }
+
+        request({
```
vied12: You are using `axios` without requiring it in `packages.json`.
I didn't know this library. I was using [Fetch with polyfill](https://www.npmjs.com/package/whatwg-fetch), a standard in recent browsers. Do you know it? Do you prefer `axios` ?

client/components/fields/LinkField.jsx
```diff
@@ -0,0 +1,81 @@
...
+        this.errorTitle = 'Could not load title'
+        this.setTitle(this.props.link)
+    }
+
+    componentWillReceiveProps(nextProps) {
+        this.setTitle(nextProps.link)
+    }
+
+    setTitle(link) {
+        if (!link || link === 'Paste link') {
```
vied12: Same here. Prefer to use `placeholder` attribute, or please don't hard code the message.

client/actions/events.js
nrvikas: Absolutely. Address it.

client/components/fields/LinkField.jsx
nrvikas: I did find a few advantages in the documentation of 'axios' over 'fetch polyfill'. On the other hand, fetch is a standard which is advantageous. I will be happy to go with your suggestion. Addressed it.

PR #106, opened 2017-03-16
[SDESK-850] Added time__short class to event item's time field as it can be visible when advanced search pane is opened

nrvikas:
Depends on client-core: https://github.com/superdesk/superdesk-client-core/pull/1426

### DIFF
--- a/client/components/EventItem/index.jsx
+++ b/client/components/EventItem/index.jsx
@@ -19,7 +19,7 @@ export const EventItem = ({ event, onClick }) => {
             <div className="line">
                 <span className="keyword">{event.name}</span>
                 <span className="item-heading">{event.definition_short}</span>
-                <time title={time}>{time}</time>
+                <time className="time__short" title={time}>{time}</time>
             </div>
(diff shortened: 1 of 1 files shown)


client/components/EventItem/index.jsx
```diff
@@ -19,7 +19,7 @@ export const EventItem = ({ event, onClick }) => {
             <div className="line">
                 <span className="keyword">{event.name}</span>
                 <span className="item-heading">{event.definition_short}</span>
-                <time title={time}>{time}</time>
```
vied12: With BEM, when we define an element state, we use `--` instead of `__`.
I made this error few times already, you can blame me when you find one in the project ;)

http://getbem.com/introduction/

nrvikas:
Closing this PR as it has multiple errors in it:
1) It references a wrong branch and ticket SDESK-850 instead of the correct one SDESK-896.
2) It has review comments which are address in the new PR with the right ticket name/branch in it.
Apologies.

PR #115, opened 2017-03-20
fix(planning list): Fix the order on planning items list

ioanpocol:


client/selectors/index.js
```diff
@@ -28,7 +28,7 @@ export const getCurrentAgendaPlannings = createSelector(
                 (pid) => (storedPlanningsObjects[pid])
             )
             .filter((d) => d !== undefined), // remove undefined
-        [(p) => (-p._created)]) // sort by new created first, or by name
+        [(p) => -(new Date(p._created))]) // sort by new created first, or by name
```
vied12: I should have checked that before, but the right way to sort a list with lodash is with `orderBy` instead of `'sortBy`, because you can add `['desc']` as the last parameter.

ioanpocol: updated to use orderBy

PR #116, opened 2017-03-20
Default values event form SDESK-100

vied12:


client/containers/AddEventContainer.jsx
```diff
@@ -4,6 +4,13 @@ import { connect } from 'react-redux'
 import { EventForm } from './index'
 import * as actions from '../actions'
 
+const DEFAULT_VALUES = {
```
vied12: It's a mistake, this whole file will be removed in another PR

PR #119, opened 2017-03-21
fix(events): Show events that started before current date&time but ar…

ioanpocol:
…e not finished yet


ioanpocol:
Currently when you set a start date is set also the end date and you are not able to delete end date.

vied12:
You're right, even ingested events have a end date. :ok_hand:

PR #120, opened 2017-03-21
Resizable panel

thnkloud9:
implements a resizable panel component using existing jquery ui feature.

currently events list panel hide does not work with resized panels.  I need some help figuring out the best way to implement it back, possible animating removal of the EventsPanelContainer component would be the best option?


thnkloud9:
Missing tests, but maybe we should merge so we can then merge lint rules PR first?

PR #134, opened 2017-03-29
fix(event): Small fixes and improvements on add/edit Event

ioanpocol:


client/containers/EventForm/index.jsx
```diff
@@ -65,6 +71,7 @@ export class Component extends React.Component {
                         </div>
                     </div>
                     <span className="subnav__page-title">Event Details</span>
+                    <button type="submit" className="btn" onClick={onBackClick}>Cancel</button>
```
vied12: I don't think this cancel button should be typed `submit`

server/planning/events.py
```diff
@@ -435,3 +441,14 @@ def generate_recurring_dates(start, frequency, interval=1, until=None, byday=Non
         return (tz.localize(dt).astimezone(pytz.UTC).replace(tzinfo=None) for dt in dates)
     else:
         return (date for date in dates)
+
+
+def setRecurringMode(event):
```
vied12: This cannot be done on the client side ? We try to handle data as much as we can on the client side

ioanpocol: I added here because this is an eve issue, when you update a dictionary actually eve is doing an merge.
Because of this we have to force some fields to be None in order to reset the current value.
Can be moved on client but I think it will be a little confusing for a front-end developer.

vied12: Are you talking about the fact that the client perform a PATCH when the object contains a `_links` and the request will contain only fields with a value set ? Then up to now, in order to reset a field we give the value `null` to this field. On the client side, because this is the client's `api` service that behaves this way.

ioanpocol: For the first level fields I agree that you have to reset from the client.
But on eve if you patch a field of type dictionary you expect that the old dictionary will be replaced with the new one.This is happening for example if you patch a field of type list the old one will be replaced with the new one.
On dictionary it will perform a deep update and will not replace the entire dictionary.
I think this is not a common patch behavior.

vied12: wow, no, it's not a common behavior... ok i'm fine then :ok_hand:

ioanpocol:
Added a few code comments

server/start_behave.py
```diff
@@ -0,0 +1,11 @@
+#!/home/ioan/sd/env/bin/python3.4
```
vied12: I don't think I have the same env as you :)

PR #136, opened 2017-03-31
[SDESK-97] Delete event feature

nrvikas:


client/containers/DeleteEventContainer.jsx
```diff
@@ -0,0 +1,37 @@
...
+import { ModalWithForm } from '../components'
+import { connect } from 'react-redux'
+import * as actions from '../actions'
+import * as selectors from '../selectors'
+
+/**
+* Modal for adding an agenda
+*/
+export const DeleteEvent = ({ show, handleHide, eventDetail }) => (
+    <ModalWithForm
```
vied12: Why do we need a `<ModalWithForm/>` ? I don't see any field to fill in this modal.
It seems more to be a simple modal to me

client/containers/DeleteEventForm.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+Component.propTypes = {
+    handleSubmit: React.PropTypes.func.isRequired,
+    handlePlanningClick: React.PropTypes.func.isRequired,
+    initialValues: React.PropTypes.object.isRequired,
+    error: React.PropTypes.object,
+}
+
+// Decorate the form component
+export const FormComponent = reduxForm({
```
vied12: Again I don't see the point to use reduxForm here

client/containers/EventsPanelContainer/index.jsx
```diff
@@ -29,9 +29,10 @@ EventsPanel.propTypes = {
 
 const mapStateToProps = (state) => ({
     // selectors.getShowEventDetails is either true, an event id, or null
-    eventToDetail: selectors.getShowEventDetails(state) === true
-        ? DEFAULT_FORM_VALUES
-        : selectors.getEventToBeDetailed(state),
+    eventToDetail: selectors.getShowEventForm(state) ? selectors.getShowEventDetails(state) === true
```
vied12: what means "EventForm" in `getShowEventForm` ? Is it this form in the deletion modal ?
This logic is a bit unclear to me, maybe because there are two ternary operations in the same instruction :)

client/reducers/events.js
```diff
@@ -67,11 +68,19 @@ const eventsReducer = (state=initialState, action) => {
...
         case 'CLOSE_EVENT_DETAILS':
             return {
                 ...state,
                 showEventDetails: null,
+                showEventForm: false,
+            }
+        case 'DELETE_EVENT':
+            return {
+                ...state,
+                showEventDetails: action.payload,
```
vied12: why updating `showEventDetails` and `showEventForm` here ? I find a little bit confusing `showEventDetails` and `showEventForm`. Maybe one of them can be removed or renamed because the names suggest to me the same thing.

server/planning/events.py
```diff
@@ -106,6 +107,22 @@ def on_create(self, docs):
         if generatedEvents:
             docs.extend(generatedEvents)
 
+    def on_delete(self, doc):
```
vied12: :cool: 

a small behave test would be even cooler :)

client/containers/DeleteEventContainer.jsx
nrvikas: Yes, I see your point. I will make change that.

client/containers/DeleteEventForm.jsx
nrvikas: I'm using redux-form to display any errors in-case of submission-errors. Would you prefer it otherwise ?

client/containers/EventsPanelContainer/index.jsx
nrvikas: When we set the showEventDetail to an event-detail object / true value, it opens the Event form for editing as it is mapped to the EventForm's props. I need the event-detail object for delete modal because I need _plannings. But, I don't want the event form to open. Hence added showEventForm to set the intention that Event form should be open. So event form can depend on that to open itself instead of showEventDetail.

client/reducers/events.js
nrvikas: I need the event object with detail _planning, but I do not want the event form to be open. Hence the two options.
Please suggest if you would prefer any other approach.

server/planning/events.py
nrvikas: Absoultely. I will add a behave test case for this.

client/containers/EventsPanelContainer/index.jsx
vied12: I see you are using `eventToDetail` to know which event is asked for deletion ?
I think you could leave `eventToDetail` for what it has been made for, and instead pass the event as a parameter to the action: `dispatch(showModal({ modalType: 'DELETE_EVENT' , modalProps: event }))`

client/containers/DeleteEventForm.jsx
vied12: Yes I think we should't use reduxForm for this only usage but only when we have an actual form. Instead you could intercept errors in the component where you are calling the deletion method.

client/containers/EventsPanelContainer/index.jsx
nrvikas: Hmmm.. yes, that was my first approach. But the event has no details about "_plannings" in the event object. "_plannings" are required to show the related plannings of an event to delete it. You can take a look at the test instance (http://sdppr-136.test.superdesk.org/), by adding an event to an agenda and deleting the event. I will give it another look to see if I'm missing anything and try not to touch showEventDetail :)

client/containers/DeleteEventForm.jsx
vied12: For another ticket I made a Confirmation modal component here: https://github.com/superdesk/superdesk-planning/pull/138

You could use it :)

PR #138, opened 2017-03-31
Ask a confirmation for planning deletion

vied12:
This adds a new ConfirmationModalContainer, which can be used like that:

```
dispatch(actions.showModal({
            modalType: 'CONFIRMATION',
            modalProps: {
                body: `Are you sure you want to delete this planning?`,
                action: () => dispatch(actions.deletePlanning(planning)),
            },
}))
```

SDESK-985

### DIFF
--- a/client/containers/ConfirmationModalContainer/index.jsx
+++ b/client/containers/ConfirmationModalContainer/index.jsx
@@ -1,0 +1,59 @@
+import React from 'react'
+import { Modal, Button } from 'react-bootstrap'
+import { connect } from 'react-redux'
+import * as actions from '../../actions'
+
+export function ConfirmationModalComponent({ show, title, body, handleHide, handleAction }) {
+    const action = () => (
+        Promise.resolve(handleAction())
+        .then(handleHide)
+    )
+    return (
+        <Modal show={show} onHide={handleHide}>
+            <Modal.Header>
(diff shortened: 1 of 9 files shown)


client/containers/ConfirmationModalContainer/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+    return (
+        <Modal show={show} onHide={handleHide}>
+            <Modal.Header>
+                <a className="close" onClick={handleHide}>
+                    <i className="icon-close-small" />
+                </a>
+                <h3>{ title }</h3>
+            </Modal.Header>
+            <Modal.Body>
+                { body }
```
nrvikas: Are we planning to have only text displayed in the modal body? Can we use a ReactElement as body object as well ? Because we might want to use some React components in the confirmation. Example would be to use RelatedPlanning component while deleting events.

vied12: body can be a react element, check the proptypes:
```
    body: React.PropTypes.oneOfType([
+        React.PropTypes.string,
+        React.PropTypes.element,
+    ]),
```

nrvikas: Sorry, I missed it. Tks.

PR #154, opened 2017-04-07
fix(events): Update recurring events

ioanpocol:


server/planning/events.py
```diff
@@ -126,6 +126,63 @@ def on_delete(self, doc):
...
+        if 'change_mode' in updates['dates'].get('recurring_rule', {}):
+            del updates['dates']['recurring_rule']['change_mode']
+        if changeMode == 'simple':
+            # this is an recursive update(see below)
+            return
+        elif changeMode == 'current':
+            # update only the current item so set it as a non recurrent task
+            updates['dates']['recurring_rule'] = {}
+            updates['recurrence_id'] = None
+        else:
```
ioanpocol: Initially was the option to change current, all and all next. I drop the option all and I keep only the current and all (all next).

PR #164, opened 2017-04-18
[SDESK-1129] Modify RelatedPlannings component to get agendas from store instead of props

nrvikas:


client/containers/RemoveAgendaConfirmation.jsx
```diff
@@ -29,7 +29,10 @@ RemoveAgendaConfirmationComponent.propTypes = {
 const mapStateToProps = (state, ownProps) => ({
     plannings: Object.keys(selectors.getStoredPlannings(state))
         .filter((pKey) => (get(ownProps.agenda, 'planning_items', []).indexOf(pKey) > -1))
-        .map((pKey) => (selectors.getStoredPlannings(state)[pKey])),
+        .map((pKey) => {
+            return extend({}, selectors.getStoredPlannings(state)[pKey],
```
vied12: Why not using the spread operator instead ? 
`{ ...selectors.getStoredPlannings(state)[pKey], _agenda: { name: ownProps.agenda.name }}`

nrvikas:
@vied12 , I like your idea of converting RelatedPlannings to a container ( as it can independently get the related information from the store and it is used in more than a couple of places in the module ).
Thanks for that. I will modify the PR.

vied12:
Please rebase with master to make the tests to pass

PR #173, opened 2017-04-21
[SDESK-1105] Time-picker component 

nrvikas:


client/components/fields/DayPickerInput/index.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react'
 import DatePicker from 'react-datepicker'
-import TimePicker from 'rc-time-picker'
+import { TimePicker } from '../TimePicker'
 import moment from 'moment'
 import 'react-datepicker/dist/react-datepicker.css'
 import 'rc-time-picker/assets/index.css'
```
vied12: this can be removed too, like the ref to rc-time-picker in the packages.json file

client/components/fields/TimePicker/styles.scss
```diff
@@ -0,0 +1,182 @@
+@import '~superdesk-core/styles/sass/mixins.scss';
+@import '~superdesk-core/styles/sass/variables.scss';
+
+// .timepicer-input and timepicker-popup classes are a part of superdesk-client-core
```
vied12: Any idea how we could reuse this style @fritzSF ?

PR #177, opened 2017-04-24
[SDESK-1127] Agenda name should be unique.

mdhaman:


server/planning/planning.py
```diff
@@ -55,6 +61,29 @@ def on_deleted(self, doc):
         if 'planning_items' in doc:
             self.delete({'_id': {'$in': doc['planning_items']}})
 
+    def _validate_unique_agenda(self, updates, original):
```
vied12: `'unique': True` in the planning schema wouldn't fit our need ?

client/actions/planning.js
```diff
@@ -4,18 +4,26 @@ import * as actions from '../actions'
...
             planning_type: 'agenda',
             name: name,
         })
         .then((agenda) => {
-            notify.success('An agenda has been added')
+            notify.success('An Agenda has been added.')
             dispatch(hideModal())
             dispatch(addOrReplaceAgenda(agenda))
             dispatch(selectAgenda(agenda._id))
+        }, (error) => {
```
vied12: We used to raise a `SubmissionError` in order to show the error in the form.
It's part of redux-form, you can have a look there: http://redux-form.com/6.6.3/examples/submitValidation/

server/planning/planning.py
mdhaman: `unique` is case-sensitve.  Most of the entities name in Superdesk are case-insensitive. Hence did not use `unique` validator. 

I could have used `iunique` (case-insensitive one) but we are storing planning item in the same collection. We are not currently using `name` for planning item but if start using it or use name field in a different way then we would have the write our own validation. 

I have done a separate PR https://github.com/superdesk/superdesk-core/pull/899 as well so that you can use the `iunique` validator if search backend is elastic.

client/actions/planning.js
mdhaman: will look into this.

server/planning/planning.py
vied12: Make sense ! Thanks for the explanation

mdhaman:
@vied12, @thnkloud9 @nrvikas please merge these changes. I don't have merge rights.

PR #178, opened 2017-04-24
Adds python coverage

vied12:


server/planning/feeding_services/event_http_service.py
```diff
@@ -87,7 +87,7 @@ def _update(self, provider, update):
         logger.info('Ingesting: %s', str(response.content))
 
         if isinstance(parser, NTBEventXMLFeedParser):
-            xml = ElementTree.parse(response.content)
+            xml = ET.fromstring(response.content)
```
vied12: @thnkloud9 I assumed here that content is a text, instead of a file because I don't understand how this could be a file. And mainly because it's more convenient with the new test ^^

PR #185, opened 2017-04-28
Search plannings SDESK-23

vied12:
New Searchbar component added

Search in planning is done client side in order to benefit from having all the plannings loaded into the store


mdhaman:
@vied12  why search is done only on the client? Why elasticsearch is not used?

vied12:
This is a simple search component to filter the list of planning.
Because we already have all of them in the browser, I find it more efficient to filter them locally.
Beside, I am also searching in the related event properties. I don't know if the current ES Superdesk integration allows that.

mdhaman:
We would have lot of planning items. For example, we can have lots events/planning items per calendar year for Sports. I am not sure if this can be filtered in browser efficiently.

vied12:
@mdhaman Good point. I will run some tests with many planning and if it goes slow, I will create a local index, maybe with lunr.js.

mdhaman:
@vied12 maybe worth investigating parent-child relationship in elasticsearch.

petrjasek:
I guess for some basic filtering it's ok, if there is no need for relevance sorting and number of items is kind of limited (under ~1k). when working on dictionary words search I had to implement trie there to get some acceptable performance for large sets, but it is doing prefix search so it was easy to optimize.

hlmnrmr:
Further revision and optimization will be done as part of https://dev.sourcefabric.org/browse/SDESK-1294

PR #190, opened 2017-05-02
[SDESK-1172] Repeat summary for recurring events

nrvikas:


nrvikas:
@vied12 , thanks for your feedback. I have updated the PR with a new component.
Thanks.

PR #198, opened 2017-05-09
[SDESK-1210] - Privileges for Planning Item Management

MarkLark86:


server/planning/planning.py
```diff
@@ -252,6 +252,6 @@ class PlanningResource(superdesk.Resource):
     resource_methods = ['GET', 'POST']
     item_methods = ['GET', 'PATCH', 'PUT', 'DELETE']
     public_methods = ['GET']
-    privileges = {'POST': 'planning',
-                  'PATCH': 'planning',
+    privileges = {'POST': 'planning_planning_management',
+                  'PATCH': 'planning_planning_management',
                   'DELETE': 'planning'}
```
ioanpocol: Why for DELETE the privilege is planning and not planning_planning_management?

MarkLark86: Because the delete will be changed to a spike functionality in a coming pull request for event, planning and agenda items.

client/actions/planning.js
```diff
@@ -221,15 +230,20 @@ const openPlanningEditor = (planning) => ({
...
  */
 const openPlanningEditorAndAgenda = (planning) => (
-    (dispatch, getState) => {
-        const agenda = selectors.getAgendas(getState()).find(
-            (a) => (a.planning_items || []).indexOf(planning) > -1
-        )
-        if (agenda && agenda._id !== selectors.getCurrentAgendaId(getState())) {
-            dispatch(selectAgenda(agenda._id))
+    (dispatch, getState, { notify, $timeout }) => {
+        if (selectors.getPrivileges(getState()).planning_planning_management === 1) {
```
vied12: Since we are going to check privileges in many actions, maybe we could create a function that check the permission which we would use to wrap the actual actions when we export them ?
```js
export const openPlanningEditorAndAgenda = checkPermission('planning_planning_management', _openPlanningEditorAndAgenda)
```
This is just a suggestion but in this way we would avoid this extra condition and logic in many actions. Also we would be able to see the relationship between actions and permissions in a more clear way, all in the same part of code.
what do you think ?

server/planning/planning.py
mdhaman: Spike should be `PATCH` method not `DELETE`. You are changing the state not deleting an item. 
Also, there is no need for `DELETE` in `item_methods`.

client/actions/planning.js
MarkLark86: Very good suggestion, thank you.

I will make those changes this morning.

MarkLark86:
I made a mistake and this commit was merged along with PR #198, so we can safely close this PR.

PR #199, opened 2017-05-09
feat(agenda): Add the option to rename agenda

ioanpocol:


client/actions/agenda.js
```diff
@@ -35,6 +35,42 @@ const createAgenda = ({ name }) => (
 )
 
 /**
+ * Action Dispatcher for updating an Agenda
+ * @param {string} _id - The id of the Agenda
+ * @param {string} name - The new name of the Agenda
+ * @return arrow function
+ */
+const updateAgenda = ({ _id, name }) => (
```
vied12: `updateAgenda` definition looks very similar to `createAgenda`. I would factorize them in order to avoid repetition in the triggered actions. Because in the future if we have to change them somewhere, we would probably have to do it also the other one.

client/components/AgendaModal.jsx
```diff
@@ -0,0 +1,34 @@
...
+            <ModalWithForm
+                title="Edit an Agenda"
+                onHide={handleHide}
+                form={CreateEditAgendaForm}
+                show={true}
+                initialValues={agenda}/>
+        )
+    } else {
+        return (
+            <ModalWithForm
```
vied12: Again, there is maybe no need to duplicate this component instantiation. We can maybe just ues the condition for the `title` and `initialValues` props.

PR #210, opened 2017-05-19
[SDESK-1236] - Spike/Unspike Privileges for Events

MarkLark86:


server/planning/planning.py
```diff
@@ -47,11 +48,6 @@ def on_create(self, docs):
         for doc in docs:
             doc['guid'] = generate_guid(type=GUID_NEWSML)
             set_original_creator(doc)
-            # remove event expiry if it is linked to the planning
```
vied12: How do we ensure an event won't be removed if it get linked to a planning ?

MarkLark86: Event, Agenda and Planning items should only be removed from the system when they're marked as spiked and purged from the system, based on a configurable number of minutes from on the **PLANNING_EXPIRY_MINUTES** system setting. Please see [SDESK-1224](https://dev.sourcefabric.org/browse/SDESK-1224) for futher information

There is logic in place to ensure that if an Event is spiked that all associated Planning items are marked as spiked as well, which the end user is informed about in the SpikeEvent modal component.

vied12:
But by now I think they are purged when an expiry date is set and every
ingested event get one. This is a global rule for every ingested item in
Superdesk. This prevent the system to be overloaded with too many unused
ingested item. Without these lines, an ingested event can be purged even if
it is linked to a planning.

PR #212, opened 2017-05-24
[SDESK-1284] Assignments for coverages

nrvikas:
Reference UI design:
https://wiki.sourcefabric.org/display/NR/Planning+interface?preview=/24052602/41910540/Planning_Split-view_2.png
and
https://wiki.sourcefabric.org/display/NR/Planning+interface?preview=/24052602/41910541/Planning_Split-view_3.png

### DIFF
--- a/server/planning/coverage.py
+++ b/server/planning/coverage.py
@@ -12,9 +12,13 @@
 
 import superdesk
 import logging
+from superdesk.errors import SuperdeskApiError
 from superdesk.metadata.utils import generate_guid
 from superdesk.metadata.item import GUID_NEWSML
 from apps.archive.common import set_original_creator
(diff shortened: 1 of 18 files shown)


server/planning/coverage.py
```diff
@@ -104,7 +104,12 @@ def on_create(self, docs):
                     }
                 }
             },
-            'assigned_to': {'type': 'string'},
+            'assigned_to': {
```
mdhaman: Can we change this to list of `dict` instead of list of `string`?

mdhaman:
As requirements are not yet finalised, to start with just have coverage assigned to a `Superdesk User` or a `Superdesk Desk`.

PR #218, opened 2017-06-01
[SDESK-1392] Subscribe to Event Creation Notifications

MarkLark86:


client/utils/retryDispatch.js
```diff
@@ -0,0 +1,38 @@
...
+                maxRetries,
+                retries,
+                interval,
+            },
+        })
+        if (retries >= maxRetries) {
+            return Promise.reject({ error_msg: 'Max retries exceeded' })
+        }
+
+        return $timeout(() => (
```
vied12: I don't think there is a need to use the angular timeout function which start a digest cycle. What about using a simple setTimeout here ?

PR #231, opened 2017-06-08
[SDESK-1448] Viewmode for planning item

nrvikas:


client/components/PlanningList/index.jsx
```diff
@@ -31,7 +39,8 @@ class PlanningList extends React.Component {
                     agenda={currentAgenda}
                     onSpike={handlePlanningSpike}
                     onUnspike={handlePlanningUnspike}
-                    onClick={openPlanningEditor.bind(null, planning._id)}
+                    onClick={handleSingleAndDoubleClick.bind(null, this,
```
vied12: A more common way to trigger an action from a component would be to add two props `onSingleClick` and `onDoubleClick`to `<PlanningItem/>`

client/components/fields/TimePicker/styles.scss
```diff
@@ -20,10 +20,10 @@
         background-color: transparent;
         padding: 0;
         height: 24px;
-        &.line-input + .btn {
-            background-color: transparent;
-            top: 0px;
-        }
+        /* &.line-input + .btn {
```
vied12: please remove unused code

client/utils/handleSingleAndDoubleClick.js
```diff
@@ -0,0 +1,30 @@
+import { debounce } from 'lodash'
+
+// Due to limitations of ES6 default strict mode on all files, we cannot bind the
+// 'this' object from calling code. Hence, passing this directly as 'self' (ugly).
```
vied12: have you tried something like that ?
```js
onClick={() => handleSingleAndDoubleClick(arg)}
```
I find this function a bit confusing. Why does it call methods from self ? callback parameters wouldn't be more clear ?

client/components/PlanningList/index.jsx
nrvikas: Yup, I agree. I have changed it accordingly.

client/components/fields/TimePicker/styles.scss
nrvikas: Done.

client/utils/handleSingleAndDoubleClick.js
nrvikas: I ran into some issues trying to use bound parameters to the lodash's debounce function. But I guess I got this working fine now. I'm passing callback functions as parameters to this utility function now. React not handling onClick and onDoubleClick on the same component made me use the lodash's debounce workaround. And we thought we'll put this in utility because EventsList might use this too. If this function was a part of the component itself, it would be much clearer. However, I do understand it is quite confusing. Please let me know what you think with the new changes. Thanks :)

vied12: I understand you are using `self` to store the debounce. I've red somewhere  that modifying properties of a Component is an anti pattern since two instances of it with the same props could behave differently. But I don't know how we could do otherwise, so it's fine :) Maybe `clickedOnce` could belong to `this.state` ?
One more thing: what about  having `this.clickedOnce = undefined` here, in a wrapper function that would call `onSingleClick()` at the end ? Then all the logic of this `handleSingleAndDoubleClick` function would stay here. What do you think ?

PR #232, opened 2017-06-12
Ask the user to confirm before he updates a recurring event

vied12:


server/planning/events.py
```diff
@@ -181,9 +164,17 @@ def on_update(self, updates, original):
...
             updates['dates']['recurring_rule'] = None
-            updates['recurrence_id'] = generate_guid(type=GUID_NEWSML)
+            updates['recurrence_id'] = None
+            # we spike all the related recurrent events
+            if 'recurrence_id' in original:
+                # retieve all the related events
+                events = self.find(where={'recurrence_id': original['recurrence_id'], '_id': {'$ne': original['_id']}})
+                spike_service = get_resource_service('events_spike')
+                # spike them
+                for event in events:
```
mdhaman: We will loose data if we spike the all related recurrences of an event. Is there a need to spike recurrence of events that have already been in the past?

vied12: I can keep the past events, good idea

vied12: updated

PR #234, opened 2017-06-14
fix(client): SDESK-1342: Events list: show "today" as the first day in the list

thnkloud9:
fix is only applied to initial state of the event list without a search applied.

### DIFF
--- a/client/selectors/index.js
+++ b/client/selectors/index.js
@@ -6,6 +6,8 @@ import { ITEM_STATE } from '../constants'
 
 export const getAgendas = (state) => state.agenda.agendas
 export const getCurrentPlanningId = (state) => state.planning.currentPlanningId
+export const getCurrentSearch = (state) => state.events.search.currentSearch ?
+    state.events.search.currentSearch : {}
 export const getEvents = (state) => state.events.events
 export const isEventListShown = (state) =>state.events.show
 export const getPreviousEventRequestParams = (state) => get(state.events, 'lastRequestParams', {})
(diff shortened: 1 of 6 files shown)


client/selectors/index.js
```diff
@@ -6,6 +6,8 @@ import { ITEM_STATE } from '../constants'
 
 export const getAgendas = (state) => state.agenda.agendas
 export const getCurrentPlanningId = (state) => state.planning.currentPlanningId
+export const getCurrentSearch = (state) => state.events.search.currentSearch ?
+    state.events.search.currentSearch : {}
```
vied12: I recently discovered it's better to avoid to return default values in selector (like `{}` here) because it repeat what the reducer's initial state declares. And when it changes somewhere, there is a risk to forget to replace the default value everywhere.

client/selectors/index.js
```diff
@@ -170,15 +172,26 @@ export const getEventsWithMoreInfo = createSelector(
 * the associated events.
 */
 export const getEventsOrderedByDay = createSelector(
-    [getEventsWithMoreInfo],
-    (events) => {
+    [getCurrentSearch, getEventsWithMoreInfo],
+    (currentSearch, events) => {
         if (!events) return []
+        // check if search exists
+        let search = (currentSearch.advancedSearch || currentSearch.fulltext) ? true : false
```
vied12: I would create directly a selector for that. Like `isASearchHappening`, but in a better English ^^). If the `currentSearch` definition changes one day, then we will avoid to go inside this `getEventsOrderedByDay` function to change it.

client/selectors/index.js
```diff
@@ -170,15 +172,26 @@ export const getEventsWithMoreInfo = createSelector(
...
         events = events.sort((a, b) => a.dates.start - b.dates.start)
         var days = {}
         function addEventToDate(event, date) {
             date = date || event.dates.start
             date = date.format('YYYY-MM-DD')
+            // if not in search mode, only add dates in the future
+            if (!search) {
+                let newDate = moment(date)
+                let now = moment()
+                if (!moment(newDate).isSameOrAfter(now)) {
```
vied12: `newDate` is already a moment instance

client/selectors/index.js
```diff
@@ -6,6 +6,10 @@ import { ITEM_STATE } from '../constants'
 
 export const getAgendas = (state) => state.agenda.agendas
 export const getCurrentPlanningId = (state) => state.planning.currentPlanningId
+export const eventSearchActive = (state) => (typeof state.events.search !== 'undefined') ? ((
+    state.events.search.currentSearch.advancedSearch ||
+    state.events.search.currentSearch.fulltext) ?
+    true : false) : false
```
thnkloud9: this causes test failure at - client/components/EventsListContainer/_test.jsx:10:24 and lient/components/tests/PlanningApp_test.jsx:20:24 for "TypeError: state.events.search.currentSearch is undefined in tests.js (line 84899)", but I have not been able to resolve the issue.  Ive tried setting it in initialState passed to createTestStore, and also setting a default in utils/index.js but nothing seems to work.  Not sure how to resolve the issue.

PR #236, opened 2017-06-15
[SDESK-1447] View/Edit mode for Events

nrvikas:


client/actions/events.js
```diff
@@ -613,6 +613,16 @@ const closeAdvancedSearch = () => (
 )
 
 /**
+ * Opens the Event in preview/read-only mode
+ * @param {object} event - The Event ID to preview
+ * @return Promise
+ */
+const previewEvent = (event) => ({
+    type: EVENTS.ACTIONS.PREVIEW_EVENT,
+    payload: get(event, '_id', event || true),
```
vied12: I find a bit confusing to mix `_id` and boolean here.
Can a user preview an event that has no `_id` (not saved event) ?

client/components/AddGeoLookupInput/index.jsx
```diff
@@ -18,15 +18,17 @@ class GeoLookupInput extends React.Component {
         return (
             <Geolookup
                 placeholder='Address, City'
-                inputClassName="line-input"
-                buttonClassName='btn btn-default geolookup__button'
+                inputClassName={(this.props.readOnly ? 'disabledInput ' : '') + 'line-input'}
```
vied12: It's totally fine. I just want to let you know that I introduced a `classNames` module that help to define classes like that.
https://github.com/JedWatson/classnames
Use it only if you like :)

client/actions/events.js
nrvikas: Good catch Edouard, I have changed it.

client/components/AddGeoLookupInput/index.jsx
nrvikas: Sure, that's a better approach. Tks.

PR #239, opened 2017-06-16
Bulk operations on events

vied12:
- Allows to select one or many events
- Bulk planning creation
- Bulk event spiking


client/actions/agenda.js
```diff
@@ -134,14 +134,19 @@ const fetchAgendaById = (_id) => (
  * @param {object} agenda - The agenda to add the planning to
  * @return Promise
  */
-const _addPlanningToAgenda = ({ planning, agenda }) => (
+const _addPlanningsToAgenda = ({ plannings, planning, agenda }) => (
```
ioanpocol: Can you keep only plannings and apply the same pattern as on line 168?

client/actions/tests/agenda_test.js
```diff
@@ -349,10 +349,10 @@ describe('agenda', () => {
                     notify,
                     $timeout,
                 })
-                .then((planning) => {
-                    expect(planning).toEqual(plannings[0])
-                    // Cannot check dispatch(addPlanningToAgenda()) using a spy on dispatch
-                    // As addPlanningToAgenda is a thunk function
+                .then((plannings) => {
+                    expect(plannings).toEqual([plannings[0]])
```
ioanpocol: Is this correct?

vied12: I think so. Nothing has changed on the planning. Only the agenda has been updated.

ioanpocol: Before it was planning and plannings and now it's plannings on both sides.

PR #241, opened 2017-06-20
[SDESK]1089] Add "all day" option to event's time

MarkLark86:


client/components/Toggle/index.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import classNames from 'classnames'
+import './style.scss'
+
+export default function Toggle({ value, onChange, readOnly }) {
+    const onClick = () => onChange({ target: { value: !!value } })
```
vied12: `!!value` will cast the value to a boolean. I guess you meant `!value` to inverse the current value.

client/utils/index.js
```diff
@@ -244,3 +244,18 @@ export const getErrorMessage = (error, defaultMessage) => {
...
 }
+
+/**
+ * Helper function to determine if the starting and ending dates
+ * occupy entire day(s)
+ * @param {moment} startingDate - A moment instance for the starting date/time
+ * @param {moment} endingDate - A moment instance for the starting date/time
+ * @return {boolean} If the date/times occupy entire day(s)
+ */
+export const isEventAllDay = (startingDate, endingDate) => {
```
vied12: there is already a function `isAllDay` in this file. It would clearer to merge them

vied12:
I have this exception when I click on the toggle:
```
Uncaught TypeError: Cannot read property 'clone' of undefined
    at t.value (app.bundle.js:721)
    at r (app.bundle.js:714)
    at Object.i (app.bundle.js:607)
    at a (app.bundle.js:607)
    at Object.s [as executeDispatchesInOrder] (app.bundle.js:607)
    at f (app.bundle.js:606)
    at m (app.bundle.js:606)
    at Array.forEach (<anonymous>)
    at n (app.bundle.js:607)
    at Object.processEventQueue (app.bundle.js:607)
    at i (app.bundle.js:621)
    at Object.handleTopLevel [as _handleTopLevel] (app.bundle.js:621)
    at o (app.bundle.js:623)
    at i.perform (app.bundle.js:607)
    at Object.batchedUpdates (app.bundle.js:623)
    at Object.o [as batchedUpdates] (app.bundle.js:607)
```

client/components/Toggle/index.jsx
MarkLark86: I only renamed the file to .jsx, added readOnly state and unit tests, the remaining is the original functionality.

I can change this Toggle to send the new Toggle state if you wish?

MarkLark86:
The following error was caused by redux-form, updating redux-form fixes the issue.
```
Uncaught TypeError: Cannot read property 'clone' of undefined
```

client/components/Toggle/index.jsx
vied12: Yes I think I even did this typo myself. And I never noticed it because I wasn't using the onChange method I guess. But now you are using it, I think it makes sense to return the new value as parameter ?

vied12:
works like a charm :+1:

PR #242, opened 2017-06-20
fix(client): make future today inclusive, only remove filter on advanced search

thnkloud9:


client/selectors/index.js
```diff
@@ -6,9 +6,7 @@ import { ITEM_STATE } from '../constants'
 
 export const getAgendas = (state) => state.agenda.agendas
 export const getCurrentPlanningId = (state) => state.planning.currentPlanningId
-export const eventSearchActive = (state) => (
-    state.events.search.currentSearch.advancedSearch ||
-    state.events.search.currentSearch.fulltext) ?
+export const eventSearchActive = (state) => (state.events.search.currentSearch.advancedSearch) ?
```
vied12: you removed the next two lines but forgot the trailing `?`

PR #243, opened 2017-06-20
Add slugline field to event

petrjasek:
when creating planning item use it for its slugline and use event
name for headline

SDESK-1449 SDESK-1451


ioanpocol (approved):

vied12 (approved):
weird... where are the travis tests ?

petrjasek:
what about the coverage? I guess it's because of this schema update, not sure how to test this more

vied12:
you did the maximum, this is because the total is lower than 90%. I just reduced the threshold.

PR #245, opened 2017-06-21
[SDESK-1238] Locking/Unlocking feature for planning items

nrvikas:


server/planning/item_lock.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def lock(self, item_id, user_id, session_id, action, resource):
+        item_service = get_resource_service(resource)
+        item = item_service.find_one(req=None, _id=item_id)
+
+        # set the lock_id it per item
+        lock_id = "item_lock {}".format(item.get(config.ID_FIELD))
+
+        if not item:
+            raise SuperdeskApiError.notFoundError()
```
MarkLark86: `if not item` should be above the `lock_id =` assignment
as `.format(item.get` will fail if item is None

client/actions/planning.js
```diff
@@ -321,9 +321,38 @@ const fetchCoverageById = (id) => (
...
+)
+
+const _unlockAndOpenPlanningEditor = (planning) => (
+    (dispatch, getState, { api, notify }) => (
+        api('planning_unlock', planning).save({})
+        .then(() => {
+            // Call openPlanningEditor to obtain a new lock for editing
+            dispatch(openPlanningEditor(planning._id))
+        }, () => {
+            notify.error('Could not unlock on the planning item.')
```
MarkLark86: It would be better to show the error message returned from the server, such as 

> Item is locked by you in another session.

client/actions/planning.js
```diff
@@ -321,9 +321,38 @@ const fetchCoverageById = (id) => (
...
+)
+
+const _lockAndOpenPlanningEditor = (planning) => (
+    (dispatch, getState, { api, notify }) => {
+        // Close planning editor if open
+        dispatch(closePlanningEditor(selectors.getCurrentPlanning(getState())))
+        api.save('planning_lock', {}, { lock_action: 'edit' }, { _id: planning }).then((item) => {
+            dispatch(_openPlanningEditor(item))
+        }, () => {
+            notify.error('Could not obtain lock on the planning item.')
```
MarkLark86: It would be better to show the error message returned from the server, such as 

> Item is locked by you in another session.

server/planning/planning_lock.py
```diff
@@ -0,0 +1,71 @@
...
+class PlanningUnlockService(BaseService):
+
+    def create(self, docs, **kwargs):
+        user = get_user(required=True)
+        auth = get_auth()
+        item_id = request.view_args['item_id']
+        item = get_component(ItemLockPlanningModule).unlock(item_id, user['_id'], auth['_id'], 'planning')
+
+        if item is None:
+            # version 1 item must have been deleted by now
```
mdhaman: There is no versioning enabled on planning. This case won't be there.

client/utils/index.js
```diff
@@ -259,3 +259,9 @@ export const isEventAllDay = (startingDate, endingDate) => {
     return startingDate.isSame(startingDate.clone().startOf('day')) &&
         endingDate.isSame(endingDate.clone().endOf('day').seconds(0).milliseconds(0))
 }
+
+export const isLockedByMe = (item, sessionDetails) => (
```
vied12: I would have created a selector for this instead. Or a selector factory with the item as parameter

client/actions/tests/agenda_test.js
```diff
@@ -113,7 +113,7 @@ describe('agenda', () => {
                     expect(apiSpy.save.args[0]).toEqual([{}, item])
                     expect(notify.success.args[0]).toEqual(['The agenda has been created/updated.'])
 
-                    expect(dispatch.callCount).toBe(10)
+                    expect(dispatch.callCount).toBe(11)
```
vied12: this kind of test doesn't mean much to me. I'm used to removed them when they start to fail.
Unless you convince me to the other way, I will continue to do it, and won't blame you if do it also ;)

client/components/EditPlanningPanelContainer/index.jsx
```diff
@@ -10,47 +10,80 @@ import { get } from 'lodash'
...
+import { UserAvatar, UnlockItem } from '../'
+import { isLockedByMe } from '../../utils'
+import classNames from 'classnames'
 import './style.scss'
 
 export class EditPlanningPanel extends React.Component {
 
     constructor(props) {
         super(props)
+        this.state = { showUnlockDetails: false }
```
vied12: Why using here an internal state instead a store's state ?

client/components/PlanningList/index.jsx
```diff
@@ -33,7 +34,8 @@ class PlanningList extends React.Component {
                     onUnspike={handlePlanningUnspike}
                     onClick={previewPlanning}
                     onDoubleClick={openPlanningEditor}
-                    privileges={privileges} />
+                    privileges={privileges}
+                    session={session}/>
```
vied12: I'm not convinced that `<PlanningItem/>  needs to be aware of the `session`. Maybe just a `isLocked` boolean might be enough ?

client/utils/index.js
nrvikas: Good one, will change it.

client/actions/tests/agenda_test.js
nrvikas: I think it is good to remove it now until we come up with a better way to test this. Will certainly look for something we can use later.

client/components/EditPlanningPanelContainer/index.jsx
nrvikas: I have renamed the prop. This is actually to open the 'Unlock' pop-up for the user to unlock the item. So, it has to be an internal state to toggle the popup open and close behavior.

PR #246, opened 2017-06-21
feat(events): add item actions menu to detail

petrjasek:
with spike, unspike and create planning item actions

SDESK-1437 SDESK-1439


petrjasek:
pls let me know if the way I'm putting this together makes sense.
also not sure how to mark those action labels for translation, any hints?
@ioanpocol @vied12

PR #250, opened 2017-06-22
Event detail: bigger name SDESK-1438

vied12:


client/components/EventForm/style.scss
```diff
@@ -9,6 +9,14 @@
         display: flex;
         align-items: center;
     }
+    [name="slugline"] {
+        color: $sd-keyword;
+        font-weight: 700;
+    }
+    [name="name"] {
+        font-size: 19px;
+        color: $gray;
```
nrvikas: When I tested this on the test instance, the font-size looked 'just a little' bigger than the mock-ups: https://wiki.sourcefabric.org/display/NR/Planning+interface?preview=/24052602/41910584/Planning_Split-view_Event-details_2.png.

Modified it to 17px on chrome and looked closer to mock-up.
I might be wrong because users have different screen resolutions and monitor sizes.

Otherwise, all great.

PR #252, opened 2017-06-23
fix(tests): Action tests with promises not catching errors

MarkLark86:
fix(tests): Lint warnings

### DIFF
--- a/client/actions/tests/privileges_test.js
+++ b/client/actions/tests/privileges_test.js
@@ -24,7 +24,7 @@ describe('privileges', () => {
             $timeout.reset()
         })
 
-        it('loadPrivileges', () => {
+        it('loadPrivileges', (done) => {
             const privileges = {
                 loaded: Promise.resolve(),
                 privileges,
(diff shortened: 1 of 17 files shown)


client/actions/tests/privileges_test.js
```diff
@@ -36,6 +36,15 @@ describe('privileges', () => {
                     type: PRIVILEGES.ACTIONS.RECEIVE_PRIVILEGES,
                     payload: privileges.privileges,
                 }])
+
+                done()
+            })
+            .catch((error) => {
+                /* eslint-disable no-console */
+                console.log('Unhandled exception: ' + error + '\n' + error.stack)
```
vied12: Do we need to keep all these `console.log` ?

MarkLark86: Yes, as they are valuable to determining what caused the error.
Otherwise it merely shows an error such as 'event is undefined'

With the console output, we get the method that was last called, i.e. the stacktrace

PR #258, opened 2017-06-27
[SDESK-1225] Events Lock/Unlock feature

nrvikas:


client/actions/events.js
```diff
@@ -690,22 +690,68 @@ const previewEvent = (event) => ({
...
 /**
  * Action to close the Edit Event panel
  * @return object
  */
-const closeEventDetails = () => (
-    { type: EVENTS.ACTIONS.CLOSE_EVENT_DETAILS }
+const closeEventDetails = (event) => (
+    (dispatch, getState, { api, notify }) => {
+        if (selectors.isEventDetailLockedInThisSession(getState())) {
+            api('events_unlock', event).save({})
```
vied12: I would rather create actions to lock and unlock an event in order to avoid code duplication

client/components/EventForm/index.jsx
```diff
@@ -104,15 +117,22 @@ export class Component extends React.Component {
...
+        const updatedReadOnly = readOnly || eventSpiked ||
+            (get(initialValues, '_id') && !lockedInThisSession)
         const creationDate = get(initialValues, '_created')
         const updatedDate = get(initialValues, '_updated')
         const id = get(initialValues, '_id')
         const author = get(initialValues, 'original_creator') && users ? users.find((u) => (u._id === initialValues.original_creator)) : null
         const versionCreator = get(initialValues, 'version_creator') && users ? users.find((u) => (u._id === initialValues.version_creator)) : null
         const isPublished = get(initialValues, 'pubstatus') === EVENTS.PUB_STATUS.USABLE
+        const lockedUser = this.getLockedUser(initialValues)
+        // const lockedUser = null
```
vied12: please remove comments

client/components/EventsListContainer/index.jsx
```diff
@@ -75,7 +75,8 @@ class EventsListComponent extends React.Component {
                                 loadMoreEvents={this.props.loadMoreEvents}
                                 selectedEvents={this.props.selectedEvents}
                                 onEventSelectChange={this.props.onEventSelectChange}
-                                privileges={privileges} />
+                                privileges={privileges}
+                                session={session} />
```
vied12: does the component need to be aware of the session, or `isLocked` could be enough ?

client/reducers/events.js
```diff
@@ -116,14 +116,39 @@ const eventsReducer = createReducer(initialState, {
...
-            ...state,
-            showEventDetails: payload,
-            highlightedEvent: payload,
-            readOnly: false,
+    [EVENTS.ACTIONS.OPEN_EVENT_DETAILS]: (state, payload) => {
+        if (payload.lock_user) {
+            let _events = cloneDeep(state.events)
+            let event = payload
+
+            // Change dates to moment objects
```
vied12: we are already doing that in more and more different places. It starts to alert me.
Just to know how we could factorize it later, do you remember in which scenario the dates here are not moment objects ?

client/selectors/index.js
```diff
@@ -277,3 +277,12 @@ export const isCurrentPlanningLockedInThisSession = createSelector(
         currentPlanning.lock_session === session.sessionId ? true : false
     )
 )
+
+export const isEventDetailLockedInThisSession = createSelector(
+    [getShowEventDetails, getEvents, getSessionDetails],
+    (showEventDetails, events, session) => {
+        const event = events[showEventDetails]
+        return event && event.lock_user === session.identity._id &&
+        event.lock_session === session.sessionId ? true : false
```
vied12: wow, this looks a bit mysterious if you forget the operator priorities. It would be a bit more explicit with parentheses.
Also I like to use `!!` to cast to boolean, but it's matter of taste.
Ex: `session.sessionId  ? true : false` === `!!session.sessionId`

client/actions/events.js
nrvikas: Lock and unlock need not update the state from reducer as the websocket notifications triggers actions that does the store update anyways.

vied12: Personally I don't mind if there are actions that don't update the state now, because they could one day.
My point was to avoid to see the same API and `notify` calls in different functions.

client/components/EventsListContainer/index.jsx
nrvikas: It is needed to pass on EventList as a prop - as EventList is not connected to the store. Instead of connecting EventList to store only for this purpuse, I thought it is better to send 'session' as a prop. EventList needs to know it so that it can compute if an EventItem is locked in this session, in which case, the owner of the lock can still go on and spike or unspike an event. Just as in Superdesk authoring where the current owner of the lock can spike an item he/she is editing.

client/actions/events.js
nrvikas: Sure, I'll change it.

client/reducers/events.js
nrvikas: This is needed when the back-end returns the event after locking it as the dates won't be moment objects. I will have the lock/unlock action update the state and see if this can be removed.

client/components/EventsListContainer/index.jsx
vied12: I see.
we are now passing too many props to `EventItem` from `EventListContainer` and `EventList` which don't need them.
We should maybe start to connect `EventItem` or a new container of it in order to avoid this in the future.

nrvikas: I would hold off on this idea a little longer because I'm concerned about 100s of event items subscribing to the store. Component creation or update is inevitable as their props change, but we should probably be a little more careful in making redux store connections.

client/selectors/index.js
nrvikas: Hmmm, I'm not convinced about the mystery behind this - it's pretty straight forward using just logical 'and' operators and a ternary operator in combination - I will consider changing this for your benefit though - no worries.

client/components/EventsListContainer/index.jsx
vied12: Thanks to react-virtualized we only render events that we can display.
And in a way, there are already connected to the store, through `EventListContainer`.

PR #260, opened 2017-06-29
[SDESK-1253] - (event history) UI:  I want to record history of actio…

sjunaid:
…ns on an Event.


client/components/AbsoluteDate/index.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import moment from 'moment'
+
+/**
+ * Display absolute date in <time> element
+ *
+ * Usage:
+ * <AbsoluteDate date={historyItem._created} />
```
vied12: We already have a `<Datetime/>` component that use the datetime format from the SD config. Did you notice it ? Is it incompatible with what you need ?

client/components/AbsoluteDate/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * Params:
+ * param {object} date - datetime string in utc
+ */
+export const AbsoluteDate = ({ date }) => {
+    const COMPARE_FORMAT = 'YYYY-M-D'
```
sjunaid: moved.

client/components/AbsoluteDate/index.jsx
sjunaid: hi @vied12,  The desired output to display timestamp for eventHistory is similar to what is displaying in article's itemHistory tab in SD(in preview pane) like `Wednesday, 28. June 13:28` or `Today, 30. June 12:06` and so on, so i tried AbsoluteDate component which is almost exact implementation of sdAbsdate directive(in /scripts/core/datetime/absdate-directive.js), used in itemHistory. whereas, DateTime comp. is giving config format. please suggest.

PR #266, opened 2017-07-04
fix(save&publish): Prevent save&publish when there are validation errors

ioanpocol:


client/components/EventForm/index.jsx
```diff
@@ -441,37 +444,26 @@ const mapStateToProps = (state) => ({
     unlockPrivilege: selectors.getPrivileges(state).planning_unlock ? true : false,
 })
 
+const dispatchSubmit = (dispatch, event) => dispatch(actions.askConfirmationBeforeSavingEvent(event))
```
nrvikas: This can be an independent action in the actions folder rather than a const function here.

PR #273, opened 2017-07-05
feat(coverage history) Add history tracking for coverage

marwoodandrew:


marwoodandrew:
I would merge this if I could, but don't seem to have merge rights!

PR #277, opened 2017-07-06
Planning UI

darconny:
- improvements

### DIFF
--- a/client/components/PlanningPanelContainer/index.jsx
+++ b/client/components/PlanningPanelContainer/index.jsx
@@ -96,12 +96,23 @@ class PlanningPanel extends React.Component {
                                 </div>
                             || !currentAgenda &&
                                 <div className="Planning-panel__empty-message">
-                                    Choose an agenda from the drop-down list above.
+                                    <div className="panel-info">
+                                        <div className="panel-info__icon">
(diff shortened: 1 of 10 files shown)


client/components/PlanningPanelContainer/index.jsx
```diff
@@ -96,12 +96,23 @@ class PlanningPanel extends React.Component {
...
+                                        <p className="panel-info__description">...from the drop-down list above.</p>
+                                    </div>
                                 </div>
                             || (planningList && planningList.length < 1) &&
                                 <div className="Planning-panel__empty-message">
-                                    There are no planning items in this agenda.<br/>
-                                    Drag an event here to start one.
+                                    <div className="panel-info">
+                                        <div className="panel-info__icon">
+                                            <i className="big-icon--add-to-list" />
```
nrvikas: The add-to-list icon somehow leads me to click on it as though it is a button but nothing happens. '+' icon used in other places are action buttons which we click to add a subject/category/planning-item. Hence the confusion. It's a matter of design choice, but thought of giving my two cents opinion on this.

darconny: Thanks, I put this icon as its closest from what we have now within big icon font. We could change it later on.

PR #279, opened 2017-07-06
Notification on AddEventToCurrentAgenda

vied12:
Display better notification when a user try to create a planning from an event if there is no agenda selected

SDESK-1589

### DIFF
--- a/client/actions/agenda.js
+++ b/client/actions/agenda.js
@@ -5,7 +5,7 @@ import { cloneDeep, get } from 'lodash'
 import { PRIVILEGES, ITEM_STATE, AGENDA } from '../constants'
 import { checkPermission, getErrorMessage } from '../utils'
 
-import { planning } from './index'
+import { planning, showModal } from './index'
 
 /**
  * Creates or updates an Agenda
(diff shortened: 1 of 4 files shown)


client/actions/agenda.js
```diff
@@ -184,13 +184,49 @@ const _addToCurrentAgenda = (plannings) => (
...
+                    body: `Do you want to add these ${events.length}
+                        events to the current agenda ?`,
+                    action: () => dispatch(addEventToCurrentAgenda(events)),
+                },
+            }))
+        } else {
+            dispatch(showModal({
+                modalType: 'CONFIRMATION',
+                modalProps: {
+                    body: 'You have to select first an agenda',
```
nrvikas: This probably should be "You have select an agenda first" or "You have to first select an agenda"

client/actions/agenda.js
```diff
@@ -184,13 +184,49 @@ const _addToCurrentAgenda = (plannings) => (
...
+ * @param {array} events - The events needed to create the planning items
+ */
+const askForAddEventToCurrentAgenda = (events) => (
+    (dispatch, getState) => {
+        const currentAgenda = selectors.getCurrentAgenda(getState())
+        if (currentAgenda) {
+            dispatch(showModal({
+                modalType: 'CONFIRMATION',
+                modalProps: {
+                    body: `Do you want to add these ${events.length}
```
nrvikas: In case of a single event, it still mentions "events" and not "event". For a single event, it can probably just be "Do you want to add this event to the current agenda" ? That's probably more common.

client/actions/agenda.js
```diff
@@ -184,13 +184,49 @@ const _addToCurrentAgenda = (plannings) => (
...
  * Action dispatcher that creates a planning item from the supplied event,
  * then adds this to the currently selected agenda
  * @param {object} event - The event used to create the planning item
  * @return Promise
  */
 const _addEventToCurrentAgenda = (events) => (
     (dispatch, getState, { notify }) => {
+        const currentAgenda = selectors.getCurrentAgenda(getState())
+        if (!currentAgenda) {
+            let errorMsg = 'Select first an agenda'
```
nrvikas: Kindly change the error message to "Select an agenda first".

client/actions/agenda.js
vied12: Thanks !

PR #280, opened 2017-07-07
[SDESK-1535] Spiking (from) a series of Recurring Events

MarkLark86:


client/actions/events/ui.js
```diff
@@ -0,0 +1,119 @@
...
+)
+
+const openSpikeModal = checkPermission(
+    _openSpikeModal,
+    PRIVILEGES.SPIKE_EVENT,
+    'Unauthorised to spike an Event'
+)
+
+const self = {
+    _openSpikeModal,
```
vied12: I think we use `_*` names for internal used variables. If we export them, they are not internal anymore.
I guess they are exported for the tests and I found recently a nice module helping for accessing and modifying internal variables: https://github.com/jhnns/rewire
I don't know if it worth it... what do you think ?

client/components/SpikeEventForm/index.js
```diff
@@ -0,0 +1,139 @@
...
+            plannings = Object.keys(storedPlannings).filter((pid) => (
+                (event._recurring.ids || []).indexOf(storedPlannings[pid].event_item) > -1
+            )).map((pid) => ({ ...storedPlannings[pid] }))
+
+        } else {
+            throw Error('Incorrect update method: ' + method)
+        }
+
+        this.setState({ plannings })
+        this.props.initialValues._plannings = plannings
```
vied12: I think modifying the props from the component is an anti pattern.
Setting the state like you did just before is not enough ?

Also I feel there are a lot of logic here, more than needed for the component. Can moving it to an action be an option ?

client/actions/events/api.js
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        return Promise.all(promises)
+        .then((responses) => {
+            let events = responses
+                .map((e) => e._items)
+                .reduce((a, b) => a.concat(b))
+
+            dispatch(self.receiveEvents(events))
+            dispatch(self.setEventsList(events.map((e) => e._id)))
```
vied12: In this api module, I think we should avoid to call UI actions.
I would call api actions from a UI action, and in a `then() `, dispatch the following UI actions there

client/actions/events/ui.js
MarkLark86: Two things with the rewire library. Rewire is:
* Unable to mock `const`
* Does not work with transpilers such as Babel

All our functions are constants, and based on the webpack config we use Babel to transpile the code. Not sure if rewire will work with our framework/environment.

client/components/SpikeEventForm/index.js
MarkLark86: I will convert this component to a `stateless functional component`, and move the filtering logic into a `reducer`.

PR #284, opened 2017-07-10
Coverage type: icon-file as the default icon SDESK-1334

vied12:


client/components/PlanningItem/index.jsx
```diff
@@ -68,7 +72,18 @@ const PlanningItem = ({
                 </div>
                 <div className="sd-list-item__row">
                     {coveragesTypes.map((c, i) => (
-                        <span key={i}><i className={coverageIcons[c]}/>&nbsp;</span>
+                        <span key={i}>
+                            <OverlayTrigger
+                                placement="bottom"
+                                overlay={
```
nrvikas: Would it be better to have the tool-tip text to have a leading uppercase letter as in "Audio" instead of "audio" ?

PR #287, opened 2017-07-10
Adds description to planning form

vied12:
SDESK-1592


ioanpocol:
Do we need any change on backend?

vied12:
we don't need because there is already a `description_text` field

PR #288, opened 2017-07-11
[SDESK-1556][SDESK-1568] (feat):Modifying planning item so that it can be added to multiple agendas.

mdhaman:
- Agenda management move to planning settings
- multiple agenda for planning items.
- planning item should be in at least one agenda


client/actions/agenda.js
```diff
@@ -295,71 +237,30 @@ const _createPlanningFromEvent = (event) => (
...
-)
-
-/**
  * Action dispatcher that fetches all planning items for the
  * currently selected agenda
  * @return arrow function
  */
 const fetchSelectedAgendaPlannings = () => (
     (dispatch, getState) => {
         const agenda = selectors.getCurrentAgenda(getState())
```
MarkLark86: Do we need `const agenda` and just use `const agendaId?`

client/components/AuditInformation/index.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import { connect } from 'react-redux'
+import * as selectors from '../../selectors'
+import { get } from 'lodash'
+import moment from 'moment'
+
+export const AuditInformationComponent = ({
```
nrvikas: +1 for this component.

mdhaman:
Closing this PR.

PR #293, opened 2017-07-12
feat(publish): publish events

petrjasek:
on publish put events into publish queue and let superdesk to transmit it

SDESK-60

requires https://github.com/superdesk/superdesk-core/pull/983

### DIFF
--- a/server/features/publish.feature
+++ b/server/features/publish.feature
@@ -1,0 +1,152 @@
+Feature: Publish
+
+    @auth
+    Scenario: Publish Event
+        When we post to "/products" with success
+        """
+        {
+            "name":"prod-1","codes":"abc,xyz", "product_type": "both"
+        }
+        """
+        And we post to "/subscribers" with success
+        """
+        {
+            "name":"News1","media_type":"media", "subscriber_type": "digital", "sequence_num_settings":{"min" : 1, "max" : 10}, "email": "test@test.com",
+            "products": ["#products._id#"],
+            "codes": "xyz, abc",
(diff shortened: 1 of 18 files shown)


server/planning/events_publish.py
```diff
@@ -0,0 +1,51 @@
...
+        return ids
+
+    def validate_event(self, event):
+        try:
+            assert event.get('pubstatus') == PUB_STATUS_USABLE
+        except AssertionError:
+            abort(409)
+
+    def publish_event(self, event):
+        event.setdefault(ITEM_TYPE, 'event')
```
mdhaman: Why is `item_type` set only on publish? We can do the same thing on create.

server/features/publish.feature
```diff
@@ -0,0 +1,88 @@
+@wip
+Feature: Publish
+
+    @auth
+    Scenario: Publish Event
+        When we post to "/products" with success
+        """
+        {
+            "name":"prod-1","codes":"abc,xyz", "product_type": "both"
```
mdhaman: Since we would be publishing event and schema is different. It would be good to write a scenario where event is published to a subscriber and the subscriber is having is content filters matching as well as not matching the event schema field. i.e. product with a category filter and product with a desk filter.

petrjasek: I'm not sure yet about this whole filtering, if filters will be shared or there will be some switch for content type on it or destination. atm it's more like spike to see if we can reuse the superdesk publishing

mdhaman: ok. got it

PR #294, opened 2017-07-12
SDESK-1591 As a user I want to have a "short description" field on events

mugurrus:


nrvikas:
All fine, but now, the long and short descriptions both have 'line-input', which visually makes very little difference to the end user. Taking out the 'line-input' class name from text area field in Event Form will make long description look like a box which indicates to the user that he/she can type a long description here. It's more of a UI appearance decision, I will leave it to you 😃

PR #296, opened 2017-07-13
[SDESK-1626] Don't lock an item if you already hold a lock on it

nrvikas:


client/components/EventsList/index.jsx
```diff
@@ -46,6 +46,15 @@ export class EventsList extends React.Component {
         }
     }
 
+    previewOrEditEvent(event) {
+        // If we have the lock in this session, dispatch openPlanningEditor instead
```
vied12: I would do this in the action in order to keep `onClick` and `onDoubleClick` simple without logic

PR #305, opened 2017-07-17
[SDESK-1562] Remove 'ends never' option from recurring events. Allow …

MarkLark86:
…only a system configured maximum number of items in a series.


server/features/steps/steps.py
```diff
@@ -53,7 +53,7 @@ def steip_impl_store_indexed_item_to_ctx(context, tag, index):
 def step_impl_then_get_event_file(context):
     assert_200(context.response)
     data = get_json_data(context.response)
-    url = '/upload/%s/raw' % data['filemeta']['media_id']
+    url = '/upload-raw/%s' % data['filemeta']['media_id']
```
vied12: the endpoint has changed ? did someone already changed it in the code base ?

MarkLark86: This was changed in superdesk-core on Apr 14 with this commit: superdesk/superdesk-core@6e0fc66

vied12: so you should maybe change it in the planning code base too

MarkLark86: The planning codebase uses the angular `upload` service from superdesk-client-core, therefor there is no need to change anything in the planning codebase.

PR #307, opened 2017-07-18
[SDESK-1643](fix): Refresh the agenda list when agenda is deleted.

mdhaman:


client/components/CreateEditAgendaForm.jsx
```diff
@@ -3,15 +3,18 @@ import { fields, Toggle, AuditInformation } from '../components'
...
 import * as actions from '../actions'
 import { ChainValidators, RequiredFieldsValidatorFactory, MaxLengthValidatorFactory } from '../validators'
 
 export class Component extends React.Component {
 
     constructor(props) {
         super(props)
-        this.state = { isEnabled: this.props.initialValues.is_enabled }
+        this.state = {
+            isEnabled: isUndefined(this.props.initialValues.is_enabled)
```
vied12: in this case I use `get(this.props.initialValues, 'is_enabled', true)`
but it's matter of taste :)

PR #326, opened 2017-07-25
[SDESK-1648] Add urgency field to planning form

nrvikas:


client/components/fields/ColoredValueSelectField/style.scss
```diff
@@ -0,0 +1,19 @@
...
+    @include border-radius(3px);
+    max-width: 200px;
+    @include border-box();
+
+    button {
+    	padding: .6rem 1.6rem;
+    	width: 100px;
+    	text-align: left;
+    }
+}
```
MarkLark86: It would be good to have the `<li>` options highlight on mouse over like in the rest of Superdesk

PR #344, opened 2017-07-31
feat(assignments): Add assignments list

ioanpocol:


client/actions/assignment.js
```diff
@@ -0,0 +1,200 @@
...
+
+        return dispatch(query({
+            filterBy: filterBy,
+            searchQuery: searchQuery,
+            orderByField: orderByField,
+            orderDirection: orderDirection,
+        }))
+        .then((data) => {
+            dispatch(receivedAssignments(data._items))
+            return Promise.resolve(data._items)
```
petrjasek: is `Promise.resolve` here and `Promise.reject` in the error handler needed?

client/actions/assignment.js
```diff
@@ -0,0 +1,200 @@
...
+ * Action dispatcher to load the list of assignments for current list settings.
+ * @param {string} filterBy - the filter by desk or user ('All', 'User')
+ * @param {string} searchQuery - the text used for free text query
+ * @param {string} orderByField - the field used to order the assignments ('Created', 'Updated')
+ * @param {string} orderDirection - the direction of order ('Asc', 'Desc')
+ */
+const loadAssignments = (filterBy, searchQuery, orderByField, orderDirection) =>
+    (dispatch) => {
+        dispatch({
+            type: 'CHANGE_LIST_SETTINGS',
```
petrjasek: imo would be good to have a function for this, to be consistent

client/actions/assignment.js
```diff
@@ -0,0 +1,200 @@
...
+            payload: { lastAssignmentLoadedPage: lastAssignmentLoadedPage + 1 },
+        })
+
+        dispatch(query({
+            filterBy: filterBy,
+            searchQuery: searchQuery,
+            orderByField: orderByField,
+            orderDirection: orderDirection,
+            page: lastAssignmentLoadedPage + 1,
```
petrjasek: imo this should use the updated state, what about first dispatching that `CHANGE_LIST_SETTINGS` and after that get the query params from state? actually query could get those params from state

client/components/AssignmentItem/index.jsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { ListItem, Checkbox, UserAvatar, AbsoluteDate } from '../index'
+import classNames from 'classnames'
+import { get } from 'lodash'
+import moment from 'moment'
+
+const getCoverageIcon = (type) => {
+    const coverageIcons = {
```
nrvikas: As we have this function in PlanningItem/index.jsx, we can probably move this to utils/index.js and use it from there.

PR #351, opened 2017-08-03
[SDESK-1708] As an Editor I want to be able to mark a Planning Item as 'not for publication'

MarkLark86:


server/planning/planning.py
```diff
@@ -147,52 +144,24 @@ def get_planning_by_agenda_id(self, agenda_id):
...
-    },
-    'original_source': {    # This value is extracted from the ingest
-        'type': 'string',
-        'mapping': not_analyzed
-    },
-    'ingest_provider_sequence': {
-        'type': 'string',
-        'mapping': not_analyzed
-    },
+    'ingest_provider': metadata_schema['ingest_provider'],
```
mdhaman: @petrjasek @ioanpocol I don't know why do we need following fields in `planning` and `coverage` schema?
- ingest_provider, source, original_source, ingest_provider_sequence, ingest_id, unique_name, version, unique_id

Do you guys see a use case where planning items will be ingested?

nrvikas:
+1 for adding dependencies to metadata_schema for plannings schema.

server/planning/planning.py
petrjasek: not really, this was probably copied form events

PR #362, opened 2017-08-09
[SDESK-1563] Use scheduled date(s) of Coverage(s) for time filtering planning views

mdhaman:
- Includes no default coverage when Planning Item is created
- Schedule date of coverage is not compulsory.
- If coverage date is not available on any of the coverages then `planning item date`  is used for sorting and filtering.
- `planning item date` is `event start date` for planning items created from associated event and for ad-hoc planning item the `planning item date` is current date time.


mdhaman:
@ioanpocol , @petrjasek if you guys update the sdp-master i.e. planning instance it requires reset of the database/index as index definition has been modified.

PR #364, opened 2017-08-10
[SDESK-1761] Unspike action on event item and restructure item-actions on event form

nrvikas:


client/components/EventForm/index.jsx
```diff
@@ -209,66 +209,25 @@ export class Component extends React.Component {
...
         let itemActions = []
-
-        const populateItemActions = () => {
-            itemActions.unshift(eventActions.EVENT_HISTORY)
-
-            if (eventUtils.canCreatePlanningFromEvent(initialValues, session, privileges)) {
-                itemActions.unshift(eventActions.CREATE_PLANNING)
+        if (existingEvent) {
+            const callBacks = {
+                [GENERIC_ITEM_ACTIONS.SPIKE.label]: () => spikeEvent(initialValues),
```
MarkLark86: Should we bind this (as well as duplicateEvent and addEventToCurrentAgenda) instead of calling it from an arrow function?

client/components/EventItem/index.jsx
```diff
@@ -24,14 +22,12 @@ export const EventItem = ({
         session,
     }) => {
     const hasBeenCanceled = get(event, 'occur_status.qcode') === 'eocstat:eos6'
-    const itemLockedInThisSession = isItemLockedInThisSession(event, session)
-
-    const callBacks = { [GENERIC_ITEM_ACTIONS.SPIKE.label]: () => onSpikeEvent(event) }
+    const callBacks = {
+        [GENERIC_ITEM_ACTIONS.SPIKE.label]: () => onSpikeEvent(event),
```
MarkLark86: Same as with my previous comment, would it be better to bind instead of calling from within an arrow function?

client/constants/index.js
```diff
@@ -39,6 +39,15 @@ export const GENERIC_ITEM_ACTIONS = {
...
     },
+    UNSPIKE: {
+        label: 'Unspike',
+        icon: 'icon-unspike',
+    },
+    DUPLICATE: {
+        label: 'Duplicate',
+        icon: 'icon-copy',
+    },
+    HISTORY: { label: 'View History' },
```
MarkLark86: Do we have an icon for history?

nrvikas: @fritzSF , as we don't have an icon for item history, we thought of using the same we use for version history in Superdesk (big-icon--history). But we don't have the small sized counterpart of that icon. Could you please advise ?

PR #365, opened 2017-08-10
[SDESK-1672] Add a label to the occurence status vocab

marwoodandrew:


client/components/fields/OccurStatusField.jsx
```diff
@@ -6,12 +6,12 @@ const mapStateToProps = (state, ownProps) => ({
     clearable: false,
     options: state.vocabularies.eventoccurstatus.map((s) => (
         {
-            label: s.name,
+            label: s.label,
```
nrvikas: Just a thought, instead of enforcing all Superdesk instances to have 'label' in their vocabulary for OccurStatus, we can use lodash's get() and do this instead ?
label: get(s, 'label', s.name)

PR #372, opened 2017-08-15
[SDESK-1692] Convert redux store into an Angular service for usage with Planning, Settings and Assignments controllers

MarkLark86:


client/services/PlanningStoreService.js
```diff
@@ -0,0 +1,102 @@
...
+    userList,
+    desks,
+    metadata,
+    session,
+    deployConfig,
+    gettextCatalog
+) {
+    let self = this
+
+    self.store = null
```
petrjasek: just wondering why we keep the reference here, it's using the store only in controller right?
I would prefer if it would be used in controller like `const store = planningStore.createStore()`; so we don't have to deal with its reseting later

MarkLark86: The service is implemented as a Singleton Instance, created upon first use. This reduces the overhead required to create the Redux store, especially when used in the Superdesk Authoring views (such as creating a Planning item from a story).

The following are the current reducers and when they're loaded:

| Permanent  | Ad hoc |
| ------------- | ------------- |
| desks | agenda |
| genres | assignment |
| ingest_providers | events |
| modal | forms |
| privileges | formsProfile |
| session | planning |
| subjects |  |
| urgency |  |
| users |  |
| vocabularies |  |

The ad hoc reducers are set to `null` when `actions.resetStore` is dispatched, and set to `initialState` when `actions.initStore` is dispatched (ready for use by other actions and components).

The other way would be to create the Redux store from scratch every time, and populate with all the reducers (with API calls for almost all reducers).

Let me know which you would prefer and I will make the changes if necessary.

petrjasek: I'm not sure if all this loading should be done via actions dispatching, imo would be better to fetch it all and only create the store after. Ideally this would be fetched via [route.resolve](https://docs.angularjs.org/api/ngRoute/provider/$routeProvider). Because before it's all loaded there is no point really to render anything.

client/services/PlanningStoreService.js
```diff
@@ -0,0 +1,149 @@
...
+                    userList,
+                    desks,
+                    metadata,
+                    session,
+                    deployConfig,
+                    gettextCatalog,
+                },
+            })
+
+            $rootScope.$on('vocabularies:updated', angular.bind(this, this._reloadVocabularies))
```
petrjasek: imo this shouldn't be part of createStore otherwise you register this every time you call that, instead of just once in constructor

PR #386, opened 2017-08-22
feat(assignment): Assignment preview

ioanpocol:


client/components/EditAssignmentContainer/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+                        <AuditInformation
+                            createdBy={author}
+                            updatedBy={versionCreator}
+                            createdAt={creationDate}
+                            updatedAt={updatedDate} />
+                    </div>
+                    <div className="EditAssignmentPanel__actions">
+                        { readOnly &&
+                            <OverlayTrigger placement="bottom" overlay={tooltips.editTooltip}>
+                                <button className="EditAssignmentPanel__actions__edit"
```
mdhaman: As per our last discussion with @karelpetrak, it will be just a read only coverage.  The user is allowed to change assignment only. @ioanpocol do we need this edit action?

ioanpocol: When I started to implement this feature it was a task to be able to edit the assignment too.
I will drop the partial implemented assignment edit.

client/selectors/index.js
```diff
@@ -78,6 +78,10 @@ export const getPlanningItemReadOnlyState = (state) => get(state, 'planning.read
 export const getEventReadOnlyState = (state) => get(state, 'events.readOnly')
 export const getSessionDetails = (state) => get(state, 'session')
 export const getCurrentUserId = (state) => get(state, 'session.identity._id')
+export const getEditorOpened = (state) => get(state, 'assignment.editorOpened')
```
nrvikas: Can getEditorOpened be renamed to 'getAssignmentEditorOpened' please ? As we already have event and planning editors.

client/components/EditAssignmentContainer/index.jsx
mdhaman: @ioanpocol you can change assignment details i.e reassign, un-assign actions are allowed but coverage details (description, slugline, etc) are readonly.

ioanpocol:
@darconny  Panel position is fixed now.
I just reused the Preview component from edit planning -> edit coverage, so the changes should be performed there and accommodate them with with planning inputs. I think this should be done in a different task.

PR #393, opened 2017-08-24
[SDESK-1669] - Allow a single user to lock more than one item at a ti…

sjunaid:
…me in Planning

### DIFF
--- a/client/actions/events/ui.js
+++ b/client/actions/events/ui.js
@@ -82,7 +82,7 @@ const previewEvent = (event) => (
 
 /**
  * Action to close the Edit Event panel
- * @return object
+ * @return Promise
  */
 const closeEventDetails = () => (
     (dispatch, getState) => {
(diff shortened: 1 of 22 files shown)


client/actions/events/ui.js
```diff
@@ -96,6 +96,36 @@ const closeEventDetails = () => (
...
+ * Unlock a Event and close editor if opened - used when item closed from workqueue
+ * @param {object} item - The Event item to unlock
+ * @return Promise
+ */
+const unlockAndCloseEditor = (item) => (
+    (dispatch, getState, { notify }) => (
+        dispatch(eventsApi.unlock({ _id: item._id }))
+        .then(() => {
+            if (selectors.getHighlightedEvent(getState()) === item._id) {
+                dispatch({ type: EVENTS.ACTIONS.CLOSE_EVENT_DETAILS })
```
MarkLark86: It would be better to use `dispatch(self.minimizeEventDetails())` here

client/selectors/index.js
```diff
@@ -308,3 +308,17 @@ export const isAdvancedDateSearch = createSelector(
         !!get(currentSearch, 'advancedSearch.dates.range', false)
     )
 )
+
+/** Returns the list of currently locked planning items */
+export const getLockedPlannings = (state) => filter(state.planning.plannings, matches(
```
MarkLark86: Use `createSelector([getStoredPlannings, getCurrentUserId], (plannings, userId) => ....` when creating this selector for better optimization.

PR #394, opened 2017-08-24
fix(assignments): Refresh the list of assignments when the current d…

ioanpocol:
…esk is changed

### DIFF
--- a/client/controllers/AssignmentController.js
+++ b/client/controllers/AssignmentController.js
@@ -7,11 +7,13 @@ import { AssignmentListContainer } from '../components'
 AssignmentController.$inject = [
     '$element',
     '$scope',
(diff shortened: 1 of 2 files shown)


client/controllers/AssignmentController.js
```diff
@@ -19,6 +19,13 @@ export function AssignmentController(
         store.dispatch(actions.initStore())
         store.dispatch(actions.loadAssignments('All', null, 'Created', 'Asc'))
         .then(() => {
+            const desks = sdPlanningStore.getDesksService()
```
mdhaman: Do we need `sdPlanningStore.getDesksService()` method? Can we inject `desks` service into the controller?

PR #416, opened 2017-09-06
feat(export): add export planning items as article

petrjasek:
it allows user to select planning items and run export

SDESK-301 SDESK-1863 SDESK-1864

### DIFF
--- a/client/actions/planning/api.js
+++ b/client/actions/planning/api.js
@@ -967,6 +967,26 @@ const markPlanningPostponed = (plan, reason) => ({
     },
 })
 
+/**
+ * Export selected planning items as a new article
+ */
+function exportAsArticle() {
+    return (dispatch, getState, { api, notify, gettext }) => {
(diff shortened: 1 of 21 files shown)


client/actions/planning/api.js
```diff
@@ -967,6 +967,24 @@ const markPlanningPostponed = (plan, reason) => ({
...
 })
 
+/**
+ * Export selected planning items as a new article
+ */
+function exportAsArticle() {
+    return (dispatch, getState, { api, notify, gettext }) => {
+        const state = getState()
+
+        return api.save('planning_export', {
```
mdhaman: can we display the error if the call fails?

petrjasek: sure, will add it

petrjasek: added just generic error message, not sure atm what could get wrong :wink:

PR #420, opened 2017-09-07
[SDESK-1879] Coverage icons in planning list to show user and desk assigned

nrvikas:


nrvikas:
@fritzSF , sure. Made the changes accordingly. Tks for the feedback.

fritzSF:
@nrvikas Yes, the new tooltips don't support multi-lines. It's OK to use the old tooltips in those specific cases.

PR #421, opened 2017-09-07
feat(export): allow user to reorder planning items before exporting

petrjasek:
SDESK-301


client/actions/planning/api.js
```diff
@@ -969,21 +969,39 @@ const markPlanningPostponed = (plan, reason) => ({
...
-        })
-        .then(() => {
-            notify.success(gettext('Article was created'))
-            dispatch(actions.planning.ui.deselectAll())
-        }, () => {
-            notify.error(gettext('There was an error when exporting'))
-        })
+        const sortableItems = state.planning.selectedItems.map((id) => ({
+            id,
+            label: state.planning.plannings[id].headline,
```
nrvikas: Headline is removed from planning form. So, the label should point to description probably as it is empty.

petrjasek: thx for notice, I've added a fallback to slugline and description if headline is not there

PR #423, opened 2017-09-08
[SDESK-1861] - Define validation for planning, event and coverage forms

sjunaid:


client/components/EventForm/fieldRenders.jsx
```diff
@@ -3,88 +3,97 @@ import React from 'react'
 import { fields } from '../index'
 import { Field, FieldArray } from 'redux-form'
 
-const renderSlugline = (readOnly) => (
+
+const renderSlugline = (args) => (
```
petrjasek: maybe it could be called `props` instead of `args` to be more consistent with other components

PR #434, opened 2017-09-14
fix(events): Adjust End time for events to be 1h later than Start time

ioanpocol:


client/components/EventScheduleForm/index.jsx
```diff
@@ -56,6 +56,10 @@ export class EventScheduleForm extends React.Component {
                 recurringRuleEdited: recurringRuleNextState,
             })
         }
+
+        if (get(this.props, 'currentSchedule.start') !== get(nextProps, 'currentSchedule.start')) {
```
gbbr: Perhaps you can put the sides of this comparison inside a constant, such as:
```js
const start = get(this.props, 'currentSchedule.start');
const nextStart = get(nextProps, 'currentSchedule.start');
```
And then use that. It would be nicer imo.

PR #443, opened 2017-09-19
chore: Use the same library versions between Superdesk-Core and Superdesk-Planning

MarkLark86:


package.json
```diff
@@ -13,72 +13,72 @@
...
+    "redux": "3.6.0",
     "redux-form": "^6.8.0",
     "redux-logger": "^2.7.0",
-    "redux-thunk": "^2.1.0",
+    "redux-thunk": "2.2.0",
     "reselect": "^2.5.4",
     "rrule": "^2.2.0",
-    "style-loader": "^0.13.1",
-    "superdesk-ui-framework": "superdesk/superdesk-ui-framework",
+    "style-loader": "0.18.2",
```
petrjasek: I guess I've missed this one, it should go into devDependencies, but not a big deal. my idea was to keep everything that is needed for testing in dev and only put packages that must be installed in superdesk env in dependencies. the build stack (webpack/babel) should be provided by superdesk-client-core dependencies. makes sense?

PR #447, opened 2017-09-20
[SDESK-1947] (bug) Convert to recurring action still available on planning item for already recurring event

nrvikas:


nrvikas:
Redundant PR as the code already exists.

PR #450, opened 2017-09-20
Assignments as separate collection and sync coverage into planning.

mdhaman:
- assignments are in separate collections with one-to-one relationship with coverages. Assignment entity will be updated from superdesk side when workflow changes. Coverage data on assignment is for search only.
- coverage and assignment information also available in planning for querying using "nested" objects.
- removed parent-child relationship between planning and coverage.
(test are still missing)

@petrjasek please review this PR.


client/actions/assignments/ui.js
```diff
@@ -11,34 +12,36 @@ import { ASSIGNMENTS } from '../constants'
...
         dispatch(
-            changeListSettings(filterBy, searchQuery, orderByField, orderDirection)
+            self.changeListSettings(filterBy, searchQuery, orderByField, orderDirection)
         )
 
-        return dispatch(query())
-        .then((data) =>
+        return dispatch(assignments.api.query())
+        .then((data) => {
             dispatch(receivedAssignments(data._items))
```
MarkLark86: Change this to `dispatch(self.receivedAssignments(data._items))`

client/actions/assignments/ui.js
```diff
@@ -49,91 +52,26 @@ const loadMoreAssignments = () =>
...
  */
-const changeLastAssignmentLoadedPage = (lastAssignmentLoadedPage) => ({
-    type: ASSIGNMENTS.ACTIONS.CHANGE_LIST_SETTINGS,
-    payload: { lastAssignmentLoadedPage },
-})
+const fetch = () =>
+    (dispatch, getState) => (
+        dispatch(assignments.api.query())
+        .then((data) => {
+            dispatch(receivedAssignments(data._items))
```
MarkLark86: Change to `dispatch(self.receivedAssignments(data._items))`

client/components/AssignmentForm/index.jsx
```diff
@@ -54,41 +45,67 @@ export class EditAssignment extends React.Component {
...
+    formProfile: PropTypes.object,
+    usersMergedCoverageProviders: PropTypes.array.isRequired,
+    desks: PropTypes.array.isRequired,
 }
 
 const mapStateToProps = (state) => ({
     initialValues: { assignment: selectors.getCurrentAssignment(state) },
     assignment: selectors.getCurrentAssignment(state),
     users: selectors.getUsers(state),
+    desks: state.desks && state.desks.length > 0 ? state.desks : [],
```
MarkLark86: Might be better to use a selector for this

client/components/PlanningAdvancedSearchForm/index.jsx
```diff
@@ -118,6 +128,8 @@ const mapStateToProps = (state) => ({
     endingDate: selector(state, 'dates.end'),
     g2_content_type: state.vocabularies.g2_content_type,
     daterange: selector(state, 'dates.range'),
+    desks: state.desks && state.desks.length > 0 ? state.desks : [],
```
MarkLark86: Might be better to use a selector for this

server/planning/assignments.py
```diff
@@ -0,0 +1,191 @@
...
+    'planning_item': planning_type,
+
+    'assigned_to': {
+        'type': 'dict',
+        'schema': {
+            'desk': {'type': 'string', 'nullable': True, 'mapping': not_analyzed},
+            'user': {'type': 'string', 'nullable': True, 'mapping': not_analyzed},
+            'assigned_by': {'type': 'string', 'mapping': not_analyzed},
+            'assigned_date': {'type': 'datetime'},
+            'state': {'type': 'string', 'mapping': not_analyzed},
```
MarkLark86: It would be good to set `allowed` to `assignment_state` for the state schema to enforce allowed values here

server/planning/coverage.py
```diff
@@ -16,36 +16,74 @@
...
 
 
 class CoverageService(superdesk.Service):
     """Service class for the coverage model."""
 
+    def on_fetched_item(self, doc):
+        self.enhance_items([doc])
+
+    def on_fetched(self, docs):
+        self.enhance_items([docs])
```
petrjasek: this should be probably `self.enhance_items(docs)`

server/planning/coverage.py
```diff
@@ -243,7 +311,18 @@ def _set_assignment_information(self, doc):
...
             'label': {'type': 'string'}
         }
-    }
+    },
+    'assigned_to': {
+        'type': 'dict',
+        'mapping': {
+            'type': 'object',
+            'properties': {
+                'assignment_id': not_analyzed,
```
petrjasek: is this supposed to be some sort of current assignment?

server/planning/coverage.py
mdhaman: rather it should be `self.enhance_items(docs[config.ITEMS])`

server/planning/coverage.py
mdhaman: It points to the assignment and on `GET` assignment information is added to the coverage.

client/components/EditAssignment/AssignmentSelect.jsx
```diff
@@ -114,10 +116,15 @@ export class CoverageAssignSelect extends React.Component {
             onChange: this.onDeskAssignChange.bind(this),
         }
 
-        return (<div className='coverageassignselect'>
-            <label>Assign</label>
-            { this.state.userAssigned && !this.state.deskAssigned &&
+        const { context } = this.props
+        const classes = classNames('coverageassignselect',
```
nrvikas: We can rename these classes to assignmentSelect ?

PR #468, opened 2017-09-27
[SDESK-1972] Clear on locks on planning entities after session end.

MarkLark86:


server/planning/item_lock.py
```diff
@@ -18,7 +18,7 @@
 from superdesk.lock import lock, unlock
 from eve.utils import config
 from superdesk import get_resource_service, get_resource_privileges
-from flask import current_app as app
+from apps.common.components.base_component import BaseComponent
```
nrvikas: Apologies for being too late, but, why are we getting the BaseComponent back again ? on_session_end was getting invoked, but there was a bug (passing item._id instead of item, which this PR also addresses) in unlock_session_for_resource why the items were not getting unlocked. I'm missing something probably.

nrvikas: Just checked that this won't work in case of 'honcho start'. My bad. All good with the code.

PR #500, opened 2017-10-19
Added "start working" button to create story from an assignment

jerome-poisson:
when a coverage is of "text" type, a new "start working" button is
displayed in assignments.
Clicking on it allows to select a template to create a story from the
assignment.

SDESK-96


jerome-poisson:
Wanted to put `WIP` label but it seems that I have not the right.

This P.R. is nearly working but I have still 2 issues to fix:
- data mergin doesn't seems to work as expected, after testing the `slugline` defined in the coverage is not assigned to the new story
- I'm not sure how to put a reference link to the planning item as requested in the issue, will need more details.

Feedbacks welcome.

mdhaman:
@jerome-poisson to link assignment to an planning item  you have to use the `assignment_link` endpoint. There is an action available in `assignments/api.js` https://github.com/superdesk/superdesk-planning/blob/master/client/actions/assignments/api.js#L158

jerome-poisson:
@mdhaman thanks for the feedback, I will check that

server/planning/assignments_content.py
```diff
@@ -40,6 +46,15 @@ def get_item_from_assignment(assignment):
...
     ednote = (assignment.get('planning') or {}).get('ednote')
 
+    planning_item = assignment.get('planning_item')
+    # we now merge planning data if they are set
+    if planning_item is not None:
+        planning = superdesk.get_resource_service('planning').find_one(req=None, _id=planning_item)
+        if planning is not None:
+            for field in FIELDS_TO_COPY:
+                if field in planning and planning[field]:
+                    item[field] = planning[field]
```
MarkLark86: When we're creating a Planning item from a News Item, we derive `planning.description_text` from either `item.abstract` or `item.headline` (whichever has a value).
I would expect when creating a News Item that `item.headline` would be derived from `planning.description_text`

client/services/PlanningStoreService.js
```diff
@@ -15,10 +15,13 @@ PlanningStoreService.$inject = [
     'notifyConnectionService',
     'userList',
     'desks',
+    'templates',
     'metadata',
     'session',
     'deployConfig',
     'gettext',
+    'content',
```
MarkLark86: Is `content` used in any actions? If not we can remove `content` from the dependencies.

jerome-poisson: indeed, was forgotten from a previous version, will update, thanks!

server/planning/assignments_content.py
jerome-poisson: OK I have changed it.

PR #514, opened 2017-10-31
[SDESK-2050] Edit Priority action

nrvikas:


client/actions/assignments/ui.js
```diff
@@ -188,6 +188,24 @@ const reassign = checkPermission(
...
+ */
+const _openEditPriorityModal = (assignment) => (
+    (dispatch) => dispatch(_openActionModal(
+        assignment,
+        ASSIGNMENTS.ITEM_ACTIONS.EDIT_PRIORITY.label,
+        'edit_priority'
+    ))
+)
+
+const editPriority = checkPermission(
```
MarkLark86: Maybe we should not use the `checkPermission` functionality but instead not show the action if there is no permission.
This removes the need for these extra dispatches.

client/components/ItemActionConfirmation/forms/editAssignmentPriorityForm.jsx
```diff
@@ -0,0 +1,74 @@
...
+        const {
+            handleSubmit,
+            initialValues,
+        } = this.props
+
+        return (<div className='ItemActionConfirmation' >
+                <form onSubmit={handleSubmit}>
+                    <div className="metadata-view">
+                        <dl>
+                            { initialValues.planning.slugline && (<dt>Slugline:</dt>) }
```
MarkLark86: The slugline's value is not being rendered

client/actions/assignments/ui.js
nrvikas: I have PR #509 where I cover privileges for actions. So, I'll take this up in that as I inject privileges as prop in that PR already.

PR #537, opened 2017-11-16
[SDESK-2035] Add Slack notifications

marwoodandrew:
…Users

### DIFF
--- a/server/planning/planning_notifications.py
+++ b/server/planning/planning_notifications.py
@@ -12,6 +12,16 @@
 import logging
 from superdesk.activity import add_activity, ACTIVITY_UPDATE
 import superdesk
+from flask import current_app as app
+from jinja2 import Template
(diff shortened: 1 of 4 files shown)


server/planning/planning_notifications.py
```diff
@@ -48,3 +61,172 @@ def notify_assignment(self, target_user=None, target_desk=None, target_desk2=Non
...
+        """
+        Superdesk user has been updated
+
+        If the update includes the Slack username then check that that user can be found in the Slack workspace
+
+        :param updates:
+        :param original:
+        :return:
+        """
+        if app.config.get('SLACK_BOT_TOKEN'):
```
MarkLark86: Might be good to check `slack_client_installed` here as well

marwoodandrew: Yep will do!

PR #544, opened 2017-11-21
ui refactoring

petrjasek:
putting together initial structure for the ui refactoring. also trying to be a bit more consistent regarding styling with superdesk core, so making some eslint config changes as well.

trying to reuse existing store for now, to get some test data, not sure yet how to approach that part anyhow.

- [x] new route for refactored ui - `/planning2` - not available via menu
- [x] defining top level components for layout (searchbar, filtersbar, main, preview, filters)
- [x] preview panel components
- [x] filters panel components
- [x] editor component
...


petrjasek:
@nrvikas @mdhaman @nrvikas does it make sense?

client/controllers/PlanningController.js
```diff
@@ -53,10 +53,12 @@ export function PlanningController(
                 store.dispatch(actions.resetStore())
             })
 
+            const App = $scope.vm.app || PlanningApp;
```
mdhaman: Why do we need to have `PlanningApp` component on scope?

mdhaman:
@petrjasek good that styling is consistent with `superdesk-core`. Not sure why it there was in first place.
Rest is all fine.

client/controllers/PlanningController.js
petrjasek: this is just temporary so I can use same controller for both old and new app

petrjasek:
I will leave it now, added missing layout components for filters panel, now it can be used to implement the functionality

PR #553, opened 2017-11-27
use shared code style

petrjasek:
I've put the eslint config into another repo, let me know if the rules are ok or you would prefer some changes first, then will start fixing the code.

https://github.com/superdesk/superdesk-client-code-style


mdhaman:
@petrjasek rules are fine. Can we have semi and indentation as warnings for now?

petrjasek:
I think these are fixable automatically, but will check

mdhaman:
yep. it automatically fixable.

mdhaman:
@petrjasek we can close this PR. I have cherry picked your changes.

PR #554, opened 2017-11-28
[SDESK-2120] Autosave feature for Event and Planning forms

MarkLark86:


petrjasek:
imo would be better to trigger the autosave from onChange handlers, especially if the autosave would eventually be done on server like in superdesk, so it's only done when needed.

MarkLark86:
@petrjasek 
We discussed this feature, and for Planning the autosave will only be in the browser stored in the redux store and not stored on the server. This is because the content being edited is not as extensive as the content from authoring. Later we can improve autosave by using `core.services.storage` (i.e. the Browser's localStorage functionality)

We also discussed using onChange functionality and decided that when we re-create the forms in the new UI we will do away with redux-form and use your form components along with their onChange functionality. In the meantime, autosave once every 3 seconds should suffice.

Our suggestion is we will enhance autosave once redux-form is removed from the form components.

PR #573, opened 2017-12-06
Added dependencies required for AAP build

nrvikas:


nrvikas:
Closing this as it is better to put it in superdesk-client-core repository.

PR #577, opened 2017-12-12
[SDESK-2204] Event preview with new UI

nrvikas:


client/components/Main/HistoryTab.jsx
```diff
@@ -1,23 +1,16 @@
...
-export const HistoryTab = () => {
-    const items = [
-        {
-            _id: 1,
-            author: 'Mika',
-            created: new Date(),
-        },
-    ];
+export const HistoryTab = ({item}) => (
+    <div>
```
MarkLark86: If the item type is `Event`, directly return the `EventHistory` component (removing the parent div node, as it is not required)

client/components/EventScheduleSummary/index.jsx
```diff
@@ -17,47 +20,35 @@ export const EventScheduleSummary = ({schedule}) => {
...
-    let startStr = moment(start).format('MMMM Do YYYY, h:mm:ss a');
-    let endStr = moment(end).format('MMMM Do YYYY, h:mm:ss a');
+    let eventDateText;
+
+    if (start.isSame(end, 'day')) {
+        eventDateText = start.format('DD/MM/YYYY @ HH:mma') + ' - ' +
+            end.format('HH:mm a');
+    } else {
+        eventDateText = start.format('DD/MM/YYYY @ HH:mma') + ' - ' +
+            end.format('DD/MM/YYYY @ HH:mm a');
```
MarkLark86: It would be better to use `config.view.dateformat` and `config.view.timeformat` to better handle internationalization with date/time strings (also the extra `a` in the format string is not required, as the hour format here is 24 hour)

client/components/Main/PreviewHeader.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {EventPreviewHeader} from '../Events';
+import {getItemType} from '../../utils';
+
+export const PreviewHeader = ({item}) => (
+    <div>
```
MarkLark86: If the item type is `Event`, directly return the `EventPreviewHeader` component (removing the parent div node, as it is not required)

client/components/Events/EventHistory.jsx
```diff
@@ -0,0 +1,163 @@
...
+        this.props.openEventPreview(duplicateId);
+    }
+
+    render() {
+        const {users} = this.props;
+        const displayUser = (recievedUserId) => get(getItemInArrayById(users, recievedUserId), 'display_name');
+
+        return (
+            <div>
+                <ul className="history-list">
```
MarkLark86: `<ul>` should be the parent component (removing the parent div node, as it is not required).

Also the `history-list` class is adding `padding: 10px` which is not required since `side-panel__content-block` is providing the padding for us

PR #585, opened 2018-01-04
[SDESK-2213] Manage Agendas new UI

nrvikas:


client/components/Agendas/AgendaItem.jsx
```diff
@@ -0,0 +1,37 @@
...
+    <List.Item shadow={1}>
+        <List.Border state={active ? 'active' : 'idle'} />
+        <List.Column grow={true} border={false}>
+            <List.Row>
+                <span className="sd-overflow-ellipsis sd-list-item--element-grow">{agenda.name}</span>
+                <time>updated {moment(agenda._updated).fromNow()}</time>
+            </List.Row>
+        </List.Column>
+        {privileges.planning_agenda_management === 1 &&
+            <div className="sd-list-item__action-menu sd-list-item__action-menu--direction-row">
```
MarkLark86: There is a `List.ActionMenu` component that you could use here, and adding a prop to control the `sd-list-item__action-menu--direction-row` CSS class

client/components/UI/SlideInPanel/Content.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Content = ({children}) => (
+    <div className="sd-slide-in-panel__content">
+        <div className="sd-slide-in-panel__content-block sd-slide-in-panel__content-block--first">
```
MarkLark86: BEM modifiers should probably be a property of the Component.
i.e. `sd-slide-in-panel__content-block--first` should be controlled via a `first` boolean property

PR #593, opened 2018-01-15
[SDESK-2310] Update metadata of recurring event confirmation

nrvikas:


client/components/ItemActionConfirmation/forms/updateRecurringEventsForm.jsx
```diff
@@ -1,84 +1,122 @@
...
+                    schedule={initialValues.dates}
+                    timeFormat={timeFormat}
+                    dateFormat={dateFormat}
+                />
+
+                {isRecurring && (
+                    <Row
+                        label={gettext('No. of Events')}
+                        value={numEvents}
+                        className="form__row--no-padding"
```
MarkLark86: There is a `noPadding` prop in the `Preview.Row` component now

client/components/ModalWithForm/index.jsx
```diff
@@ -24,7 +27,11 @@ export class Component extends React.Component {
     }
 
     submit() {
-        this.getFormInstance().submit();
+        this.setState({submitting: true});
+        this.getFormInstance().submit()
```
MarkLark86: In the `render` function, can we change `ref: 'form'` to not use strings ref but a function instead.

i.e. `ref: (node) => this.dom.form = node` and update all `this.refs.form` to be `this.dom.form` to remove deprecated string refs.

PR #600, opened 2018-01-17
[SDESK-2103] Combined view for events and planning.

mdhaman:


mdhaman:
@fritzSF couple styles need modification:
1) https://github.com/superdesk/superdesk-ui-framework/blob/master/app/styles/components/_list-item.scss#L390 
Can you please remove `max-height` as there could be more than 2 related planning items for an event?

2) https://github.com/superdesk/superdesk-ui-framework/blob/master/app/styles/components/_list-item.scss#L364
You can't click on the items below the event once you show/hide related plannings. Maybe `z-index` can be handled in `collapsed` and `expanded` styles.

PR #602, opened 2018-01-23
[SDESK-2303] Add to planning ported to new UI design

nrvikas:


client/components/Planning/PlanningEditor/index.jsx
```diff
@@ -20,20 +20,87 @@ import {ToggleBox} from '../../UI';
...
         super(props);
 
         this.dom = {slugline: null};
         this.onChange = this.onChange.bind(this);
+        this.createNewPlanningFromNewsItem = this.createNewPlanningFromNewsItem.bind(this);
+    }
+
+    componentWillMount() {
+        // If we are creating a new planning item for 'add-to-planning'
+        if (this.props.addNewsItemToPlanning) {
```
mdhaman: Can you simplify the `if` condition over here i.e. remove the nesting.

client/components/ItemActionConfirmation/forms/updateAssignmentForm.jsx
```diff
@@ -72,7 +72,7 @@ class Component extends React.Component {
             {
                 key: 'Slugline:',
                 value: (
-                    <span className="ItemActionConfirmation__metadata__slugline">
+                    <span className="MetadataView__metadata__slugline">
```
mdhaman: We can keep `Metadataview` css class. But do we need other classes like `MetadataView__metadata__slugline`?

nrvikas: I had kept it as a part to do when we refactor updateAssignments. But have changed it now to use <preview/row> component instead and taken out MetadataView__metadata__slugline.

client/components/Planning/PlanningEditor/index.jsx
nrvikas: I think each nest of 'if' statement is required here because it meets those conditions. This will however change when I try to resolve merge conflicts during coverage item actions in another PR.

nrvikas:
@MarkLark86 , I have updated the PR with scroll ability.

client/components/Planning/PlanningEditor/index.jsx
MarkLark86:
```
if (!this.props.addNewsItemToPlanning) {
    return
}
if (!get(this.props, 'item._id')) {
    const newPlanning = this.createNewPlanningFromNewsItem();
    this.props.onChangeHandler(null, newPlanning);
} else if (get(this.props, 'item._addNewCoverage')) {
    // As per current logic
}
```

nrvikas: Per my comment, I meant to update it in another PR when I resolve the mrege conflicts as this file will change again during rebase - not that it is not doable or should not be done. Anyways, I will update it here. Cheers!

PR #684, opened 2018-03-02
[SDESK-2443] Published Events and planning items to be republished after actions

nrvikas:


server/features/planning_cancel.feature
```diff
@@ -382,6 +455,117 @@ Feature: Cancel all coverage
       }
       """
 
+
+    @auth
+    @notification
+    Scenario:
```
mdhaman: Scenario definition missing.

PR #712, opened 2018-03-12
[SDESK-2502] - Some tooltips needed in the new planning and assignments.

sjunaid:


client/components/AssignmentItem/index.jsx
```diff
@@ -87,21 +98,26 @@ export const AssignmentItem = ({
...
+                        </span>
                         <InternalNoteLabel
                             item={assignment}
                             prefix="planning."
                             marginRight={false}
                         />
-                        {get(assignment, 'planning.scheduled') ? (
-                            <AbsoluteDate date={get(assignment, 'planning.scheduled').toString()} />
-                        ) : (<time><span>{gettext('\'not scheduled yet\'')}</span></time>)}
+                        <span title={gettext('Due Date')}>
```
nrvikas: We are showing the exact same date in the tooltip above. Is it expected to show the same information again next to each other ?

client/components/ItemIcon.jsx
```diff
@@ -67,20 +69,25 @@ export const ItemIcon = ({item, big, white, blue, showRepeating}) => {
             return null;
         }
 
+
         if (!showRepeating || !multiValidator(item)) {
+            title = gettext(title);
+
             return big ?
-                (<span className="double-size-icn double-size-icn--light">
+                (<span title={title} className="double-size-icn double-size-icn--light">
```
nrvikas: Some details are displayed as a 'span' with a white background where as some are ui-framework's tooltips. Should we be having consistent behavior across the entire module ?

PR #718, opened 2018-03-14
[SDESK-2530] Merge planning_types schema in database with default schema

nrvikas:


server/planning/planning/planning_types.py
```diff
@@ -217,15 +218,24 @@ def find_one(self, req, **lookup):
 
     def get(self, req, lookup):
         planning_types = list(super().get(req, lookup))
+        merged_planning_types = []
```
MarkLark86: This functionality will also be required in the `find_one` method (as this is used for publish validation)

PR #726, opened 2018-03-16
(fix) - Contact form modal related overriding styles issue

sjunaid:


client/components/fields/SelectSearchTermsField/style.scss
```diff
@@ -153,62 +153,70 @@
...
-        bottom: 15%;
-        left: 40%;
-        border: 1px solid #fff;
-    }
-}
 
-.header-info {
-    background-color: rgba(94, 169, 200, 0.11);
-    padding: 10px 0 0;
+    .header-info {
```
mdhaman: what about the header-info class? can we use BEM.

PR #727, opened 2018-03-16
[SDESK-2476] New Planning item action 'Add as Event'

MarkLark86:


client/actions/events/api.js
```diff
@@ -1101,46 +1101,59 @@ const _saveLocation = (event) => (
...
+            return api('events').save(original, updates);
+        })
+        .then((data) => {
+            // return a list of events (can have several because of reccurence)
+            const events = get(data, '_items') || [data];
+
+            // If this Event has been created from a Planning item
+            // then link the new Event to it
+            if (!get(eventUpdates, '_id') && get(eventUpdates, '_planning_item') && get(events, '[0]._id')) {
+                return dispatch(planningApi.linkToEvent(eventUpdates._planning_item, events[0]._id))
```
nrvikas: Would it be better to achieve this in the backend when we create the event instead of another call from planning api?

MarkLark86: Good call, I'll make the changes

PR #739, opened 2018-03-23
[SDESK-2533] - Update the assignment count once an assignment was rem…

sjunaid:
…oved

Also covered [SDESK-1973](https://dev.sourcefabric.org/browse/SDESK-1973)


client/actions/assignments/ui.js
```diff
@@ -30,6 +30,26 @@ const loadAssignments = (
     return dispatch(queryAndSetAssignmentListGroups(filterByState));
 };
 
+const myAssignments = (
```
MarkLark86: This action is redundant considering all it does is call `queryAndGetMyAssignments`. Consider replacing all usages of `myAssignments` with `queryAndGetMyAssignments`

client/actions/assignments/ui.js
```diff
@@ -30,6 +30,26 @@ const loadAssignments = (
...
+    filterByState = null
+) => (dispatch) => dispatch(queryAndGetMyAssignments(filterByState));
+
+const queryAndGetMyAssignments = (filterByState) => (
+    (dispatch, getState) => {
+        let querySearchSettings = selectors.getAssignmentSearch(getState());
```
MarkLark86: Whenever we modify the response from a call to `selector.` you must cloneDeep the object returned. Otherwise any modifications to the object will be reflected on other calls to the `selector.` function

client/components/AssignmentListContainer/index.jsx
```diff
@@ -193,6 +215,7 @@ const mapDispatchToProps = (dispatch) => ({
             filterBy, searchQuery, orderByField, orderDirection, filterByState,
             filterByType, filterByPriority
         )),
+    fetchMyAssignmentsCount: (filterByState) => dispatch(actions.assignments.ui.myAssignments(filterByState)),
```
MarkLark86: This dispatch could be changed to `() => dispatch(actions.assignments.ui.myAssignments(['assigned', 'submitted']))` and remove `getMyAssignmentsCount` method from the class and use `this.props.fetchMyAssignmentsCount` instead

client/actions/assignments/notifications.js
```diff
@@ -43,12 +51,16 @@ const onAssignmentUpdated = (_e, data) => (
             return;
         }
 
-        if (currentDesk === data.assigned_desk || currentDesk === data.original_assigned_desk) {
+        // Updates my assignments count.
+        dispatch(assignments.ui.myAssignments(['assigned', 'submitted']));
```
MarkLark86: Consider using `ASSIGNMENTS.WORKFLOW_STATE.ASSIGNED` and `ASSIGNMENTS.WORKFLOW_STATE.SUBMITTED` throughout this PR

client/actions/assignments/notifications.js
```diff
@@ -15,15 +15,21 @@ const onAssignmentCreated = (_e, data) => (
     (dispatch, getState) => {
         const currentDesk = selectors.getCurrentDeskId(getState());
 
-        if (currentDesk &&
+        let querySearchSettings = selectors.getAssignmentSearch(getState());
+
+        // Updates my assignment count
+        dispatch(assignments.ui.myAssignments(['assigned', 'submitted']));
```
mdhaman: use defined constants.

PR #741, opened 2018-03-26
[SDANSA-129] Send email with assignment to an assignee

pablopunk:
Same workflow as slack notifications but with emails. Checks for user preferences on email notifications before sending anything.


pablopunk:
This should be ready to review

pablopunk:
@mdhaman right, I just extended what was done with Slack but I agree that we should have a template for the email

pablopunk:
BTW I don't have merge permissions here, if someone can merge it that would be great.

PR #810, opened 2018-05-01
[SDESK-2829] Add tooltips at the top of the event and planning edit p…

marwoodandrew:
…anels

### DIFF
--- a/client/components/ItemActionsMenu/index.jsx
+++ b/client/components/ItemActionsMenu/index.jsx
@@ -46,7 +46,7 @@ export class ItemActionsMenu extends React.Component {
         }
 
         return (
-            <button className={className} onClick={this.toggleMenu}>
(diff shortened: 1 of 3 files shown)


client/components/ItemActionsMenu/index.jsx
```diff
@@ -46,7 +46,7 @@ export class ItemActionsMenu extends React.Component {
         }
 
         return (
-            <button className={className} onClick={this.toggleMenu}>
+            <button className={className} onClick={this.toggleMenu} title={gettext(TOOLTIPS.actions)}>
```
MarkLark86: Using `gettext` with a variable will not provide translations to `.po` files.
Use `gettext` with a **static string** only

marwoodandrew: Done, but left the constants might be useful!

PR #824, opened 2018-05-14
[SDESK-2770] Change 'Update' to 'Save & Post' in Ignore/Cancel modal if the item is unposted (killed)

MarkLark86:


client/components/IgnoreCancelSaveModal.jsx
```diff
@@ -103,30 +108,41 @@ export class IgnoreCancelSaveModalComponent extends React.Component {
         );
     }
 
+    getOkText() {
+        const {item, onGoTo, onSaveAndPost} = this.props.modalProps || {};
+        let okText;
+
+        if (onGoTo) {
+            okText = gettext('GoTo');
```
nrvikas: Need a space between, 'Go' and 'To' when used in gettext() 😃

MarkLark86: Changed it to 'Go-To' as this is the wording used in Authoring

PR #826, opened 2018-05-15
[SDESK-2903] - Todo and Inprogress count of Assignments are increasin…

sjunaid:
…g when User scroll to bottom


client/actions/assignments/ui.js
```diff
@@ -339,6 +331,18 @@ const setAssignmentsTodoList = (ids, totalNoOfItems) => ({
     },
 });
 
+/**
+ * Action that adds the list of assignments items in to-do state
+ * @param {Array} ids - An array of assignments item ids
+ */
+const addAssignmentsTodoList = (ids, totalNoOfItems) => ({
```
mdhaman: you don't need this extra actions. You can pass the payload from `addToAssignmentListGroup` action.

PR #837, opened 2018-05-18
Add missing client configurations

ride90:
Check if suggested client configuration works for you.


README.md
```diff
@@ -28,9 +28,18 @@ apps: [
     ....,
     'superdesk-planning'
 ],
+importApps: [
```
petrjasek: this shouldn't be needed if you use apps

PR #841, opened 2018-05-23
[SDESK-2895] - If the user does not have Event Management privilege t…

sjunaid:
…hen User cannot edit event and create event


client/components/Main/CreateNewSubnavDropdown.jsx
```diff
@@ -21,20 +24,21 @@ export const CreateNewSubnavDropdown = ({addEvent, addPlanning, createPlanningOn
     }
 
     return (
-        <Dropdown
+        <div>{items.length > 0 && <Dropdown
```
nrvikas: You can return null instead of a an empty <div> if items.length === 0

PR #861, opened 2018-05-30
fix(cvs): fix custom cvs for ntb

petrjasek:
- use `ANPA Category` label to be consistent with superdesk
- set scheme for custom subject so it will show up there in authoring

SDNTB-492

### DIFF
--- a/client/services/PlanningStoreService.js
+++ b/client/services/PlanningStoreService.js
@@ -1,4 +1,4 @@
-import {isNil, zipObject, get} from 'lodash';
+import {isNil, zipObject, get, isEmpty} from 'lodash';
 import {createStore} from '../utils';
 import {ITEM_TYPE} from '../constants';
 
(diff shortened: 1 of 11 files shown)


client/services/PlanningStoreService.js
```diff
@@ -221,3 +220,11 @@ export function PlanningStoreService(
 
     $rootScope.$on('vocabularies:updated', angular.bind(this, this._reloadVocabularies));
 }
+
+function getSubjects(values) {
+    if (values.subject_custom) {
+        return values.subject_custom.map((item) => Object.assign({}, item, {scheme: 'subject_custom'}));
+    }
+
+    return values.subject;
```
MarkLark86: Should this be:
`return values.subjectcodes`

client/utils/index.js
```diff
@@ -217,6 +217,7 @@ export const createStore = (params = {}, app = planningApp) => {
     ];
     // return the store
 
+    console.info('state', initialState);
```
ioanpocol: Do we need this log?

petrjasek: I don't think so :+1:

petrjasek:
seems like there is some issue with genre not being saved atm, but that's also on sdp-master, @MarkLark86 is that known issue?

MarkLark86:
@petrjasek The latest Autosave improvement (save autosave on the server) has been causing multiple issues. We're in the progress of investigating these issues and will be creating tickets shortly

PR #873, opened 2018-06-07
[REDUX] Activate redux dev tools and redux logger for non production instances

ride90:
Hi @MarkLark86 ,
I've added [redux dev tools](https://github.com/zalmoxisus/redux-devtools-extension) (will be activated if it's installed in the browser) for non production instances. 
I think it also makes sense to apply redux logger only for non prod instances. 
Let me know if it works for you.


MarkLark86:
Hi @ride90,

Looks fine, tested in the test instance and my local instance.

Keep in mind that Redux Time Travel may not always work in our case (as not all data in the store is able to be serialized, such as the `'SHOW_MODAL'` action)

PR #889, opened 2018-06-18
fix(gettext): fix missing translations

petrjasek:
SDESK-3040

### DIFF
--- a/client/components/Assignments/AssignmentGroupList.jsx
+++ b/client/components/Assignments/AssignmentGroupList.jsx
@@ -6,7 +6,8 @@ import {get} from 'lodash';
 import {ASSIGNMENTS, UI} from '../../constants';
 import * as selectors from '../../selectors';
 import * as actions from '../../actions';
-import {assignmentUtils, gettext} from '../../utils';
+import {assignmentUtils} from '../../utils';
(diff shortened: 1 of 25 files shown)


client/components/Assignments/AssignmentGroupList.jsx
```diff
@@ -119,7 +120,7 @@ class AssignmentGroupListComponent extends React.Component {
                             className="sd-list-header__name sd-list-header__name--cursorPointer"
                             onClick={this.changeAssignmentListSingleGroupView}
                         >
-                            <span>{ASSIGNMENTS.LIST_GROUPS[groupKey].label}</span>
+                            <span>{gettext(ASSIGNMENTS.LIST_GROUPS[groupKey].label)}</span>
```
mdhaman: `gettext` is already called here https://github.com/superdesk/superdesk-planning/pull/889/files#diff-c90901640fd209305b37428d71c2dcafR68
Do we need to call `gettext` twice?

petrjasek: yes, problem is that when it's defined on top level as a constant it won't have the translations loaded yet so it will just return the same string. so the translation will only happen when rendering. but we have to keep the gettext there for string extraction

PR #894, opened 2018-06-19
fix(ntb): fix ntb event xml content

petrjasek:
- only use definition_short for content field in output

SDNTB-500

### DIFF
--- a/server/planning/output_formatters/ntb_event.py
+++ b/server/planning/output_formatters/ntb_event.py
@@ -50,7 +50,7 @@ class NTBEventFormatter(Formatter):
         priority = etree.SubElement(doc, 'priority')
         priority.text = str(item.get('priority', self.PRIORITY))
         content = etree.SubElement(doc, 'content')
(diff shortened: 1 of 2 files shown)


server/planning/output_formatters/ntb_event.py
```diff
@@ -50,7 +50,7 @@ def _format_doc(self, doc, item):
         priority = etree.SubElement(doc, 'priority')
         priority.text = str(item.get('priority', self.PRIORITY))
         content = etree.SubElement(doc, 'content')
-        content.text = item.get('definition_long', item.get('definition_short'))
+        content.text = item.get('definition_short', '')
```
MarkLark86: Because **definition_short/definition_long** can be enabled/disabled using **planning_types**, would it be better to use the following:
`content.text = item.get('definition_sort') or item.get('definition_long) or ''`

petrjasek: afaik they don't want to ever export definition_long, it's kind of internal info for ntb

PR #897, opened 2018-06-20
[SDESK-3034] Item actions were not opening if item was open in the editor

nrvikas:


client/utils/index.js
```diff
@@ -759,10 +760,9 @@ export const generateTempId = () => TEMP_ID_PREFIX + moment().valueOf();
  */
 export const removeAutosaveFields = (item, stripLockFields = false) => {
     const fieldsToKeep = ['_id', '_planning_item'];
-    const fieldsToIgnore = ['planning_ids', 'reason', 'update_method'];
 
     if (stripLockFields) {
-        fieldsToIgnore.concat([
+        AUTOSAVE.IGNORE_FIELDS.concat([
```
MarkLark86: This statement has no effect, as `Array.concat` returns a new array.

server/planning/autosave.py
```diff
@@ -20,6 +20,7 @@ class AutosaveService(Service):
     def on_create(self, docs):
         for doc in docs:
             AutosaveService._validate(doc)
+            doc.pop('expired', None)
```
MarkLark86: This is already handled in `AUTOSAVE.IGNORE_FIELDS`.

It would be best for the endpoint to simply save all the supplied data, leaving the client in control of what gets saved.

nrvikas: 'expired' is a default field is 'events' and 'planning' schema, so it comes up at the server side.

PR #899, opened 2018-06-20
fix(planning): remove planning name from editor/preview

petrjasek:
and add headline instead. there should be no name in planning item,
but we need headline for export.

### DIFF
--- a/client/components/Planning/PlanningEditor/index.jsx
+++ b/client/components/Planning/PlanningEditor/index.jsx
@@ -369,8 +369,8 @@ export class PlanningEditorComponent extends React.Component {
 
                     <Field
                         component={TextInput}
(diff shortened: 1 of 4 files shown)


petrjasek:
I was just wondering if there is any usecase for such field in planning, I don't think it would make it to any output, or would it? @MarkLark86 @mdhaman

mdhaman:
We had `headline` and it was confusing for our users as well as vince/karel. We kept `headline` in the schema if somebody needs `headline`. Moreover, if you use `headline` then we won't be able to change the label to `name`.

petrjasek:
I see. will do it like mark suggested then

petrjasek:
will do the changes in different PR

PR #906, opened 2018-06-21
add headline to planning editor

petrjasek:


mdhaman:
@petrjasek `headline` field is missing in the `planning preview` and `assignment preview`.

PR #910, opened 2018-06-22
[SDESK-3071] Add-as-event was holding planning lock if creating new event was cancelled in the form

nrvikas:


MarkLark86:
When attempting to create the Event from a Planning item (as per the ticket), if you click `CREATE` instead of `IGNORE`, the Event is created, the Form Editor closes, but the lock remains

PR #911, opened 2018-06-22
fix(planning): populate headline using event name

petrjasek:
when creating planning item for an event.

### DIFF
--- a/server/features/planning.feature
+++ b/server/features/planning.feature
@@ -1546,3 +1546,105 @@ Feature: Planning
             }
         }]
         """
+
+    @auth
+    Scenario: Headline is not populated when creating item from event using default type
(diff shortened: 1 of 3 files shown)


MarkLark86:
There is a function that we use for this called `createPlanningFromEvent` in the file `client/actions/agenda.js:217`.

Ideally this function should use the `planning_types` to populate the new Planning item from the Event.

petrjasek:
imo such logic is better done on the server, when there will be mobile client or something it would have to duplicate it. so what about moving it there?

mdhaman:
agree with you, we need an endpoint to create item from `planning_types`.

petrjasek:
using `planning_type` now to check if headline is enabled

PR #914, opened 2018-06-25
fix(feed) - Added missing parameters

akintolga:
Missing parameters causes the following error: `TypeError: register_feeding_service() missing 2 required positional arguments: 'service_class' and 'errors'`


akintolga:
Closing this as I was using an earlier version of `superdesk-core` in newsroom which was causing this problem locally, bumping the version to 1.21 resolved the issue...

PR #916, opened 2018-06-29
[SDESK-2958] Save versions of published planning items and asyn call …

marwoodandrew:
…to enqueue to publish

### DIFF
--- a/server/planning/planning/planning_post.py
+++ b/server/planning/planning/planning_post.py
@@ -9,16 +9,15 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 from flask import abort
-from superdesk import get_resource_service
+from superdesk import get_resource_service, logger
 from superdesk.resource import Resource
 from superdesk.services import BaseService
 from superdesk.notification import push_notification
-from apps.publish.enqueue import get_enqueue_service
-
 
 from eve.utils import config
 from planning.planning import PlanningResource
-from planning.common import WORKFLOW_STATE, POST_STATE, post_state, get_item_post_state
+from planning.common import WORKFLOW_STATE, POST_STATE, post_state, get_item_post_state, enqueue_planning_item
+from datetime import datetime
 
 
 class PlanningPostResource(PlanningResource):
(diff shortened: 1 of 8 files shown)


server/planning/planning/planning_post.py
```diff
@@ -97,12 +98,20 @@ def post_planning(self, plan, new_post_state):
         get_resource_service('planning').update(plan['_id'], updates, plan)
         get_resource_service('planning_history')._save_history(plan, updates, 'post')
 
-        # enqueue the planning item
-        # these fields are set for enqueue process to work. otherwise not needed
-        plan.setdefault(config.VERSION, 1)
+        # Set a version number
+        version = int((datetime.utcnow() - datetime.min).total_seconds() * 100000.0)
```
mdhaman: have `version` in the history so that we can show the delivery in the history.

server/planning/planning/planning_post.py
```diff
@@ -113,3 +122,21 @@ def _get_post_state(self, plan, new_post_state):
             return WORKFLOW_STATE.SCHEDULED
 
         return plan.get('state')
+
+    @staticmethod
+    @celery.task(soft_time_limit=600)
+    def _enqueue_planning(id):
```
mdhaman: `_enqueue_planning` and `_enqueue_event`  can be just one method `enqueue_item`. There is `type` store in the collection.

server/planning/planning_published_versions.py
```diff
@@ -0,0 +1,54 @@
...
+        'type': {
+            'type': 'string'
+        },
+        # The item as published
+        'published_item': {
+            'type': 'dict'
+        }
+    }
+
+    privileges = {'POST': 'archive'}
```
mdhaman: no privileges are required.
and following.
item_methods = ['GET']
RESOURCE_METHODS = ['GET']

marwoodandrew: OK

server/planning/planning/planning_post.py
marwoodandrew: I'll move it to common.py

server/planning/planning/planning_post.py
marwoodandrew: Yep

PR #933, opened 2018-07-16
[SDESK-3008] Featured Stories Modal

nrvikas:


client/actions/planning/api.js
```diff
@@ -868,6 +888,29 @@ const lockFeaturedPlanning = () => (
...
+ * @return Promise
+ */
+const saveFeaturedPlanning = (updates) => (
+    (dispatch, getState, {api}) => {
+        const item = selectors.featuredPlanning.featuredPlanningItem(getState()) || {};
+
+        return api('planning_featured').save(cloneDeep(item), {
+            ...(pickBy(cloneDeep(item), (v, k) => (!k.startsWith('_')))),
+            ...updates,
+            tz: moment.tz.guess(),
```
MarkLark86: I believe this should use the configured default timezone from the config and not the client browser's timezone.
i.e. `config.defaultTimezone`

client/components/Planning/FeaturedPlanning/FeaturedPlanningSelectedList.jsx
```diff
@@ -0,0 +1,72 @@
...
+        desks={desks}
+        users={users}
+        onRemoveFromSelectedFeaturedPlanning={onRemoveFromSelectedFeaturedPlanning}
+        selectedPlanningIds={selectedPlanningIds}
+        hideItemActions
+        onSortEnd={onSortEnd}
+        onSortStart={onSortStart}
+        sortable={!readOnly}
+        leftBorder={leftBorder}
+        header={gettext(`Featured stories for ${currentSearchDate.format('dddd LL')}`)}
```
MarkLark86: This should use gettext with params instead of injecting the value in the string that is passed to gettext.
i.e. `getttext('Featured stores for {{ date }}', {date: currentSearchDate.format('dddd LL')})`

client/components/Planning/FeaturedPlanning/FeaturedPlanningModal.jsx
```diff
@@ -1,46 +1,375 @@
...
+            unselectedPlanningIds: difference(props.featuredPlanningItems.map((i) => i._id),
+                get(props, 'featuredPlanningItem.items', [])),
+            selectedPlanningIds: get(props, 'featuredPlanningItem.items', []),
+            dirty: false,
+            highlights: [],
+        };
+    }
+
+    onDateChange(date) {
+        this.props.loadFeaturedPlanningsData(date);
```
MarkLark86: Make sure to use the configured defaultTimezone when selecting the date to show, as the date could change based on the client browsers timezone. I figure that this entire feature will be based on `config.defaultTimezone`.

It might be good if `<DateInputPopup>` had a timezone attribute that could control this.

PR #947, opened 2018-08-02
[SDESK-3158][SDESK-2501] Bugs related to spiking item while editing

nrvikas:


server/planning/common.py
```diff
@@ -220,6 +220,17 @@ def get_coverage_type_name(qcode):
     return coverage_type.get('name', qcode)
 
 
+def remove_autosave_on_spike(item):
+    if item.get('lock_action') == 'edit':
+        autosave_service = get_resource_service('planning_autosave')
+        if item.get('type') == 'event':
+            autosave_service = get_resource_service('event_autosave')
+
+        autosaved_item = autosave_service.find_one(None, _id=item.get(config.ID_FIELD))
```
mdhaman: No need for `find_one`. The `delete_action` already does that.

server/planning/common.py
```diff
@@ -220,6 +220,17 @@ def get_coverage_type_name(qcode):
     return coverage_type.get('name', qcode)
 
 
+def remove_autosave_on_spike(item):
```
mdhaman: We now have ignore/cancel/save and `openactionfromeditor` unlocks the item. Do we still need this method on the server?

nrvikas: Only for those during bulk select and spike. Because as per monitoring behavior there are no prompts to 'ignore-cancel-save' in this scenario. And only in this bulk-spike mode 'original' will have the lock information as 'edit'. Hence that check in the function (remove_autosave_on_spike) to avoid redundant delete action. Otherwise, it will have to send unlock for all those bulk-selected items which are on work-queue from front end (thought that would be too many promises to be resolved when doing it in backend would be simple).

PR #974, opened 2018-08-23
[SDESK-3157] Autosave error during spike

nrvikas:


client/actions/main.js
```diff
@@ -601,6 +600,21 @@ const closePreviewAndEditorForItems = (items, actionMessage = '', field = '_id')
             }
         }
 
+        const itemType = (get(items, `[0]${field}`) in selectors.planning.storedPlannings(getState())) ?
+            ITEM_TYPE.PLANNING : ITEM_TYPE.EVENT;
+
+        items.forEach((item) => {
+            const autoSaves = selectors.forms.autosaves(getState());
+            let autoSaveInStore = get(autoSaves, `${itemType}['${item[field]}']`);
```
MarkLark86: It might be better to use the common utility function to get the autosave item.
From client/utils/index.js:
`export const getAutosaveItem = (autosaves, itemType, itemId) => (`

Also might be better to get the item type inside the forEach loop, rather than assume every item passed in are of the same type

PR #975, opened 2018-08-23
[SDESK-3278] Autosave files

nrvikas:


server/planning/autosave.py
```diff
@@ -22,6 +27,10 @@ def on_create(self, docs):
             AutosaveService._validate(doc)
             doc.pop('expired', None)
 
+    def on_delete(self, doc):
+        if doc.get(ITEM_TYPE) == 'event' and doc.get(config.ID_FIELD).startswith(TEMP_ID_PREFIX):
```
mdhaman: what about existing event and the user adds a file and then decides to ignore changes?

nrvikas: Valid scenario. Will update the code.

PR #1001, opened 2018-09-13
[SDESK-3343](fix): Planning list not being updated when Agenda name is changed

MarkLark86:


client/components/Planning/PlanningItem.jsx
```diff
@@ -42,7 +42,9 @@ export class PlanningItem extends React.Component {
     }
 
     shouldComponentUpdate(nextProps, nextState) {
-        return isItemDifferent(this.props, nextProps) || this.state.hover !== nextState.hover;
+        return isItemDifferent(this.props, nextProps) ||
+            this.state.hover !== nextState.hover ||
+            !isEqual(this.getAgendaNames(this.props), this.getAgendaNames(nextProps));
```
mdhaman: What about calendars?

PR #1010, opened 2018-09-21
Put avatar at the top of assignment preview instead of the left so it fits in the right pane

tomaskikutis:
REQUIRED FOR: SDESK-3432


mdhaman:
Hi @tomaskikutis / @fritzSF, 
As the 'task details' pane is not wide enough, how about removing the avatar for task details pane.

PR #1023, opened 2018-10-09
[SDESK-3453] AAP: Hide the Slugline on the Event form

nrvikas:


client/components/Events/EventEditor/index.jsx
```diff
@@ -60,7 +60,7 @@ export class EventEditorComponent extends React.Component {
     }
 
     componentDidMount() {
-        if (!get(this.props, 'navigation.scrollToViewItem')) {
+        if (!get(this.props, 'navigation.scrollToViewItem') && this.dom.slugline) {
```
MarkLark86: If the form does not have a **slugline** input field, then we should focus the **name** input field.

client/components/Events/EventEditor/index.jsx
```diff
@@ -76,8 +76,8 @@ export class EventEditorComponent extends React.Component {
         const currentItemId = getItemId(this.props.item);
 
         // If item changed or it got locked for editing
-        if ((prevItemId !== currentItemId) ||
-            (!get(prevProps, 'diff.lock_user') && get(this.props, 'diff.lock_user'))) {
+        if (((prevItemId !== currentItemId) ||
+            (!get(prevProps, 'diff.lock_user') && get(this.props, 'diff.lock_user'))) && this.dom.slugline) {
```
MarkLark86: Same goes here, focus **name** input field is **slugline** is not on the form

PR #1025, opened 2018-10-11
[SDESK-3279] Add planning related types into 'type' vocabulary and add config for planning related celery tasks.

nrvikas:


server/planning/__init__.py
```diff
@@ -114,6 +118,49 @@ def init_app(app):
     app.client_config['street_map_url'] = get_street_map_url(app)
     app.client_config['max_multi_day_event_duration'] = get_event_max_multi_day_duration(app)
 
+    # Set up Celery task options
+    if not app.config.get('CELERY_TASK_ROUTES'):
+        app.config['CELERY_TASK_ROUTES'] = CTR
+
+    if not app.config.get('CELERY_TASK_ROUTES').get('planning.flag_expired'):
```
MarkLark86: if `app.config['PLANNING_EXPIRY_MINUTES'] == 0`, then the command `planning:flag_expired` does nothing.

It might be good here to not add it to the message queue.

server/planning/__init__.py
```diff
@@ -114,6 +118,49 @@ def init_app(app):
...
+            added_types = []
+            type_names = [t['name'] for t in items]
+
+            planning_type_list = [
+                {"is_active": True, "name": "planning item", "qcode": "planning"},
+                {"is_active": True, "name": "event", "qcode": "event"},
+                {"is_active": True, "name": "Featured Stories", "qcode": "planning_featured"}
+            ]
+
+            for type in planning_type_list:
```
MarkLark86: `type` is an in-built function in python and should be avoided for class/variable names.

server/planning/__init__.py
MarkLark86: if `app.config['PLANNING_EXPIRY_MINUTES'] == 0`, then the command `planning:flag_expired` does nothing.

It might be good here to not add it to the message queue.

server/planning/__init__.py
MarkLark86: `type` is an in-built function in python and should be avoided for class/variable names.

server/planning/__init__.py
```diff
@@ -114,6 +118,49 @@ def init_app(app):
...
+                if type['name'] not in type_names:
```
mdhaman: use `qcode` instead.

PR #1030, opened 2018-10-17
[SDESK-3492] Display inactive contacts of an event

nrvikas:
Depends on: https://github.com/superdesk/superdesk-client-core/pull/2603

### DIFF
--- a/client/components/Contacts/ContactMetaData/index.jsx
+++ b/client/components/Contacts/ContactMetaData/index.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {StateLabel} from '../../';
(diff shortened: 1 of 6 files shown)


client/components/Contacts/ContactMetaData/index.jsx
```diff
@@ -56,15 +57,18 @@ export const ContactMetaData = ({
...
                     <Row>
-                        <span className="sd-overflow-ellipsis sd-list-item--element-grow">
-                            <span className="sd-list-item__text-strong">
-                                <ContactLabel contact={contact} />
-                            </span>
-                        </span>
+                        <ContactLabel contact={contact} />
                     </Row>
                 </Column>
+                <Column>
```
mdhaman: Unwanted like border is showing up when the contact is active.

PR #1031, opened 2018-10-17
History list is not updated when Planning is duplicated in multi-user…

GyanP:
… scenario.

https://dev.sourcefabric.org/browse/SDESK-3401


MarkLark86:
@GyanP 
The history tab now reloads as expected. My guess would be rebuilding the test instance fixed the issue.

Can you squash these two commits, then we can merge this fix.

GyanP:
@mdhaman @MarkLark86  squashed the commits.

PR #1037, opened 2018-10-22
SDESK-3506 Ingested event retains 'ingested' status after editing

ride90:


server/planning/events/events.py
```diff
@@ -382,6 +382,9 @@ def _update_single_event(self, updates, original):
             )
         else:
             # This updates Event metadata only
+            if original.get(ITEM_STATE) == WORKFLOW_STATE.INGESTED:
```
mdhaman: @ride90 we would like to have event state as `ingested` for events that are getting `patched` by `feed_parsers`.  Also check for `versioncreator` or the patch is the context of the `user`.

We have events, especially sports events that gets patched depending on the data availability from the sports provider. We would like to have the state of the event as `ingested` till the `user` modifies the item.

ride90:
Hi @nrvikas  @mdhaman ,
Now only updates by user will change state from ingest to draft.

ride90:
Hi @nrvikas  @mdhaman ,please review it.

PR #1045, opened 2018-10-25
feat(list): make list fields configurable

petrjasek:
SDNTB-527

### DIFF
--- a/client/components/fields/description.jsx
+++ b/client/components/fields/description.jsx
@@ -1,0 +1,9 @@
+import PropTypes from 'prop-types';
+
+export const description = ({item}) => item.description_text || null;
+
+description.propTypes = {
+    item: PropTypes.shape({
+        description_text: PropTypes.string,
+    }).isRequired,
+};
--- a/client/apps/Planning/PlanningList.jsx
+++ b/client/apps/Planning/PlanningList.jsx
@@ -58,6 +58,7 @@ export class PlanningListComponent extends React.Component {
             itemActions,
             hideItemActions,
             showAddCoverage,
+            listFields,
         } = this.props;
 
         return (
@@ -87,6 +88,7 @@ export class PlanningListComponent extends React.Component {
                 hideItemActions={hideItemActions}
                 showAddCoverage={showAddCoverage}
                 calendars={calendars}
+                listFields={listFields}
             />
         );
     }
@@ -120,6 +122,7 @@ PlanningListComponent.propTypes = {
     hideItemActions: PropTypes.bool,
     showAddCoverage: PropTypes.bool,
     calendars: PropTypes.array,
+    listFields: PropTypes.object,
 };
 
 const mapStateToProps = (state) => ({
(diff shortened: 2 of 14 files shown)


nrvikas (commented):
Hi @petrjasek , can we consider moving this configuration to the 'planning_types' collection where we have editor fields configuration already ? That way the planning config stays in one place.
Also it would be great to have ES6 React function syntax for new 'fields' components you've created.
Thanks.

mdhaman:
@petrjasek agree with @nrvikas. lets keep config in one place.

petrjasek:
hi, I don't think it's the same as planning types or profiles in superdesk, this should be more or less profile agnostic, defines what users need to see in lists. unless there is a usecase that it would show different fields for different types, which I'm not aware of

client/components/fields/description.jsx
```diff
@@ -0,0 +1,11 @@
+import PropTypes from 'prop-types';
+
+export function description({item}) {
```
petrjasek: @nrvikas es6 react func you mean `export const description = ({item) => ...` ?

nrvikas: Yes please.

mdhaman:
We don't have content profiles like superdesk i.e. we are using `planning_types` as `UI config` for editor of each type. On that basis, if we have `list` config document in the `planning_types` collection so that all configs are in one place. I think that is what @nrvikas is suggesting as well. As a suggestion, we can store the UI config (list, advance search) in separate collection (not in planning types). 

We have ticket to configure "Advanced Search" panel for each view as well (SDESK-3490).  The ticket description is simple but I think nobody is using place field. Moreover, we have removed "slugline" field from event editor for our instance. Hence we cannot have the "slugline" field in advance search for events view. 

`Content profiles` like Superdesk needs more thought i.e. do we need multiple events/planning types per instance? How we pre-fill data between types for actions like `Create Planning item from event` and "Add as event" from planning item.

petrjasek:
ok np, will give it a try

petrjasek:
any feedback @mdhaman @nrvikas ? using planning_types now

nrvikas:
Hi @petrjasek , it looks great.  We were thinking of refactoring this (by making 'list' as a separate document in planning_types collection) at a later stage when we include advanced search config as a separate document (as it cannot sit under just event and planning as we have a combined view). Would it be a concern for any client to reconfigure this at a later stage if 'list' is another document in planning_types with 'event', 'planning' as dict objects with its respective list configs ?

Just another thing I noted is that default fields are coming from server in case of Editor Config, but from front-end in the list config. Do you want to keep it this way ?

petrjasek:
I don't think it's an issue, but also I'm not sure if it makes sense to be a separate document. search should work for both at the same time, but list item is always of a certain type.

I don't care really about the defaults, assuming it will be configured via planning_types so this is just a fallback, not a config really.

mdhaman (approved):

nrvikas (approved):

PR #1048, opened 2018-10-30
select all items on a search results page in Planning

GyanP:
https://dev.sourcefabric.org/browse/SDNTB-528


client/actions/events/api.js
```diff
@@ -653,6 +653,10 @@ const query = (
     (dispatch, getState, {api}) => {
         let must = [];
 
+        if (selectors.multiSelect.isTotalSelected(getState())) {
+            maxResults = selectors.main.eventsTotalItems(getState()); // eslint-disable-line no-param-reassign
```
mdhaman: `API` is configured to only return max 200 items. Hence you won't get more than 200 items. Moreover, `refetch` action only fetches based on the page number set in the store.

GyanP:
Is it possible to increase the limit from 200 to maximum number?

mdhaman:
@GyanP we should not increase the limit. We don't know what is the `maximum number`. You already have total number of items in the store when the first query is done. I think you are better of paginating with `pagesize` as 200. 

Before proceeding with changes, you should discuss following with @karelpetrak :

1) If "Select All" equates to 500 or more items then how is the user going to re-order those items in the "Export As Article" modal.

2) The current logic of "Create Planning" does a `Post` to the planning endpoint to create Planning item from the event. If the user has selected 500 or more events, then the browser would issue 500 or more post to the server and the server would send 500 or more socket notifications and based on the socket notifications the browser would generate 500 or more calls to the server to refresh the planning or events view. 

It is a similar problem if the user scrolls down to 10 pages or more and does the above actions. I think that you require separate endpoints to handle these scenarios. 

Also, do we limit `no. of items` on which the user can perform these actions?

PR #1055, opened 2018-11-05
[SDESK-3409] Cofigure advance-search UI from backend settings

nrvikas:


client/actions/planning/api.js
```diff
@@ -300,6 +300,15 @@ const getCriteria = ({
                 must.push({term: {featured: true}});
             },
         },
+        {
+            condition: () => (Array.isArray(advancedSearch.place) &&
```
mdhaman: @nrvikas you might have too add place search clause in combined view search endpoint?

PR #1061, opened 2018-11-13
[SDESK-3595] Change the way saved item gets updated to front-end store

nrvikas:


client/actions/events/notifications.js
```diff
@@ -269,23 +269,19 @@ const onRecurringEventCreated = (_e, data) => (
...
-                        if (!loadedFromRefetch && (currentPreviewId === data.item || currentEditId === data.item)) {
-                            dispatch(eventsApi.fetchById(data.item, {force: true}));
-                        }
-                        dispatch(fetchItemHistoryOnRecurringNotitication(data));
-                    });
-                });
+            const storedEvent = selectors.events.storedEvents(getState())[data.item];
+            const currentEditId = selectors.forms.currentItemId(getState());
+
+            if (get(storedEvent, 'recurrence_id') || get(storedEvent, '_id') !== currentEditId
```
MarkLark86: It might be better to use `_.get(data, 'recurrence_id')` and `_.get(data, 'item')` rather than from `storedEvent`.

The Event that was updated might not be in the current session's redux store, where as the websocket notification contains the `recurrence_id` if it came from a series

client/actions/planning/notifications.js
```diff
@@ -44,30 +44,35 @@ const onPlanningCreated = (_e, data) => (
...
-
-                    if (!loadedFromRefetch && (
-                        selectors.general.currentWorkspace(getState()) === WORKSPACE.ASSIGNMENTS ||
-                        currentPreviewId === data.item || currentEditId === data.item)) {
-                        dispatch(planning.api.fetchById(data.item, {force: true}));
-                    }
+            const storedPlan = selectors.planning.storedPlannings(getState())[data.item];
+            const currentEditId = selectors.forms.currentItemId(getState());
+
+            if (get(storedPlan, '_id') !== currentEditId ||
```
MarkLark86: Same here, use `_.get(data, 'item')` in case the Planning item is not stored in the session's redux store.

server/planning/events/events.py
```diff
@@ -338,6 +338,13 @@ def on_updated(self, updates, original):
         if updates.get('recurrence_id') and not original.get('recurrence_id'):
             get_resource_service('planning').on_event_converted_to_recurring(updates, original)
 
+        updates['_etag']
```
MarkLark86: ?

PR #1064, opened 2018-11-13
[SDESK-3607] Date validations for Events and Planning

nrvikas:


server/planning/events/events.py
```diff
@@ -189,41 +189,46 @@ def on_create(self, docs):
...
-        self._validate_multiday_event_duration(event)
-        self._validate_dates(event)
+        self._validate_multiday_event_duration(updates)
+        self._validate_dates(updates, original)
 
-        if len(event.get('calendars', [])) > 0:
+        if len(updates.get('calendars', [])) > 0:
             existing_calendars = get_resource_service('vocabularies').find_one(req=None, _id='event_calendars')
-            for calendar in event['calendars']:
+            for calendar in updates['calendars']:
```
MarkLark86: Comment this out, as the client interface removes disabled Calendars from the selection (and the qcodes must exist in that selection as they came from the CV in the first place).

server/planning/events/events.py
```diff
@@ -189,41 +189,46 @@ def on_create(self, docs):
...
         if not start_date or not end_date:
-            return
+            raise SuperdeskApiError(message="Event START DATE and END DATE are mandatory.")
 
         if end_date < start_date:
             raise SuperdeskApiError(message="END TIME should be after START TIME")
 
+        if event.get('dates', {}).get('recurring_rule') and \
+                not event['dates']['recurring_rule'].get('until') and not event['dates']['recurring_rule'].get('count'):
+            raise SuperdeskApiError(message="Recurring event should have an end (until or count)")
```
MarkLark86: Also check that `until` is not before `start` or `end` dates

PR #1071, opened 2018-11-20
[SDESK-3604] Adding new field for recording 'reason' during item actions and keeping ednote & internal note untouched for 'reason' purposes.

nrvikas:


client/actions/main.js
```diff
@@ -237,6 +237,8 @@ const save = (item, withConfirmation = true, noSubsequentEditing = false) => (
                             dispatch(planningApi.receivePlannings([planningUtils.modifyForClient(savedItem)]));
                             break;
                         }
+
+                        return Promise.resolve(savedItem);
```
MarkLark86: Returning here means the end user doesn't get the following message on save:
`The Event/Planning has been saved`

client/components/InternalNoteLabel/style.scss
```diff
@@ -10,3 +10,11 @@
         margin-left: 0.6rem;
     }
 }
+
+.icon--yellow2 {
+    color: #d2a800 !important
```
MarkLark86: In the UI-Framework the `yellow2` and `highlight2` are assigned a variable for the colour not a hash value. These two classes should use the same, so that if the colour changes in the UI-Framework, Planning will automatically get the updated colour values too.

Import the UI-Framework colours file:
`@import '~superdesk-ui-framework/app/styles/variables/_colors.scss';`

And use the following variables instead of has values:
`icon--yellow2` = `$oldGold !important`
`icon--hightlight2` = `$deepPink !important`

client/components/ItemActionConfirmation/forms/cancelPlanningCoveragesForm.jsx
```diff
@@ -70,9 +70,19 @@ PlanningCovergeCancelComponent.propTypes = {
...
 const mapDispatchToProps = (dispatch) => ({
     onSubmit: (plan) => {
-        let cancelDispatch = plan._cancelAllCoverage ?
-            actions.planning.ui.cancelAllCoverage :
-            actions.planning.ui.cancelPlanning;
+        let cancelDispatch, reasonPrefix;
+
+        if (plan._cancelAllCoverage) {
+            cancelDispatch = actions.planning.ui.cancelAllCoverage;
+            reasonPrefix = 'All coverages cancelled: ';
```
MarkLark86: Will these two reason prefixes will require translations (i.e. gettext)?

client/components/Planning/PlanningEditor/index.jsx
```diff
@@ -159,16 +158,12 @@ export class PlanningEditorComponent extends React.Component {
...
-            coverageToUpdate.planning = {
-                ...coverageToUpdate.planning,
-                internal_note: `------------------------------------------------------------
-        Coverage cancelled
-        `,
-                ednote: `------------------------------------------------------------
-        Coverage cancelled
-        `,
-            };
+            coverageToUpdate.planning.workflow_status_reason = 'Coverage cancelled';
```
mdhaman: `gettext` missing here

PR #1083, opened 2018-11-29
[SDESK-3598] Planning notifications use jinja templates

marwoodandrew:


marwoodandrew:
Depends on https://github.com/superdesk/superdesk-core/pull/1430

PR #1084, opened 2018-12-03
[SDESK-3611] Event and planning unpost and spike action

nrvikas:


MarkLark86:
- Planning items that are not made public are not spiked

Actions on Assignments proceed during unpost action:
1. Create Event, Planning and Assignments (all posted)
2. Open a new window and open an action dialog on an Assignment (i.e. reassign)
3. Perform the unpost on the Event
4. Finalise the action on the Assignment.

Notice that the Assignment action proceeds, with no notification to the user.
Based on the ticket, actions on Assignments after parent item is unposted is meant to fail with a notification stating the Planning item has been unposted.

mdhaman:
If the planning item is canceled and the event is unposted then we should not change the workflow state of the planning to killed and only change the pubstatus to cancelled. 

We may need another filter for pubstatus (cancelled)?

server/planning/events/events_spike.py
```diff
@@ -62,6 +62,52 @@ def update(self, id, updates, original):
 
         return item
 
+    def on_updated(self, updates, original):
+        # Spike associated planning
+        planning_spike_service = get_resource_service('planning_spike')
+        for planning in list(get_resource_service('planning').find(where={
```
MarkLark86: 1. It would be better to use `.search(` with a source query to use elastic instead of mongo
2. The `list(` is not required as you're iterating over a cursor instance
3. No need to check if `planning['state'] == 'draft'` as the query will only return draft items

MarkLark86: 1. There is no need for `list(` here as your're already iterating over a cursor object
2. It would be better to use `search(` here with a source query to use elastic instead of mongo
3. There is no need to check `if planning['state'] == 'draft'` as you're already doing this in your query

server/planning/events/events_spike.py
```diff
@@ -62,6 +62,52 @@ def update(self, id, updates, original):
...
+                    {'state': 'spiked'}
+                )
+
+        # When a planning item associated with this event is spiked
+        # If there were any failures in removing assignments
+        # Send those notifications here
+
+        assignments_to_fetch = []
+        assignments_service = get_resource_service('assignments')
+        for planning in list(get_resource_service('planning').find(
```
MarkLark86: It might actually be better to perform a single query for `event_item` only, and use an `if planning['state'] == 'draft'/'spiked'` instead of performing two fetch/iterate cycles (i.e. merging above spike and below assignment to fetch)

client/actions/assignments/ui.js
```diff
@@ -568,7 +568,11 @@ const openSelectTemplateModal = (assignment) => (
                         dispatch(assignments.api.createFromTemplateAndShow(
                             assignment._id,
                             template.template_name
-                        ))
+                        )).catch((error) => {
+                            dispatch(self.unlockAssignment(assignment));
+                            notify.error(getErrorMessage(error, 'Failed to create an archive item.'));
```
mdhaman: gettext missing here

server/features/events_post.feature
```diff
@@ -249,3 +249,589 @@ Feature: Events Post
...
+        """
+        { "state": "killed" }
+        """
+        When we get "/assignments"
+        Then we get list with 0 items
+
+        @auth
+    @notification
+    @vocabulary
+    Scenario: Unposting an event will unpost associated planning item
```
mdhaman: looks like a duplicate scenario?

server/planning/assignments/assignments.py
```diff
@@ -696,24 +703,59 @@ def can_edit(self, item, user_id):
...
+
+        if not assignment_locked:
             raise SuperdeskApiError.forbiddenError(
-                message='Lock is not obtained on the associated Planning item'
+                message='Lock is not obtained on the Assignment item'
             )
 
+        # Make sure the content linked to assignment (if) is also not locked
+        # This is needed when the planing item is being unposted/spiked
+        archive_item = get_resource_service('archive').find_one(req=None,
```
mdhaman: we don't have mongo index on `assignment_id`. we can use elasticsearch here.

server/planning/events/events_spike.py
```diff
@@ -62,6 +62,50 @@ def update(self, id, updates, original):
...
+                    spiked_items.append(planning_spike_service.patch(planning[config.ID_FIELD],
+                                                                     {'state': 'spiked'}))
+
+            # When a planning item associated with this event is spiked
+            # If there were any failures in removing assignments
+            # Send those notifications here
+
+            assignments_to_fetch = []
+            assignments_service = get_resource_service('assignments')
+            for planning in spiked_items:
```
mdhaman: here we can query by planning id using elasticsearch as `assignments` store planning id.

server/planning/events/events_spike.py
```diff
@@ -62,6 +62,50 @@ def update(self, id, updates, original):
...
+                if planning['state'] == WORKFLOW_STATE.SPIKED:
+                    assignments_to_fetch.extend(c.get('coverage_id') for c in planning.get('coverages', []))
+
+            if len(assignments_to_fetch) > 0:
+                assignments = list(
```
mdhaman: list not required as you can use assignments.count

PR #1098, opened 2018-12-20
fix(events): make occur status nullable

petrjasek:
when there are not enough values in cv it's set to null,
but then it can't be saved.

SDESK-3759

### DIFF
--- a/server/planning/events/events_schema.py
+++ b/server/planning/events/events_schema.py
@@ -175,6 +175,7 @@ events_schema = {
         }
     },
     'occur_status': {
+        'nullable': True,
         'type': 'dict',
         'schema': {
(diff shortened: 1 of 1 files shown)


MarkLark86:
The bug is being caused when creating default values for an Event (accessing occur statuses using an index for the array rather then searching for qcode).

If there is no use case where an Event does not require an occurrence status, then improving the above logic would be a better solution:
https://github.com/superdesk/superdesk-planning/blob/master/client/utils/events.js#L767

petrjasek:
that's the usecase, they don't require occurrence status. I was thinking about fixing that default value, but this makes more sense imo also for future projects

PR #1106, opened 2019-01-07
Move menu item creation to planning index

pablopunk:
SDESK-3560

Depends on https://github.com/superdesk/superdesk-client-core/pull/2754


pablopunk:
Master build is failing, maybe because we need to update client-core in this repo?

PR #1120, opened 2019-01-29
SDESK-3799 Cleanup and document commands in manage.py

ride90:
Related PR https://github.com/superdesk/superdesk-core/pull/1505


server/planning/commands/populate_planning_types_test.py
```diff
@@ -11,13 +11,12 @@
 import os
 import json
 
-from .populate_planning_types import PopulatePlanningTypesCommand
```
MarkLark86: You can remove this test file as the command no longer exists, and because tests for AppPopulateCommand are in superdesk-core.

ride90: Hi @MarkLark86 , I left it here just to test population of `planning_types`, in the core I can't really do it, because this service is a part of planning component.

PR #1122, opened 2019-01-31
fix add to planning item validation to use planning schema

petrjasek:
SDESK-3881


MarkLark86 (approved):
Besides the one failing test, much better!

mdhaman (approved):

PR #1133, opened 2019-02-28
[SDESK-4031](fix): Using timezone offset to determine if the event is in different timezone.

mdhaman:


mdhaman:
Changes are moved to #1135

PR #1134, opened 2019-03-04
[SDESK-4045] Treat 'content item type' of text equivalent to a text coverage

nrvikas:


client/utils/assignments.js
```diff
@@ -7,8 +7,10 @@ const isNotLockRestricted = (assignment, session) => (
         lockUtils.isItemLockedInThisSession(assignment, session)
 );
 
-const isTextAssignment = (assignment) => (
-    get(assignment, 'planning.g2_content_type') === PLANNING.G2_CONTENT_TYPE.TEXT
+const isTextAssignment = (assignment, contentTypes = []) => (
+    get(assignment, 'planning.g2_content_type') === PLANNING.G2_CONTENT_TYPE.TEXT ||
```
MarkLark86: `PLANNING.G2_CONTENT_TYPE.TEXT` points to a translation for text, so this might not work as expected in other languages.

Also I believe checking `contentTypes` only would be a better way to go, as the CV entry for text could be changed by an administrator.

server/planning/assignments/assignments.py
```diff
@@ -829,6 +829,18 @@ def is_assignment_being_activated(self, updates, original):
         return original.get('assigned_to').get('state') == ASSIGNMENT_WORKFLOW_STATE.DRAFT and\
             updates.get('assigned_to', {}).get('state') == ASSIGNMENT_WORKFLOW_STATE.ASSIGNED
 
+    def is_text_assignment(self, assignment):
+        if assignment.get('planning', {}).get('g2_content_type') == 'text':
```
MarkLark86: Same goes here, use the CV to determine if the content type is text.

server/planning/assignments/assignments.py
```diff
@@ -829,6 +829,18 @@ def is_assignment_being_activated(self, updates, original):
...
+            return True
+
+        content_types = get_resource_service('vocabularies').find_one(req=None, _id='g2_content_type')
+        if content_types:
+            text_types = [t.get('qcode') for t in (content_types.get('items') or []) if
+                          t.get('qcode') == 'text' or t.get('content item type') == 'text']
```
MarkLark86: And here as well, use `content item type` instead of `qcode`.

@mdhaman This might mean we will require an upgrade script or to manually add `content item type` to our CV!

client/utils/assignments.js
```diff
@@ -1,26 +1,30 @@
...
         lockUtils.isItemLockedInThisSession(assignment, session)
 );
 
-const isTextAssignment = (assignment) => (
-    get(assignment, 'planning.g2_content_type') === PLANNING.G2_CONTENT_TYPE.TEXT
-);
+const isTextAssignment = (assignment, contentTypes = []) => {
+    const textContentTypes = contentTypes.filter((c) => get(c, 'content item type') === 'text');
+
+    return textContentTypes.length === 0 ? get(assignment, 'planning.g2_content_type') === 'text' :
```
MarkLark86: This fails if:
`contentTypes = [{qcode: 'text'}, {qcode: 'new_text', 'content item type': 'text'}`
`assignment.planning.g2_content_type = 'text'`

Because `textContentTypes.length === 1`, but the entry for `text` does not have a `'content item type'` value assigned.
It might be better to get the entry in the CV by qcode, then use `lodash.get` to get the `'content item type'` and default to `qcode` instead.

server/planning/assignments/assignments.py
```diff
@@ -829,6 +829,20 @@ def is_assignment_being_activated(self, updates, original):
...
+    def is_text_assignment(self, assignment):
+        text_assignment = False
+        text_types = []
+        content_types = get_resource_service('vocabularies').find_one(req=None, _id='g2_content_type')
+        if content_types:
+            text_types = [t.get('qcode') for t in (content_types.get('items') or [])
+                          if t.get('content item type') == 'text']
+            text_assignment = assignment.get('planning', {}).get('g2_content_type') in text_types
+
+        if len(text_types) == 0 and assignment.get('planning', {}).get('g2_content_type') == 'text':
```
MarkLark86: Same goes here, simply get the item from `content_types` via `qcode`, and get the `content item type`, defaulting to `qcode`.

PR #1142, opened 2019-03-08
add planning details authoring widget

petrjasek:
SDBELGA-70


MarkLark86:
Depends on https://github.com/superdesk/superdesk-client-core/pull/2867

petrjasek:
I will merge it now, that client PR is only required to display badge, but the widget should work as it is

PR #1147, opened 2019-03-13
[SDESK-3980] Inactive or disabled users should not be available for assignment selection

nrvikas:


client/components/Assignments/AssignmentEditor.jsx
```diff
@@ -195,7 +195,7 @@ export class AssignmentEditorComponent extends React.Component {
                     label={gettext('User')}
                     value={this.state.user}
                     onChange={this.onUserChange}
-                    users={this.state.filteredUsers}
+                    users={this.state.filteredUsers.filter((u) => u.is_active && u.is_enabled)}
```
MarkLark86: Remove the filtering of users here, simply pass through `this.state.filteredUsers`

client/components/UI/Form/SelectUserInput/index.jsx
```diff
@@ -67,9 +67,17 @@ export class SelectUserInput extends React.Component {
...
+            return list;
+        } else {
+            return list.filter((u) => u.is_active && u.is_enabled && !u.needs_activation);
+        }
+    }
+
     render() {
         const {value, users, popupContainer, label, readOnly} = this.props;
-        const userList = this.state.searchText ? this.state.filteredUserList : users;
+        const userList = this.state.searchText ? this.state.filteredUserList : this.getUsersToDisplay(users);
```
MarkLark86: Don't filter the users in any render function, as multiple renders can occur in a very short period of time. Instead place this filtering in component lifecycles such as `willReceiveProps` or `onChange`

PR #1162, opened 2019-03-25
[SDESK-4026] Extend the Coverage concept to include iterations/updates to the original text coverage item

nrvikas:


server/planning/assignments/assignments_unlink.py
```diff
@@ -28,55 +28,68 @@ def create(self, docs):
...
-                'item_id': item[config.ID_FIELD]
-            })
+                ids.append(item[config.ID_FIELD])
+                updated_items.append(item)
 
-            doc.update(item)
-            ids.append(doc[config.ID_FIELD])
-            items.append(item)
+            # Delete all delivery records for this assignment
+            get_resource_service('delivery').delete_action(lookup={'assignment_id': assignment[config.ID_FIELD]})
```
MarkLark86: When performing `Unlink as Update` action, won't this remove the delivery records for ALL items?
Should this be removing the delivery records using the `items` array instead?

nrvikas: Yup, good catch, will update.

server/planning/assignments/assignments_unlink.py
```diff
@@ -28,55 +28,68 @@ def create(self, docs):
...
-            # Update published collection
-            if item.get(ITEM_STATE) in PUBLISH_STATES:
-                get_resource_service('published').update_published_items(
+            actioned_item_id = doc.pop('item_id')
+            actioned_item = production.find_one(req=None, _id=actioned_item_id)
+
+            # To support 'unlink as update' we need to link only that update. Not the entire chain
+            if actioned_item and actioned_item.get('rewrite_of'):
+                items = production.find({
+                    'family_id': actioned_item.get('family_id'),
```
mdhaman: You will have to use the `event_id` here. 
See lines for code:
https://github.com/superdesk/superdesk-core/blob/master/apps/archive/archive_rewrite.py#L107
https://github.com/superdesk/superdesk-core/blob/master/apps/archive/archive_rewrite.py#L153
https://github.com/superdesk/superdesk-core/blob/master/apps/archive/archive_rewrite.py#L292
https://github.com/superdesk/superdesk-client-core/blob/master/scripts/apps/archive/services/FamilyService.ts#L98

Original logic was based on find_one i.e. find by `_id` but now it different field. We have to use `elastic index` over here for `event_id` for archive and published endpoints. If the item is corrected/killed you will get more docs for single item_id.  There is also `rewrite_sequence` field not sure if you need it to order deliveries.

server/features/content_rewrite.feature
```diff
@@ -89,13 +153,63 @@ Feature: Rewrite content
...
+        Then we get list with 2 items
+        """
+        {
+        "_items": [
+            { "_id": "123", "rewritten_by": "#REWRITE_ID#", "assignment_id": "#firstassignment#" },
+            {"_id": "#REWRITE_ID#", "rewrite_of": "123", "assignment_id": "#firstassignment#"}
+        ]}
+        """
+        When we enqueue published
+        When we get "/publish_queue"
```
mdhaman: As this is critical change, can we have end-to-end behave test which includes publishing of `planning item` using json planning formatter. (
see feature https://github.com/superdesk/superdesk-planning/blob/master/server/features/assignments_planning_publish.feature#L610
see step https://github.com/superdesk/superdesk-planning/blob/master/server/features/assignments_planning_publish.feature#L953
)
In the step you can assert the deliveries as well.

PR #1171, opened 2019-04-03
[SDESK-4131] Show 'Ignore/Cancel/Save' dialog for 'UpdateRepetitions' action

MarkLark86:


mdhaman:
@MarkLark86  Rebase needed

PR #1179, opened 2019-04-17
[SDESK-4094][SDESK-4125] Export templates for planning and events

nrvikas:


server/planning/planning_article_export.py
```diff
@@ -0,0 +1,131 @@
...
+        if item['event'] and item['event'].get('event_contact_info'):
+            item['contacts'] = []
+            contacts = superdesk.get_resource_service('contacts').find(where={
+                '_id': {'$in': (item['event']['event_contact_info'])}
+            })
+            for contact in contacts:
+                if contact.get('public'):
+                    item['contacts'].append(contact)
+    elif resource == 'events':
+        item['plannings'] = superdesk.get_resource_service('events').get_plannings_for_event(item)
```
MarkLark86: At the moment `contacts` are only exported when exporting a Planning item associated with an Event.

It would be better if `contacts` were also exported when exporting an Event.

server/planning/planning_export_templates.py
```diff
@@ -0,0 +1,86 @@
...
+<p>---</p>
+{% endfor %}
+'''
+    }
+}, {
+    'name': 'default_event',
+    'label': 'Default Event Template',
+    'type': 'event',
+    'template': {
+        'body_html': '''
```
petrjasek: imo would be useful to be able to use actual template file for something more complex,
maybe we could have either inline `body_html` field or `body_html_template` which would be template file name and would be rendered via `flask.render_template`

nrvikas: @petrjasek could you please take a look now. I have kept the default template a string to reduce new files being added, if you prefer that to be file too, I can change it.

petrjasek: I would rather not have json in template, but something like:
```
{
    "name": "inline_event",
    "data": {
        "slugline": "foo",
        "body_html": "{% ... %}",
    },
},
{
    "name": "template_event",
    "data": {
        "slugline": "bar",
        "body_html_template": "export-event.template.html",
    }
}
```

server/features/events_export.feature
```diff
@@ -0,0 +1,131 @@
...
+
+        Given "planning_export_templates"
+        """
+        [
+            {
+                "name": "conf_template",
+                "label": "Default Event Template",
+                "type": "event",
+                "data": {
+                    "body_html": "test_event_export_template.html"
```
petrjasek: imo it would be better explicit, so if it's using template it could be different field name like `body_html_template`

PR #1205, opened 2019-05-21
Send agendas to export templates

pablopunk:
SDBELGA-110

Send an array of agendas to be rendered on templates. Required by https://github.com/superdesk/superdesk-belga/pull/44


pablopunk:
WIP. Missing translations of unassigned agenda name

PR #1225, opened 2019-06-05
SDBELGA-108 Event templates API.

ride90:
Requests examples (no headers):

- List templates
```
curl -X GET http://localhost:5000/api/events_template
```

- Create template
```
curl -X POST http://localhost:5000/api/events_template \
  -d '{
	"template_name": "some name",
	"based_on_event": "_event_id_"
}'
```

- Get recent templates
```
curl -X GET 'http://localhost:5000/api/recent_events_template?limit=3' 
```

### DIFF
--- a/server/planning/events/events_template.py
+++ b/server/planning/events/events_template.py
@@ -216,6 +216,7 @@ class RecentEventsTemplateService(BaseService):
         if req is None:
             req = ParsedRequest()
 
+        limit = req.args.get('limit', type=int)
         pipeline = [
             {
                 '$match': {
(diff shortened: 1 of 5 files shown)


server/planning/events/events_template.py
```diff
@@ -0,0 +1,248 @@
...
+            'empty': False,
+            'unique': True
+        },
+        'based_on_event': Resource.rel(
+            'events',
+            type=metadata_schema[config.ID_FIELD]['type'],
+            embeddable=False,
+            required=True
+        ),
+        # events related fields
```
petrjasek: I think it's better to put this all into some dict so we don't mix template metadata with its data

server/planning/events/events_template.py
```diff
@@ -0,0 +1,250 @@
...
+            'empty': False,
+            'unique': True
+        },
+        'based_on_event': Resource.rel(
+            'events',
+            type=metadata_schema[config.ID_FIELD]['type'],
+            embeddable=False,
+            required=True
+        ),
+        **_event_fields
```
petrjasek: I would do it more like: `'data': {'type': 'dict', 'schema': _event_fields}`

ride90: Ahh, got it. Sure, it makes sense.

PR #1243, opened 2019-07-03
Convert agenda IDs to objects so they can be used in templates

pablopunk:
SDBELGA-129

Needed for https://github.com/superdesk/superdesk-belga/pull/51

### DIFF
--- a/server/planning/planning_article_export.py
+++ b/server/planning/planning_article_export.py
@@ -88,6 +88,36 @@ def group_items_by_agenda(items):
     return agendas
 
 
+def inject_internal_converages(items):
+    items_injected = []
+
+    coverage_labels = {}
+    cv = get_resource_service('vocabularies').find_one(req=None, _id='g2_content_type')
+    if cv:
+        coverage_labels= {_type['qcode']: _type['name'] for _type in cv['items']}
+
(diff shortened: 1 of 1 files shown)


server/planning/planning_article_export.py
```diff
@@ -72,9 +74,50 @@ def group_items_by_agenda(items):
...
+                    user = assigned_to['coverage_provider']
+                elif assigned_to.get('user'):
+                    user = get_resource_service('users').find_one(req=None, _id=assigned_to.get('user'))
+
+                if user is not None:
+                    coverage_type = coverage.get('planning').get('g2_content_type')
+                    label = coverage_labels.get(coverage_type, coverage_type)
+                    item['internal_coverages'].append({ "user": user, "type": label })
+
+        items_injected.append(item)
```
MarkLark86: Here you're creating a new array and adding every single item into this new array and returning that.

It might be better to modify the passed in `items` argument instead.

pablopunk: @MarkLark86 right, I was aware of that. I intended to make a "pure" function where the arguments are not modified, but I guess that's not python friendly :)

I'll change it 👍

PR #1244, opened 2019-07-04
[SDESK-4454] Allowing use of templates for event downloads

nrvikas:


server/planning/templates/event_download_default.html
```diff
@@ -0,0 +1,3 @@
+{% for item in items %}{% if item.get('place') %}{{ item.place | join(', ') }}{{"\r\n"}}{% endif %}{% if item.formatted_state %}({{ item.formatted_state.upper() }}) {% endif %}{{ item.get('name') }}{% if item.get('definition_short') %}: {{ item.definition_short }}{% endif %}{{ "\r\n" }}Schedule: {{ schedule }}{{ "\r\n" }}{% if item.formatted_location %}Location: {{ item.formatted_location }}{{ "\r\n" }}{% endif %}{% for contact in item.contacts %}Media contact: {{ contact }}{{ "\r\n" }}{% endfor %}
```
MarkLark86: The schedule is no longer rendering in with template.

Change `{{ schedule }}` to `{{ item.schedule }}`

client/actions/multiSelect.js
```diff
@@ -157,7 +157,13 @@ const exportAsArticle = (items = []) => (
             const itemIds = items.map((item) => item.id);
 
             if (download) {
-                window.open(`${getState().config.server.url}/planning_download/events/${itemIds.join(',')}`, '_blank');
+                let queryString = `${getState().config.server.url}/planning_download/events/${itemIds.join(',')}`;
```
MarkLark86: Use `selectors.config.getServerUrl(getState())` instead of directly accessing the state

PR #1260, opened 2019-07-17
[SDESK-4469] Introduce modal to prompt for a 'reason' to cancel individual coverages

nrvikas:


client/actions/planning/ui.js
```diff
@@ -558,6 +558,36 @@ const addNewCoverageToPlanning = (coverageType, item) => (
...
+                actionType: COVERAGES.ITEM_ACTIONS.CANCEL_COVERAGE.label,
+                coverage: coverage,
+                index: index,
+                onSubmit: onSubmit,
+                onCancel: onCancel,
+            },
+        }))
+);
+
+const cancelCoverage = (original, updatedCoverage, index) => (
```
MarkLark86: This action should be in `client/actions/planning/api.js` as it's an api call.

client/components/Coverages/CoverageEditor/index.jsx
```diff
@@ -69,11 +70,10 @@ export const CoverageEditor = ({
             callback: duplicateActions,
         }];
 
-        if (value.coverage_id) {
+        if (isExistingItem(value, 'coverage_id')) {
```
MarkLark86: This is better off placed in the `canCancelCoverage` utility function.
On the same note, a new utility function for `canAddToWorkflow` should be created and used as well.

This keeps the item checking vs available actions in the one place.

client/components/ItemActionConfirmation/forms/cancelCoverageForm.jsx
```diff
@@ -0,0 +1,131 @@
...
+import {TextAreaInput} from '../../UI/Form';
+import '../style.scss';
+
+export class CancelCoverageComponent extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {reason: '', errors: {}};
+
+        this.onReasonChange = this.onReasonChange.bind(this);
+        this.formProfile = this.props.coverageCancelProfile;
```
MarkLark86: I know the `formProfile` won't change, but it is always best practice to access it via props instead of an instance local variable.

client/components/ItemActionConfirmation/modal.jsx
```diff
@@ -150,6 +151,10 @@ export class ItemActionConfirmationModal extends React.Component {
                 },
                 form: CancelPlanningCoveragesForm,
             },
+            [COVERAGES.ITEM_ACTIONS.CANCEL_COVERAGE.label]: {
+                title: get(COVERAGES, 'ITEM_ACTIONS.CANCEL_COVERAGE.label'),
+                form: CancelCoverageForm,
```
MarkLark86: This should probably have a `saveText: gettext('Cancel Coverage')`. Otherwise the button will show `SAVE`.

client/actions/planning/ui.js
nrvikas: Subsequent to the api call, there is UI notification and further processing of returned data. Similar examples in the same file: addCoverageToWorkflow, removeAssignment, spike, etc.

MarkLark86: Fair enough

PR #1265, opened 2019-07-19
[SDESK-4224][SDESK-4510] (fix): Not returning to the previous editor after action modals

MarkLark86:


MarkLark86:
Creating another PR that will fix this

PR #1267, opened 2019-07-24
[SDESK-4513] (fix): Coverages revert to draft when cancelling from popup editor

MarkLark86:


client/components/Planning/PlanningEditor/index.jsx
```diff
@@ -85,6 +85,33 @@ export class PlanningEditorComponent extends React.Component {
         if (this.props.event) {
             this.props.fetchEventFiles(this.props.event);
         }
+
+        // If this is the popup-editor, check if any coverages have changed to cancelled
+        if (this.props.inModalView) {
```
nrvikas: We have very similar functionality here: https://github.com/superdesk/superdesk-planning/blob/f6fe9a6fb901bb3af563ee470e136b03d3d84dbb/client/components/Planning/PlanningEditor/index.jsx#L294

Since most of the functionality is the same, we can extract common parts into one function so the code won't be repeated ?

MarkLark86:
Creating another PR that will fix this

PR #1272, opened 2019-07-31
[SDESK-3286] Close popup modals with ESC key

marwoodandrew:


client/components/AddToPlanningModal/index.jsx
```diff
@@ -8,12 +8,31 @@ import {AddToPlanningApp} from '../../apps';
...
 import {gettext} from '../../utils';
+import {KEYCODES} from '../../constants';
 
 export class AddToPlanningComponent extends React.Component {
     constructor(props) {
         super(props);
 
         this.dom = {popupContainer: null};
+        this.handleKeydown = this.handleKeydown.bind(this);
+        this.handleHide = this.props.handleHide;
```
MarkLark86: Whenever you're accessing anything from `props`, it is always better to use `this.props` everywhere in this file.
This makes it clear that you're accessing something passed down from a higher up component

client/components/Assignments/SelectDeskTemplate.jsx
```diff
@@ -91,3 +119,5 @@ SelectDeskTemplate.propTypes = {
         privateTemplates: PropTypes.array,
     }),
 };
+
+export const SelectDeskTemplate = connect(null, null)(SelectDeskTemplateComponent);
```
MarkLark86: Using `connect` here should not be required (as `null, null` is being passed through).
You should be able to use `export class SelectDeskTemplate` instead

client/components/FulFilAssignmentModal/index.jsx
```diff
@@ -9,79 +9,105 @@ import {Modal} from '../index';
...
+        document.removeEventListener('keydown', this.handleKeydown);
+    }
 
-        if (onCancel) {
-            onCancel();
+    handleKeydown(event) {
+        if (event.keyCode === KEYCODES.ESCAPE) {
+            event.preventDefault();
+            this.handleCancel();
+            // this.props.modalProps.$scope.resolve();
```
MarkLark86: Remove comment

client/components/NotificationModal.jsx
```diff
@@ -42,3 +69,5 @@ NotificationModal.propTypes = {
         action: PropTypes.func,
     }),
 };
+
+export const NotificationModal = connect(null, null)(NotificationComponent);
```
MarkLark86: Using `connect` here should not be required (as `null, null` is being passed through).
You should be able to use `export class NotificationModal` instead

marwoodandrew: Done

PR #1273, opened 2019-07-31
[SDESK-4427] New Event action 'Mark as Completed'

nrvikas:


MarkLark86:
This mostly works, there is just one bug I found. When marking a series as completed, there ends up being two events on today (see the test instance for an example).

This happens no matter which event I use to perform this action on.

PR #1274, opened 2019-08-01
[SDESK-4224][SDESK-4510][SDESK-4513] (fix): Don't unmount the PopupEditor when action modal is shown

MarkLark86:
Uses React.Portal for the PopupEditor and 'display: none' on the editor when an action modal is shown


nrvikas:
Noticed a couple of issues:
1) User A is forcefully unlocked by User B. When User B changes editor to 'Popup' mode, browser crashes.
2) Item actions when executed on Modal does not reflect changes (eg. Update time still shows previous time) and renders Popup as dirty.

PR #1293, opened 2019-08-19
[SDESK-4561] Coverage forward updates

nrvikas:


MarkLark86:
The following needs to be addressed:
* An update coverage should have the a genre field in the schema, and set to the 'Update' genre by default (maybe the form should have genre field disabled with this value)
* When adding an update coverage, the mini-form should be opened automatically
* List item tooltips does not contain 'Update due:', and the icon is the same as a normal text coverage
* Scheduled updates button is available in non-text coverages (i.e. picture)

PR #1294, opened 2019-08-19
feat(draggable): Added ability to make modals draggable

pavlovicnemanja:
[SDNTB-584] - As a user I want to be able to move the modal windows


MarkLark86:
@pavlovicnemanja Are all the modals to be movable or just the PopupEditor? The ticket use the PopupEditor as an example. Would it be better to have all the action modals movable as well?

pavlovicnemanja:
@MarkLark86 hey Mark, currently only EditInPopup modal is draggable, but now it's configurable, you just need to set `draggable` prop on Modal Component, and also `backdrop={false}` to disable backdrop and make it clickable beyond it. Request was only for this modal, you can open new ticket and I can add options in whatever modal you request. Question, are tests failing because of my changes, or test instance is broken? I can't really get it from travis log.

MarkLark86:
> @MarkLark86 hey Mark, currently only EditInPopup modal is draggable, but now it's configurable, you just need to set `draggable` prop on Modal Component, and also `backdrop={false}` to disable backdrop and make it clickable beyond it. Request was only for this modal, you can open new ticket and I can add options in whatever modal you request. Question, are tests failing because of my changes, or test instance is broken? I can't really get it from travis log.

I wasn't sure which modals were required to be draggable, as the ticket does not mention this requirement (only an example). All good.

As for Travis, Planning instances are currently broken again because of a change to superdesk-core (updated celery/kombu). I will be creating a PR soon to fix this

pavlovicnemanja:
@MarkLark86 I discussed this with Fritz and Karel, and feedback was to add only on this modal. As I sad, It's not big deal, small change to add it on others. Just need specification on which one you want :)

pavlovicnemanja:
Sorry, closed it by accident. Stupid button on wrong place.

PR #1308, opened 2019-09-02
[SDESK-4662] Close Assignment preview on desk reassignment

MarkLark86:


MarkLark86:
Won't fix, unless brought to our attention as an end-user issue

PR #1311, opened 2019-09-03
[SDESK-4595] Move the attachment icon in lists

marwoodandrew:


nrvikas:
> The paperclip icon behavior in combined view with and without a planning item is very different.

This appears fine sorry, my bad.

PR #1316, opened 2019-09-06
[SDESK-4647] Item actions for Scheduled Updates

nrvikas:


client/components/Coverages/index.jsx
```diff
@@ -1,7 +1,7 @@
 export {CoverageArrayInput} from './CoverageArrayInput';
 export {CoverageEditor} from './CoverageEditor';
 export {CoverageItem} from './CoverageItem';
-export {CoveragePreview} from './CoveragePreview';
+export {CoveragePreview} from './CoveragePreview/index';
```
MarkLark86: `readOnly` prop is not being passed into `<ScheduledUpdate` component, resulting in React crashing

PR #1322, opened 2019-09-10
add compatibility fix so event templates commit can be restored

tomaskikutis:
Fixes the issue described in https://github.com/superdesk/superdesk-planning/pull/1319. @MarkLark86 can you restore event templates after this one is merged?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
+    "postinstall": "node tasks/backwards-compatibility-fixes.js",
     "fix": "eslint client --ext=jsx --ext=js --ext=tsx --ext=ts --ext=spec --fix",
     "hint": "eslint client --ext=jsx --ext=js --ext=tsx --ext=ts --ext=spec",
     "lint": "eslint client --ext=jsx --ext=js --ext=tsx --ext=ts --ext=spec --quiet",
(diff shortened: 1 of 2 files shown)


MarkLark86:
@tomaskikutis This would be better placed in superdesk-client-core as a backwards compatibility fix (programmatically creating files in another module is not a good practice and should be avoided).

Also using `__dirname` and relative paths do not work if you have used `npm link` to link to a local development package. This is also the case with extensions and finding a custom `superdesk.config.js` file (i.e. if we have linked superdesk-client-core to our superdesk-aap's node_modules)

MarkLark86:
Why are you using git submodule to add superdesk? This addition makes compiling fail on superdesk-aap with `git submodule fatal no submodule mapping found in .gitmodules for path for 'superdesk'`?

What is the use case for a submodule here that points to a particular commit from client-core?

tomaskikutis:
I've added the fix to 1.30 https://github.com/superdesk/superdesk-client-core/commit/50f078108d8d1757fa8b22e52da746fa4cb2bc3d

I'm aware of the issue with npm linked extensions. Most of our extensions are in the core, so it's not a big problem for use yet, but I'll try to address it soon.

Regarding your second comment about git submodules it looks like you meant to comment somewhere else since there's nothing related to submodules in this PR.

PR #1324, opened 2019-09-11
[SDESK-4562] Logically constrain the scheduling of planned updates to a coverage

nrvikas:
… a coverage


MarkLark86:
Validation fails when attempting to create a new planning item, with a coverage and a scheduled update (1 hour after the parent coverage). Only works if I set the scheduled update to be hours after the parent coverage

MarkLark86:
React crashes in the test instance when you go to 'Featured Planning' on navigate to 17th or 18th (for a Planning item with a coverage and a scheduled update)

PR #1344, opened 2019-09-26
[SDESK-4565] Independently sort Assignment lists

MarkLark86:


nrvikas:
1) Lists are not updating their sort order on push notifications
2) Items are not moving between stages when 'Confirm Availability' or 'Revert Availability' is executed.
3) Sort button is highlighted even after clicking and out of focus.
4) No sorting option found on single-group view

PR #1363, opened 2019-10-16
SDNTB-604 Move NTB related feed parser and formatter from planning to NTB repo

ride90:
This PR goes side by side with https://github.com/superdesk/superdesk-ntb/pull/344.
PR for NTB is broken because it has planning 1.7 as a dependency, instead, I've triggered a test instance https://github.com/superdesk/superdesk-ntb/pull/345 (https://ntbpr-345.test.superdesk.org).

NOTE: I was able to ingest and output events on a test instance https://ntbpr-345.test.superdesk.org


server/planning/feeding_services/event_email_service.py
```diff
@@ -117,7 +121,7 @@ def _update(self, provider, update):
                                                 content = io.BytesIO(attachment)
                                                 res = process_file_from_stream(content, part.get_content_type())
                                                 file_name, content_type, metadata = res
-                                                if isinstance(parser, NTBEventXMLFeedParser):
+                                                if NTBEventXMLFeedParser and isinstance(parser, NTBEventXMLFeedParser):
```
MarkLark86: Instead of attempting to import the `NTBEventXMLFeedParser` and testing `isinstance`, I think it might be better to use something similar to `can_parse` of the registered feed parser itself. This way these 3 usages for both `NTBEventXMLFeedParser` and `IcsTwoFeedParser` can be removed and move the logic into the parser themselves.

ride90: The logic we have right now in `event_file_service.py`, `event_http_service.py` and `event_email_service.py` checks a class anyway. I don't see a way how a method in a parser can help here. We can avoid attempting to import `NTBEventXMLFeedParser` and import 'IcsTwoFeedParser' by checking a class name `if parser.__class__.__name__ == 'NTBEventXMLFeedParser'` which is IMO a less reliable.
Can you provide me an example of how you see a method similar to `can_parse ` ?

ride90: Hi @MarkLark86 , any update on it :arrow_up: ?

MarkLark86: You can use python's `getattr` to determine if the parser has a specific method defined, otherwise use the default. The purpose is to remove all specific NTB/ICS code from these services.

You could use something like the following:
```python
if getattr(parser, 'parse_email'):
    parser.parse_email(....)
else:
    logger.warn('Ingesting event with unknown parser'
    new_items.append(parser.parse(data, provider)
```

And then implement something similar to `parse_file` and `parse_http`.

This will make it easier to maintain the code by separating out the implementation of the parser from the service entirely.

ride90: @MarkLark86 check it, please. 
Сorresponding сommit for NTB https://github.com/superdesk/superdesk-ntb/pull/344/commits/9d213172b0b21d62a11e9a122917f2806665b3ce

MarkLark86: Good to go

PR #1373, opened 2019-10-30
Fix buggy behaviour when trying to remove subjects from event and planning item

tomaskikutis:
SDNTB-613

Values from the same array are used both for subject and category fields thus removing items by index can't be used.

category-subject


MarkLark86:
@tomaskikutis We will be doing the final preparations for the release of v1.9.0 tomorrow. If you would like this in the release, please fix the failing tests so we can merge this.

PR #1392, opened 2019-11-25
[SDESK-4772] Avoid planning lists' scroll position from jumping to start on item notifications

nrvikas:


client/actions/agenda.js
```diff
@@ -92,8 +92,14 @@ const selectAgenda = (agendaId, params = {}) => (
 
         // update the url (deep linking)
         $timeout(() => ($location.search('agenda', agendaId)));
+
         // reload the plannings list
-        return dispatch(fetchSelectedAgendaPlannings(params));
+        dispatch(main.setUnsetUserInitiatedSearch(true));
+        return dispatch(fetchSelectedAgendaPlannings(params))
+            .then((data) => {
+                dispatch(main.setUnsetUserInitiatedSearch(false));
```
MarkLark86: Dispatch this action in the `finally` phase of a `Promise`, to ensure the value is reset in all conditions (similar to how `setUnsetLoadingIndicator` uses `finally`).

PR #1412, opened 2020-01-17
SDBELGA-262 [EVENT FILES] Save additional file information.

ride90:
In mongo:
Screenshot 2020-01-17 at 16 28 41

Production API response:
Screenshot 2020-01-17 at 16 38 51

### DIFF
--- a/server/planning/events/events_files.py
+++ b/server/planning/events/events_files.py
@@ -8,10 +8,11 @@
 
 """Superdesk Files"""
 
+import logging
+from flask import current_app as app
 import superdesk
 from superdesk import get_resource_service
 from superdesk.errors import SuperdeskApiError
-import logging
 
 logger = logging.getLogger(__name__)
 
@@ -45,7 +46,15 @@ class EventsFilesService(superdesk.Service):
     def on_create(self, docs):
         for doc in docs:
             # save the media id to retrieve the file later
-            doc['filemeta'] = {'media_id': doc['media']}
+            if 'media' in doc:
+                _file = app.media.get(doc['media'])
+                if _file:
+                    doc['filemeta'] = {
+                        'media_id': doc['media'],
+                        'content_type': _file.content_type,
+                        'filename': _file.filename,
+                        'length': _file.length
+                    }
 
     def on_created(self, docs):
         for doc in docs:
--- a/server/features/publish.feature
+++ b/server/features/publish.feature
@@ -107,7 +107,7 @@ Feature: Publish
                     "type": "event",
                     "published_item": {
                         "_id": "#events._id#",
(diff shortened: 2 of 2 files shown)


server/planning/events/events_files.py
```diff
@@ -45,7 +46,16 @@ class EventsFilesService(superdesk.Service):
...
             # save the media id to retrieve the file later
-            doc['filemeta'] = {'media_id': doc['media']}
+            _file = app.media.get(doc['media'])
+            doc['filemeta'] = {
+                'media_id': doc['media'],
+                'chunk_size': _file.chunk_size,
+                'content_type': _file.content_type,
+                'filename': _file.filename,
+                'length': _file.length,
+                'name': _file.name,
```
petrjasek: isn't this same as filename?

server/planning/events/events_files.py
```diff
@@ -45,7 +46,16 @@ class EventsFilesService(superdesk.Service):
     def on_create(self, docs):
         for doc in docs:
             # save the media id to retrieve the file later
-            doc['filemeta'] = {'media_id': doc['media']}
+            _file = app.media.get(doc['media'])
+            doc['filemeta'] = {
+                'media_id': doc['media'],
+                'chunk_size': _file.chunk_size,
```
petrjasek: I think this is mongo specific, wouldn't expose it

server/planning/events/events_files.py
```diff
@@ -45,7 +46,16 @@ class EventsFilesService(superdesk.Service):
...
+                'content_type': _file.content_type,
+                'filename': _file.filename,
+                'length': _file.length,
+                'name': _file.name,
+                'upload_date': _file.upload_date
```
petrjasek: not sure we need this, probably same as `_created` on the item

server/planning/events/events_files.py
ride90: yes, it is actually an alias for filename. Will drop it.

server/planning/events/events_files.py
ride90: yep, `AmazonObjectWrapper` does not have it. will drop it.

server/planning/events/events_files.py
```diff
@@ -45,7 +46,13 @@ class EventsFilesService(superdesk.Service):
     def on_create(self, docs):
         for doc in docs:
             # save the media id to retrieve the file later
-            doc['filemeta'] = {'media_id': doc['media']}
+            _file = app.media.get(doc['media'])
+            doc['filemeta'] = {
+                'media_id': doc['media'],
```
petrjasek: personally would drop this one too, it's there already as media

ride90: I thought about it, but not sure, it might break something. Let's try to drop it.

PR #1459, opened 2020-04-21
Config for assignments list fields

pablopunk:
SDESK-5155


client/components/Assignments/AssignmentItem/index.jsx
```diff
@@ -309,4 +255,17 @@ AssignmentItem.propTypes = {
     contentTypes: PropTypes.array,
     assignedDesk: PropTypes.object,
     contacts: PropTypes.object,
+    config: PropTypes.object,
 };
+
+const mapStateToProps = (state) => ({
+    config: state.config,
```
MarkLark86, line 262: You should be able to start using `import {appConfig} from 'appConfig';` without my PR being merged, as my PR removes the config from the redux store

client/components/Assignments/AssignmentItem/fields/Name.tsx
```diff
@@ -0,0 +1,29 @@
...
+    fetchPlanningItemForAssignment: any;
+}
+
+export const NameComponent = ({
+    assignment,
+    fetchPlanningItemForAssignment,
+}: IProps) => {
+    const [name, setName] = React.useState(null);
+
+    if (name === null) {
```
MarkLark86, line 15: There are data syncing and performance concerns with this approach:

1. If the name of the Planning item changes, this component will not update
2. An API request is sent for every Assignment in the list view, for every session currently in the Assignment workspace (albeit on load only)

To get around this issue we decided to sync the Planning details required to the Assignment.

1. Add `name` to the Assignment schema (https://github.com/superdesk/superdesk-planning/blob/master/server/planning/assignments/assignments.py#L1241)
2. Sync `name` from the Planning item to the Assignment (https://github.com/superdesk/superdesk-planning/blob/master/server/planning/planning/planning.py#L716)
3. Finally update the mongo/elastic records for the Assignment (https://github.com/superdesk/superdesk-planning/blob/master/server/planning/planning/planning.py#L734)

Then you can access the `name` in the same way that `description_text` is being used.

client/components/Assignments/AssignmentItem/fields/Content.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+interface IProps {
+    assignment: any;
+}
+
+export const ContentComponent = ({assignment}: IProps) => {
+    const hasContent = assignmentUtils.assignmentHasContent(assignment);
+
+    if (!hasContent) {
+        return <span />;
```
MarkLark86, line 13: Is returning `<span />` here achieving anything, or can `null` be returned instead?

client/components/Assignments/AssignmentItem/fields/DueDate.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    toBeConfirmed={get(
+                        assignment,
+                        `planning.${TO_BE_CONFIRMED_FIELD}`
+                    )}
+                />
+            ) : (
+                <span>{gettext('\'not scheduled yet\'')}</span>
+            )}
+            {isOverdue && (
+                <span className="label label--warning label--hollow">due</span>
```
MarkLark86, line 42: Missing `gettext`

client/components/Assignments/AssignmentItem/fields/Headline.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const itemId = assignment.item_ids[0];
+
+    if (!itemId) {
+        return null;
+    }
+
+    const [item, setItem] = React.useState(null);
+
+    if (!item) {
+        loadArchiveItem(assignment).then(setItem);
```
MarkLark86, line 25: Any ideas on how to handle changes to archive item's headline?

client/components/Assignments/AssignmentItem/fields/index.ts
```diff
@@ -0,0 +1,55 @@
...
+        return PriorityComponent;
+    case 'slugline':
+        return SluglineComponent;
+    case 'state':
+        return StateComponent;
+    }
+};
+
+export const DEFAULT_ASSSIGNMENTS_LIST_VIEW = {
+    firstLine: ['slugline', 'description_text', 'name'],
```
MarkLark86, line 44: In the current codebase, only `slugline` and `description_text` are displayed, the default should reflect this.

client/components/Assignments/AssignmentItem/fields/DescriptionText.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+interface IProps {
+    assignment: any;
+}
+
+export const DescriptionTextComponent = ({assignment}: IProps) => {
+    const descriptionText = get(assignment, 'description_text');
+
+    return (
+        <span className="sd-overflow-ellipsis sd-list-item--element-grow">
```
MarkLark86, line 12: There is a UI glitch here if the description is too long then the slugline spans multiple lines inside the first row.

In the current codebase, there is a span element encasing the `slugline` and `decription_text` which has these two classes. This enables the ellipsis across the row.

Play around and make sure long descriptions/names/sluglines places an ellipsis and does not increase the height of the first line (maybe using the two classes `sd-overflow-ellipsis sd-list-item--element-grow` on the first row in `renderContentColumn`)

client/components/Assignments/AssignmentItem/fields/Headline.tsx
```diff
@@ -0,0 +1,30 @@
...
+        return null;
+    }
+
+    const itemId = assignment.item_ids[0];
+
+    if (!itemId) {
+        return null;
+    }
+
+    const [item, setItem] = React.useState(null);
```
tomaskikutis, line 22: can you stick to class API?

client/components/Assignments/AssignmentItem/fields/Content.tsx
pablopunk: Yeah I meant to put null

client/components/Assignments/AssignmentItem/fields/DescriptionText.tsx
pablopunk: Good to know, I'll test it with long descriptions 👍

client/components/Assignments/AssignmentItem/fields/Headline.tsx
pablopunk: I guess we should tackle this like the `name` field approach, data changes should trigger a re-render of the parent component, and the child shouldn't be aware of data changes. Do we use events like `content:update` or so in planning?

client/components/Assignments/AssignmentItem/fields/Name.tsx
pablopunk: I guess I looked for where to change this on the client instead of the server... Thanks for sharing 👍

client/components/Assignments/AssignmentItem/fields/index.ts
pablopunk: Yeah I left it there by mistake, I'll change it back

client/components/Assignments/AssignmentItem/fields/Headline.tsx
pablopunk: I'll probably just change this to accept the item as a prop instead of fetching it from here. But why do you prefer the class API? We already use function components, and hooks are the way to go to make them stateful if there's no reason to create a class

tomaskikutis: See this discussion https://github.com/superdesk/superdesk-ui-framework/pull/436#discussion_r383839845

Hooks are a totally different way of dealing with state, which has a significant learning curve. It increases the number of things everyone needs to understand in order to maintain the codebase but doesn't bring many advantages. There may be one here and there, but for the most part, one can achieve the same things using classes and higher-order components.

I'm not saying that we should never use hooks, but until we bump into cases where hooks would have a significant impact, I'd say let's keep it simple and use the same thing in all places.

server/planning/assignments/assignments.py
```diff
@@ -1239,6 +1239,7 @@ def accept_assignment(self, assignment_id, assignee):
     # coverage details
     'planning': deepcopy(coverage_schema['planning']),
     'description_text': metadata_schema['description_text'],
+    'name': metadata_schema['name'],
```
MarkLark86, line 1242: `name` isn't defined in the `metadata_schema` but actually in the `planning_schema`.

``` python
# 1. Import the planning_schema
from planning.planning.planning import planning_schema

# 2. Change metadata_schema to planning_schema
'name': planning_schema['name']
```

client/components/Assignments/AssignmentItem/fields/Headline.tsx
MarkLark86: Below is a suggested solution to this problem, including reducing performance impact of loading archive items.

Keep in mind that loading/syncing archive items should only happen if any field that requires archive items, such as `headline`, is configured in `appConfig.assignmentsList`.

**1. Bulk loading required archive items**
Instead of sending an API query for every Assignment that has content linked, it might be best to send a single query to fetch all associated content when receiving assignments.

Create a new dispatch action for assignments that bulk loads archive items (similar to the existing `loadArchiveItem`, https://github.com/superdesk/superdesk-planning/blob/master/client/actions/assignments/api.js#L485), and dispatch that action when receiving assignments at https://github.com/superdesk/superdesk-planning/blob/master/client/actions/assignments/api.js#L263.

**2. Keeping archive items up to date**
To listen for angular broadcast events in Planning you can add a new entry into the respective actions/notifications file.

i.e.
Create an `onContentUpdated` function that re-fetches the archive item and updates the store if the item exists in `selectors.assignments.getStoredArchiveItems(getState()`)

Add this function to the assignment notifications map in https://github.com/superdesk/superdesk-planning/blob/master/client/actions/assignments/notifications.js#L376:

``` js
'content:update': () => (self.onContentUpdated),
```
This should keep any currently loaded archive item up to date.

**3. Sending the archive item to the AssignmentItem component**
Add `selectors.assignments.getStoredArchiveItems(state)` to the `AssignmentGroupList` component props, which can then be passed down to the `AssignmentItem` and automatically updated if/when the store updates.

client/components/Assignments/AssignmentItem/fields/Headline.tsx
pablopunk: FC are also a different way (compared to Classes) of creating components, but that's not a reason to use one only. To maintain the codebase you need to know React, and hooks are a more automatic and simpler (IMO) way of handling state, you do the same with less code.

Also, even tho they do the same as setState, they're not really the same because they allow extra functionality that can be very useful in the future. They allow us to create common hooks to modify state and even use other modules ([this one](https://github.com/zeit/swr) as an example) that simplify our code doing a lot of things automatically, without having to deal with classes hierarchy.

Anyway, I'm not gonna discuss it anymore cause I won't push any hooks to this PR, as I'm probably gonna replace this component with one that gets the data as a prop 👍

tomaskikutis: > FC are also a different way (compared to Classes) of creating components, but that's not a reason to use one only.

Function components have almost no complexity or learning curve. You copy-paste the code from the render block of a class component into a function, update props usage and you have a function component. That being said, function components have limitations that class components don't have - usage or refs or lifecycle methods.

By sticking to one way of doing things we make the codebase easier for others to contribute/maintain. If we use only class components, no one has to read about the differences to function components and when to use which. That I think is a pretty good reason to use only one.

Now functional components aren't a big deal and there isn't _that_ much to learn, so I'm not complaining about both being used. It's different with hooks since there's much more everyone would need to learn.

> hooks are a more automatic and simpler (IMO) way of handling state, you do the same with less code.

What exactly do you find simpler? Regarding less code, I don't think that 5 lines of constructor code are a problem. We should focus more to keeping low complexity than saving a few lines of boilerplate.

I'm not saying hooks are bad or can't be helpful in the future. They can, but I suggest waiting for that future instead of spending resources upfront before we need it.

Do you have any specific ideas on improving something in superdesk that can't be done/is hard to do using higher-order components?

Don't take this personally, I'm just trying to look at the cost/benefit ratio, look at the objective arguments for/against and base the decisions on that.

pablopunk: HOC can also be made with functional components, I'm ok with that.

client/actions/assignments/api.ts
```diff
@@ -469,12 +473,55 @@ const receiveAssignmentHistory = (items) => ({
...
+    dispatch,
+    getState,
+    {api, notify, search}
+) => {
+    const assignmentIds = assignments.map((a) => a._id);
+
+    const query = search.query();
+    const filter = {
+        bool: {
+            should: assignmentIds.map((id) => ({term: {assignment_id: id}})),
```
MarkLark86, line 494: Re-fetching all assignments on `content:update` might be a bit much. It might be better to only reload the content items that are changed based on the ids provided in the `content:update` notification.

`should` can also be changed to a `must.terms` query, i.e.
`must: {terms: {_id: content_ids}}`

client/actions/assignments/notifications.js
```diff
@@ -351,6 +352,29 @@ const onAssignmentDeleted = (_e, data) => (
     }
 );
 
+export const onContentUpdate = (_e, data) => (
+    (dispatch, getState) => {
+        const updatedIds = Object.keys(data.items);
+        const currentItems = Object.values(getState().assignment.archive);
```
MarkLark86, line 358: Whenever accessing anything from the redux store, its ideal to use selectors. This way developers don't have to worry about where the data is in the store (or make changes if the store structure changes), all developers have to worry about is what selectors are available to them.

client/components/Assignments/AssignmentItem/fields/Headline.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+interface IProps {
+    assignment: any;
+    archiveItemForAssignment: {[assignmentId: string]: any}
+}
+
+export const HeadlineComponent = ({assignment, archiveItemForAssignment}: IProps) => {
+    const item = archiveItemForAssignment[assignment._id];
```
MarkLark86, line 10: It could be possible that `archiveItemForAssignment` is undefined, in which case this statement will cause an exception

client/actions/assignments/api.ts
```diff
@@ -491,7 +491,11 @@ const loadArchiveItems = (assignments: Array<any>) => (
     const query = search.query();
     const filter = {
         bool: {
-            should: assignmentIds.map((id) => ({term: {assignment_id: id}})),
+            must: {
+                terms: {
+                    assignment_id: [assignmentIds]
```
MarkLark86, line 496: `assignmentIds` is already an array

MarkLark86:
Travis-CI was passing but not reflecting in Github checks. So I force merged this PR

PR #1465, opened 2020-05-01
[SDESK-5223] Replace React hooks with Class for ExpandableText component

MarkLark86:
PlanningDetails widget threw the React error code https://reactjs.org/docs/error-decoder.html/?invariant=321
This may be happening because Superdesk has React 16.9 while UI-Framework has 16.8
as the UI-Framework is bundling React inside its dist/superdesk-ui.bundle.js

Also added padding for this widget (SDESK-4327)

### DIFF
--- a/client/components/UI/Preview/ExpandableText.jsx
+++ b/client/components/UI/Preview/ExpandableText.jsx
@@ -1,4 +1,4 @@
-import React, {useRef, useState} from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import {get} from 'lodash';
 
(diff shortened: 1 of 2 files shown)


client/components/UI/Preview/ExpandableText.jsx
```diff
@@ -1,52 +1,66 @@
...
+        this.dom = {parent: null};
+        this.toggleExpanded = this.toggleExpanded.bind(this);
+        this.setDomRef = this.setDomRef.bind(this);
+    }
 
-    const toggleExpanded = () => {
-        setExpanded(!expanded);
-        if (valueRef.current && valueRef.current.parentNode) {
-            valueRef.current.parentNode.scrollIntoView();
+    toggleExpanded(expanded) {
```
pablopunk, line 18: you're not using that argument. Also, that would be an Event right?

MarkLark86: Good point I missed that, blame it on copy and paste!!!

PR #1467, opened 2020-05-08
[SDESK-5245] fix: Planning item not unlocked after cancelling

MarkLark86:


MarkLark86:
Currently failing due to Eve/Flask updates (requires https://dev.sourcefabric.org/browse/SDESK-5204)

PR #1471, opened 2020-05-20
FIX: Cannot create planning item from events after disabling agenda. [SDESK-4479]

GyanP:


MarkLark86:
@GyanP I think it would be best not to allow a Planning item to be created with a disabled Agenda.
When creating a Planning item from an Event, the currently selected Agenda is used. If this Agenda is disabled, I think it would be better to remove this automatically assigned Agenda when saving.

https://github.com/superdesk/superdesk-planning/blob/develop/client/actions/agenda.js#L190

PR #1473, opened 2020-06-15
[SDNTB-633] Ability to define export templates in settings.py

MarkLark86:


MarkLark86:
@ride90 The change itself is https://github.com/superdesk/superdesk-planning/pull/1473/files#diff-0f3bdd0dbeefd20c72e32d3e5adbb874, the rest is changes to tests etc

PR #1478, opened 2020-07-09
[SDESK-5389] fix: Locking Assignment assigns incorrect _links attribute

MarkLark86:


client/actions/assignments/api.ts
```diff
@@ -393,10 +394,11 @@ const revert = (item) => (
 
 /**
  * Action to lock an assignment
- * @param {String} item - Assignment to be unlocked
+ * @param {IAssignmentItem} assignment - Assignment to be unlocked
+ * @param {String} action - The action to assign to the lock
```
GyanP, line 398: I think we don't need them anymore. Everything is clear with types itself.

MarkLark86: Agree'd, just habit I suppose.

petrjasek:
isn't it rather a backend issue?

MarkLark86:
> isn't it rather a backend issue?

@petrjasek I have updated the PR to fix the issue on the server now

PR #1480, opened 2020-07-13
[SDESK-5005] Add optional internal reference field to events

marwoodandrew:


client/components/Events/EventEditor/index.jsx
```diff
@@ -283,6 +283,15 @@ export class EventEditorComponent extends React.Component {
                         onFocus={onFocusEvent}
                     />
 
+                    <Field
+                        component={TextInput}
+                        field="reference"
+                        label={gettext('External Reference')}
+                        enabled={get(formProfile, 'editor.reference.enabled', false)}
```
MarkLark86, line 290: Can you add the `reference` field to the following `formProfile` definitions:

- `EventSchema` -  https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning_types.py#L53
- `DEFAULT_EDITOR.event` - https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning_types.py#L149
- `DEFAULT_EDITOR.advanced_search` - https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning_types.py#L206

marwoodandrew: I've added the field to planning_types.py and defaulted it to not appear, I assume other installations don't required it.

PR #1482, opened 2020-07-28
[SDESK-5414] fix: Planning search fails on elastic 7

MarkLark86:
Depends on https://github.com/petrjasek/eve-elastic/pull/91


MarkLark86:
@petrjasek To make it easier to review, all search parameters have tests covered, and all elastic queries are generated in the utils file (https://github.com/superdesk/superdesk-planning/pull/1482/commits/3287b77df21042dfab4fd3219cf00acbad0842c9#diff-b44267528c24eeec2046502452bde5d4).

So if you review the elastic queries in the 2 test commits, that will tell you what is generated for all search parameters.

PR #1486, opened 2020-08-26
FIX: Cannot add planning item to featured stories. [SDESK-5456]

GyanP:


MarkLark86:
I think it might be better to add `build_custom_hateoas` in the `on_created` method instead (I believe a change in Eve made the usage of `CUSTOM_HATEOAS_PLANNING` in `update_returned_document` no longer work).

An example can be found in the fix for `Assignment Lock` endpoint (https://github.com/superdesk/superdesk-planning/commit/dce0e19610ca8ac7accf5eeedf8916ca972da55a#diff-fc7779db17246c58f81fa7f4f655b8a8)

MarkLark86 (approved):

PR #1499, opened 2020-11-25
[SDAN-680] fix updating of events in newshub

petrjasek:
there was `_current_version` sent in ninjs
which is huge number and can't fit into elastic integer.

SDAN-680

### DIFF
--- a/server/planning/common.py
+++ b/server/planning/common.py
@@ -9,8 +9,8 @@
 # at https://www.sourcefabric.org/superdesk/license
 
 import re
+import time
 from flask import current_app as app
-from datetime import datetime, timedelta
 from collections import namedtuple
 from superdesk.resource import not_analyzed, build_custom_hateoas
 from superdesk import get_resource_service, logger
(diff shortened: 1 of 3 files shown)


petrjasek:
not sure actually why we use such a number in https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/common.py#L319

MarkLark86:
> not sure actually why we use such a big number in https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/common.py#L319

This seems bizarre to me too.
What if we calculate seconds since a much newer date instead, i.e.:

```python
>>> int((datetime.utcnow() - datetime(2018, 1, 1)).total_seconds())
91575300
```

Or is there a better way we can generate versions?

petrjasek:
we can use standard [timestamp](https://docs.python.org/3/library/time.html#time.time). in superdesk it's a counter and it's populated automatically if versioning is enabled

MarkLark86 (approved):

PR #1503, opened 2020-12-03
[SDCP-388] Add language attribute to ColouredValueInput component

MarkLark86:
Fix other translation bugs
Fix Coverage form using Planning language not the Coverage language


client/components/Assignments/AssignmentPreviewContainer/_test.jsx
```diff
@@ -113,7 +113,7 @@ describe('<AssignmentPreviewContainer />', () => {
             expect(topTools.contains(
                 <span
                     className="priority-label priority-label--1 sd-list-item__inline-icon"
-                    data-sd-tooltip="Priority: High"
+                    data-sd-tooltip="Priority: {{ name }}"
```
petrjasek, line 116: in tests gettext returns the string untouched?

MarkLark86: When bootstrapping tests we don't instantiate Superdesk (or angular services).
Instead we mock these interfaces required (https://github.com/superdesk/superdesk-planning/blob/develop/client/tests.js#L26).

We could improve this at a later date by getting using get-superdesk-api-implementation.tsx, or a mocked version thereof

petrjasek: np, I was just checking

petrjasek (approved):

PR #1509, opened 2020-12-22
[SDCP-404] Add more fields to search and filter panels

MarkLark86:


MarkLark86:
Closing as a new PR has been created with client side only commits

PR #1511, opened 2021-01-06
[SDCP-404] [Server] Support new search params in planning types and filters

MarkLark86:


MarkLark86:
This is the server side code extracted from the bigger PR https://github.com/superdesk/superdesk-planning/pull/1509

server/planning/events/events_schema.py
```diff
@@ -351,5 +351,31 @@
...
+
+    # HACK: Add coverages.planning.slugline to elastic mapping
+    # Otherwise searching slugline in combined view fails on events type
+    'coverages': {
+        'type': 'list',
+        'mapping': {
+            'type': 'nested',
+            'properties': {
+                'planning': {
+                    'type': 'object',
```
petrjasek, line 364: might be good to set this not dynamic to avoid indexing other fields

server/planning/search/eventsplanning_filters.py
```diff
@@ -12,38 +12,200 @@
...
-    'calendars': {
-        'type': 'list',
-    },
-    'places': {
-        'type': 'list',
+    'item_type': {
+        'type': 'string',
+        'allowed': tuple(ITEM_TYPES),
+        'default': ITEM_TYPES.COMBINED,
+        'mapping': not_analyzed,
```
petrjasek, line 55: not sure why there is a mapping here, does it get saved to elastic?

petrjasek (approved):

server/planning/search/eventsplanning_filters.py
MarkLark86: Elastic is not used on this resource. Considering mappings are added here through the use of `metadata_schema` I decided to add mappings everywhere for this schema.

I could remove all manual mappings if you think that's better (and just leave the mappings added through `metadata_schema`)?

server/planning/events/events_schema.py
MarkLark86: Good point, will update this tomorrow.
I didn't like this hack, but the other alternative was to send two queries to elasticsearch.

server/planning/search/eventsplanning_filters.py
petrjasek: yep that should be fine

PR #1513, opened 2021-01-12
[SDCP-408] [Server] Scheduled exports of event and planning filters

MarkLark86:


server/planning/search/planning_search.py
```diff
@@ -70,35 +71,26 @@ def get(self, req, lookup):
...
-            doc = format_doc(hit, schema, dates)
-            build_custom_hateoas({'self': {'title': doc['_type'], 'href': '/{}/{{_id}}'.format(doc['_type'])}}, doc)
-            docs.append(doc)
-
-        return ElasticCursor(hits, docs)
+    def _format_nested_dates(self, doc):
+        if doc.get('_type') == 'events' and doc.get('dates'):
+            if doc['dates'].get('start'):
+                doc['dates']['start'] = parse_date(doc['dates']['start'])
+            if doc['dates']['end']:
```
GyanP, line 90: I think It would be good if you use `doc['dates'].get('end')` instead of `doc['dates']['end']`

GyanP (changes requested):

server/planning/search/planning_search.py
GyanP, line 90: I think It would be good if you use `doc['dates'].get('end')` instead of `doc['dates']['end']`

petrjasek:
I wonder if we could resuse some parts of report scheduling from core where we use
https://pypi.org/project/croniter/ , but not sure if the requirements are the same

petrjasek (approved):

server/planning/search/planning_search.py
MarkLark86: Good catch

MarkLark86:
> I wonder if we could reuse some parts of report scheduling from core where we use
> https://pypi.org/project/croniter/ , but not sure if the requirements are the same

This might be a good improvement down the line. I had a quick look at superdesk-core, there are multiple functions that do similar things (working out next schedule date etc). We also have scheduling in Analytics (which uses the same logic I've implemented here).

So it might be better to merge all 5 scheduling systems into a single system as an improvement later

petrjasek:
yep doing some refactoring in core first would make sense

> I also didn't like the idea of storing the schedule in a very specific format, i.e. using cron format. I believe we should not tie ? down our DB schema to a particular implementation, instead we can create a converter from our DB schema to the usage (in this case cron). I could change this code to have this converter and use croniter instead.

I would always prefer some kind of standard format to our custom, that's like reinventing the wheel, but there might be more descriptive format than cron for sure

MarkLark86:
Makes sense. Would you mind if I merge this one, and we can discuss later the specifics of a more generalised solution that will work everywhere?

petrjasek:
you can go ahead and merge

PR #1515, opened 2021-01-13
Improve translations for CP

tomaskikutis:
SDCP-425

### DIFF
--- a/client/components/Locations/LocationsSubNav.jsx
+++ b/client/components/Locations/LocationsSubNav.jsx
@@ -34,8 +34,8 @@ export class LocationsSubNavComponent extends React.Component {
                 />
                 <Dropdown
                     buttonLabel={this.props.searchType ? gettext('Search') : gettext('Browse')}
-                    items={[{label: 'Search', action: this.props.search},
-                        {label: 'Browse', action: this.props.browse}]}
+                    items={[{label: gettext('Search'), action: this.props.search},
+                        {label: gettext('Browse'), action: this.props.browse}]}
                     tooltip={gettext('Select either Search or Browse the locations')}
                 />
                 {!this.props.editOpen && (
--- a/client/components/Locations/LocationsSubNav.jsx
+++ b/client/components/Locations/LocationsSubNav.jsx
@@ -5,7 +5,7 @@ import {SearchBox, Button} from '../UI';
 import {connect} from 'react-redux';
 import * as selectors from '../../selectors';
 import * as actions from '../../actions';
-import {gettext} from '../../utils';
+import {superdeskApi} from '../../superdeskApi';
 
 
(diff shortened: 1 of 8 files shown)


client/components/Locations/LocationsSubNav.jsx
```diff
@@ -34,8 +34,8 @@ export class LocationsSubNavComponent extends React.Component {
                 />
                 <Dropdown
                     buttonLabel={this.props.searchType ? gettext('Search') : gettext('Browse')}
-                    items={[{label: 'Search', action: this.props.search},
-                        {label: 'Browse', action: this.props.browse}]}
+                    items={[{label: gettext('Search'), action: this.props.search},
```
MarkLark86, line 37: Inside functions & classes we can use the `SuperdeskApi.localization.gettext` instead. The following is the recommended way to use translation in the `Superdesk-Planning` module:

```js
import {superdeskApi} from '../../superdeskApi';

export class SomeComponent extends React.Component<IProps> {
    render() {
    }
}
```

I believe the only time we can't guarantee the api will be initialised is within AngularJS code.

MarkLark86:
The client unit tests are failing to run: https://github.com/superdesk/superdesk-planning/pull/1515/checks?check_run_id=1697816266

client/components/Locations/LocationsSubNav.jsx
tomaskikutis: That's what I tried to use initially.

In reality, it's not inside functions and classes that `superdeskApi.localization.gettext` works. It depends when those functions are called and classes initialized. If it is used before [api initializes](https://github.com/superdesk/superdesk-planning/blob/develop/client/index.js#L131) - it won't work. It's the case when registering activities.

MarkLark86: > That's what I tried to use initially.

Did this method not work in React Components (using the Superdesk API for localization)? I know using the API will not work in AngularJS init functions (which your proposal fixes), but it would be good if could guarantee the API is initialised and ready for use before the React root node is rendered.

We use `ReactDOM` from within AngularJS Controllers, such as https://github.com/superdesk/superdesk-planning/blob/develop/client/controllers/PlanningController.js#L48. I know it's not ideal, but we could use something like `setInterval` until the Planning local `superdeskApi` value has it's value defined, and only then render the root node.

tomaskikutis: `superdeskApi` initializes before `angular.bootstrap` is called. Controller code is ran afterwards, so translations using the API will work in constructors of controllers and everywhere later in the call stack including React components where a root node is rendered from inside the controller. I will update the react code to use the API.

MarkLark86 (approved):

PR #1517, opened 2021-01-15
FIX: When spiking a series of recurring events some were not affected. [SDESK-5733]

GyanP:


client/components/ItemActionConfirmation/UpdateMethodSelection.jsx
```diff
@@ -32,7 +33,7 @@ export const UpdateMethodSelection = ({
             </div>
         )}
 
-        { relatedPlannings.length > 0 && (
+        { relatedPlannings.length > 0 && eventsInUse.length <= 0 && (
```
MarkLark86, line 36: This causes a crash when attempting to cancel an Event in the series (attempted in the associated test instance)
`TypeError: Cannot read property 'length' of undefined`.

It might be better to not provide the `eventsInUse` but instead to not provide `relatedPlanning` from the `SpikeEventComponent`, which in turn will not show the related plannings warning

MarkLark86 (changes requested):

client/components/ItemActionConfirmation/UpdateMethodSelection.jsx
GyanP: That makes sense.

MarkLark86 (approved):

PR #1518, opened 2021-01-15
check if tests are passing without changes

tomaskikutis:


MarkLark86:
@tomaskikutis changes from client-core https://github.com/superdesk/superdesk-client-core/pull/3762 (remove the copied folder after patching) does not work in fireq after restarting the instance.

```
Error: ENOTEMPTY: directory not empty, rmdir '/opt/superdesk/client/patches'
```

I believe this may be because FireQ is using node v10.x:

```
+ node --version
v10.16.3
+ npm --version
6.11.3
```

Until FireQ is upgraded could we maybe use the following instead in https://github.com/superdesk/superdesk-client-core/blob/develop/tasks/patch-package.js:

```js
function removeNonEmptyDirectory(path) {
    execSync('rm -Rf ' + path, {stdio: 'inherit'});
}
```

And replace all usage of `fs.rmdirSync` with `removeNonEmptyDirectory`

PR #1519, opened 2021-01-19
[SDCP-409] [Client] Scheduled exports of search filters

MarkLark86:
Depends on https://github.com/superdesk/superdesk-client-core/pull/3764

### DIFF
--- a/client/components/EventsPlanningFilters/ManageFiltersModal.tsx
+++ b/client/components/EventsPlanningFilters/ManageFiltersModal.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 
 import {superdeskApi} from '../../superdeskApi';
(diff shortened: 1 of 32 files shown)


client/components/EventsPlanningFilters/ManageFiltersModal.tsx
```diff
@@ -41,6 +42,17 @@ const mapDispatchToProps = (dispatch) => ({
             autoClose: true,
         },
     })),
+    deleteFilterSchedule: (filter) => dispatch(actions.showModal({
+        modalType: MODALS.CONFIRMATION,
+        modalProps: {
+            body: 'Are you sure you want to delete this schedule?',
```
petrjasek, line 48: gettext missing

client/components/fields/editor/Days.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export class EditorFieldDays extends React.PureComponent<IProps> {
+    render() {
+        const {gettext} = superdeskApi.localization;
+
+        return (
+            <EditorFieldCheckboxGroup
+                field={this.props.field ?? 'week_days'}
+                label={this.props.label ?? gettext('Week Days')}
+                options={[{
```
petrjasek, line 20: I think we will need some config for starting day of the week, maybe we even have it somewhere

client/components/fields/editor/ScheduleHour.tsx
```diff
@@ -0,0 +1,55 @@
...
+        return (
+            <EditorFieldSelect
+                field={this.props.field ?? 'hour'}
+                label={this.props.label ?? gettext('Hour')}
+                labelField="label"
+                keyField="value"
+                valueAsString={true}
+                defaultValue={-1}
+                options={[
+                    {value: -1, label: gettext('Every Hour')},
```
petrjasek, line 24: actually not sure if we should do this via gettext, might be better to use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString

petrjasek (approved):

PR #1527, opened 2021-01-29
[SDCP-439] fix: Search fields in Filters editor doesnt work

MarkLark86:


MarkLark86:
There's a failing e2e test here that is fixed in https://github.com/superdesk/superdesk-planning/pull/1529

GyanP (approved):

PR #1535, opened 2021-02-05
[SDCP-443] Support search filters in all views

MarkLark86:


client/components/Main/FilterSubnavDropdown.tsx
```diff
@@ -0,0 +1,266 @@
...
+import {Dropdown} from '../UI/SubNav';
+
+interface IDropdownItem {
+    id?: string;
+    label?: string;
+    action?(): void;
+    disabled?: boolean;
+    icon?: string;
+    divider?: boolean;
+}
```
petrjasek, line 23: would probably go rather for something like:
```ts
interface IDropdownAction {
    id: string;
    label: string;
    action: () => void;
    ...
}

interface IDropdownDivider {
    divider: true;
}

type IDropdownItem = IDropdownAction | IDropdownDivider;
```

so it wouldn't need everything optional

client/components/Main/FilterSubnavDropdown.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+        const items: Array<IDropdownItem> = [{
+            label: gettext('All Planning Items'),
+            action: () => planningApi.ui.list.changeAgendaId(AGENDA.FILTER.ALL_PLANNING),
+        }, {
+            label: gettext('No Agenda Assigned'),
+            action: () => planningApi.ui.list.changeAgendaId(AGENDA.FILTER.NO_AGENDA_ASSIGNED),
+        }];
+
+        const enabledAgenads = (this.props.agendas ?? []).filter(
```
petrjasek, line 157: typo in Agen**AD**s

client/components/Main/FilterSubnavDropdown.tsx
```diff
@@ -0,0 +1,266 @@
...
+            (agenda) => agenda.is_enabled
+        );
+        const disabledAgenads = (this.props.agendas ?? []).filter(
```
petrjasek, line 160: typo in Agen**AD**s

petrjasek (approved):

PR #1536, opened 2021-02-08
[SDCP-450] Dynamically hide contradicting search param inputs

MarkLark86:


client/components/Main/SearchPanel.tsx
```diff
@@ -99,7 +91,7 @@ export class SearchPanelComponent extends React.Component<IProps, IState> {
     }
 
     search() {
-        this.props.search(this.props.activeFilter, this.state.params);
+        planningApi.ui.list.search(this.state.params);
```
petrjasek, line 94: how does this actually work without using dispatch?

petrjasek: ok I found you import dispatch in the action, why is that?

petrjasek (approved):

client/components/Main/SearchPanel.tsx
MarkLark86: I'm trying to make the APIs (anything under https://github.com/superdesk/superdesk-planning/tree/develop/client/api) be able to be called as normal functions.

Not everything needs to be a redux action dispatcher, and in the past this had made things more complex. The idea is to slowly move things away that are higher level (such as UI actions, server interactions) away from redux action dispatches.

It would be good if we can keep these action dispatchers to purely updating the store, having a more well defined Planning API using the typescript interfaces etc (and being able to be called from anywhere)

petrjasek: this is probably out of scope here, but I would like to know what made things more complex and what is the idea actually, seems like it's importing the store and calling dispatch from actions rather than calling the action via store, do I get it right?

MarkLark86: Previously the only way we could access superdesk services (api, notify, $timeout, $location etc) was via Redux thunk arguments (provided when we construct the store), i.e.
```js
function doSomething(item) {
    return (dispatch, getState, {api, notify}) => (
        api.save('resource', item).then(
            (response) => {
                notify.success('Yay');
                return response;
            }, (error) => {
                notify.error('Noooo');
                return Promise.reject(error);
            }
        )
    )
}
```

Now that we have access to the SuperdeskAPI, we no longer require this format to access core functionality.

This meant that ALL ui interactions had to use dispatch actions, regardless if they change state or not (https://github.com/superdesk/superdesk-planning/blob/develop/client/actions/events/ui.ts#L145).

Some React Components also needed to `connect` to the store, when all they required was dispatch props, i.e. where `mapPropsToState` is never used (https://github.com/superdesk/superdesk-planning/blob/develop/client/components/Coverages/CoverageEditor/CoverageFormHeader.jsx)

Another example is when reacting to websocket notifications, we required the store so that we can dispatch actions to reload a list, for example. If using this newer Planning API, these reactions to websocket messages would not require the store, but simply call `planningApi.ui.list.reloadList()`.

It's also easier to mock API functions than it is to mock dispatch actions (https://github.com/superdesk/superdesk-planning/blob/develop/client/utils/testUtils.ts#L21).

So what I would like to do is to change all higher level UI interactions to use the Planning API, and internally it will use dispatch actions to update the state. And only use `connect` in React Components to connect to state. Otherwise we would have to make ALL ui interactions to be dispatch actions, regardless if they change state or not.

MarkLark86: > this is probably out of scope here, but I would like to know what made things more complex and what is the idea actually, seems like it's importing the store and calling dispatch from actions rather than calling the action via store, do I get it right?

Not all of these UI API functions will require the redux dispatch either, some function migrate later might just lock an item then display a modal (i.e. when rescheduling an Event). In this case the websocket notifications are in charge of updating the store if the action is completed

petrjasek: so this is about avoiding to use dispatch to get hold of services? makes sense

PR #1537, opened 2021-02-08
FIX: Planning items belonging to events disappear once event repetitions are updated. [SDESK-5510]

GyanP:


server/planning/search/planning_search.py
```diff
@@ -75,7 +75,6 @@ def get(self, req, lookup):
                     app.config['ITEMS']: [
                         doc
                         for doc in docs
-                        if doc['_type'] == resource
```
MarkLark86, line 78: I have noticed that sometimes `_type` is missing but `type` is not. Check here for example (https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/search/planning_search.py#L87)

MarkLark86 (approved):

PR #1538, opened 2021-02-08
FIX: Custom layout button was disabled

GyanP:


client/components/fields/editor/ExportTemplate.tsx
```diff
@@ -23,6 +24,7 @@ export class EditorFieldExportTemplateComponent extends React.PureComponent<IPro
             this.props.templates :
             this.props.templates.filter(
                 (filter) => filter.type === this.props.itemType
+                    || ITEM_TYPE.EVENT || ITEM_TYPE.PLANNING
```
MarkLark86, line 27: Won't this also provide `Planning` templates in the list when the filter is for `Events` (and vice-versa)?
Considering the problem is with Events/Event, maybe something like this would suffice:

```js
(filter) => (
    filter.type === this.props.itemType ||
    (filter.type === ITEM_TYPE.EVENT && this.props.itemType === 'event')
)
```

GyanP: Good Catch.

client/interfaces.ts
```diff
@@ -1197,7 +1197,7 @@ export interface IEventsPlanningContentPanelProps {
 
 export interface IPlanningExportTemplate extends IBaseRestApiResponse {
     name: string;
-    type: 'event' | 'planning' | 'combined';
+    type: 'events' | 'planning' | 'combined';
```
MarkLark86, line 1200: Does this cause lint error? As the `planning_export_template` schema registered with Eve is actually `event`:
https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning_export_templates.py#L29

GyanP: yes, it was causing a lint error, that's why I changed it.

MarkLark86: I chose the incorrect type name for the `events_planning_filters` resource. We'll have to tackle this with another ticket (along with a migration script to fix it). This should be okay for now

MarkLark86 (approved):

PR #1539, opened 2021-02-09
[SDCP-464] fix(translation): Use CV not item value to show names

MarkLark86:


GyanP (approved):

client/components/fields/preview/SimpleList/Calendars.tsx
```diff
@@ -1,18 +1,33 @@
...
-        const calendarNames = (get(this.props.item, field) || [])
-            .map((calendar) => calendar.name)
-            .join(', ');
+        const calendarNames = getVocabularyItemNames(
+            get(this.props.item, field) || [],
+            this.props.calendars,
+            'qcode',
+            'name',
+            this.props.language
+        );
```
petrjasek, line 30: seems like the usage is pretty much the same, so maybe there could be some higher level helper to avoid some duplication

petrjasek (approved):

PR #1544, opened 2021-02-12
[SDCP-467][SDESK-5792] fix(search): Location and lock_state not filtering results

MarkLark86:


client/actions/events/api.ts
```diff
@@ -113,99 +112,49 @@ const unspike = (events) => (
...
-            filter_id: filter_id || selectors.main.currentSearchFilterId(getState()),
-        })
+        return planningApis.events.search(eventParamsToSearchParams({
+            ...params,
+            itemIds: itemIds,
+            timezoneOffset: getTimeZoneOffset(),
+            spikeState: params.spikeState ?? 'draft',
+            onlyFuture: params.onlyFuture ?? true,
+            page: params.page ?? 1,
+            maxResults: params.maxResults ?? MAIN.PAGE_SIZE,
```
petrjasek, line 155: just wondering why not handle all those defaults in the helper

petrjasek (approved):

client/actions/events/api.ts
MarkLark86: These were left over from the previous code while migrating to the helpers. I'll create a new commit to move these

PR #1548, opened 2021-02-17
FIX: Planning: Create action in the popup editor closes the popup unexpectedly. [SDESK-5776]

GyanP:


client/components/Main/ItemEditor/ItemManager.js
```diff
@@ -417,20 +419,25 @@ export class ItemManager {
...
+                            submitting: false,
+                            dirty: false,
+                            initialValues: updatedItem,
+                            diff: removeAutosaveFields(
+                                cloneDeep(updatedItem),
+                                true,
+                                true
+                            ),
+                        }, null, true);
+                    if (!appConfig.planning_auto_close_popup_editor) {
```
MarkLark86, line 437: What is the scenario here, why is the editor closing?

GyanP: It's mentioned in the ticket requirement:
If the config is set to false, clicking on "post" (or "unpost" for posted events/plannings) should close the popup.
So after updating the Item, we just close the popup.

MarkLark86 (approved):

PR #1550, opened 2021-02-21
FIX: Only the my assignment button should display in the dropdown if user doesn't have the desk privilege. [SDBELGA-448]

GyanP:


client/components/Assignments/DesksSubNavDropDown.jsx
```diff
@@ -28,21 +35,27 @@ export const DesksSubnavDropdown = ({userDesks, selectedDeskId, selectAssignment
         });
     }
 
-    items.push({divider: true});
+    if (showDeskAssignmentView) {
+        items.push({divider: true});
```
MarkLark86, line 39: This could be moved to the next if statement, no need to have two identical if statements

MarkLark86 (approved):

PR #1551, opened 2021-02-22
[SDCP-479] fix(filters): Filter schedule templates by desk

MarkLark86:
Increase page size to 200 when getting content filters

### DIFF
--- a/client/components/fields/editor/ContentTemplate.tsx
+++ b/client/components/fields/editor/ContentTemplate.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {connect} from 'react-redux';
+import {get} from 'lodash';
 
 import {superdeskApi} from '../../../superdeskApi';
 import {IDesk} from 'superdesk-api';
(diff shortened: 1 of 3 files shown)


client/components/fields/editor/ContentTemplate.tsx
```diff
@@ -18,20 +19,65 @@ const mapStatToProps = (state) => ({
...
 export class EditorFieldContentTemplateComponent extends React.PureComponent<IProps> {
-    render() {
-        const {gettext} = superdeskApi.localization;
-        const deskId = this.props.desk?._id;
-        const templates = deskId == null ?
+    get field(): string {
+        return this.props.field ?? 'content_template';
+    }
+
+    getAvailableTemplates(): Array<IContentTemplate> {
```
petrjasek, line 26: I wouldn't mix the syntax, both methods are getters but each one defined differently, lets keep it consistent

MarkLark86: Good point, I'll fix that

petrjasek (approved):

PR #1553, opened 2021-02-23
[SDESK-5680] fix(react-redux): Replace deprecated withRef attribute with forwardRef

MarkLark86:
Depends on https://github.com/superdesk/superdesk-client-core/pull/3810


MarkLark86:
Merging, as this was ported from the `sams` branch from PR https://github.com/superdesk/superdesk-planning/pull/1500

PR #1555, opened 2021-03-01
[SDNTB-673] Support filtering/sorting by created/updated

MarkLark86:


petrjasek:
not sure why we should use `_created` and `_updated` rather than `firstcreated` and `versioncreated` fields, any reason for that?

MarkLark86:
> not sure why we should use `_created` and `_updated` rather than `firstcreated` and `versioncreated` fields, any reason for that?

I'm not sure why we have 4 fields for the same thing. So I chose `_created` and `_updated`. Is there any difference between these fields?

petrjasek:
that `_created` and `_updated` is automatically set by eve, while the others are handled manually. so we can ingest those and use them in the output without being changed

petrjasek (approved):

PR #1561, opened 2021-03-17
[SDCP-499] fix mapping of Address from OpenStreetMaps to local DB

MarkLark86:


petrjasek (approved):
would be nice to have those conversions from js(x) to ts(x) in a separate commit, sometimes github doesn't get it

PR #1565, opened 2021-03-31
Assignments widget in the toolbar

tomaskikutis:
SDNTB-674


client/planning-extension/src/assignments-overview/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+                (event) => {
+                    const {resource, _id} = event.detail.extra;
+
+                    this.handleContentChanges([{changeType: 'deleted', resource: resource, itemId: _id}]);
+                },
+            ),
+        );
+    }
+
+    handleContentChanges(changes: Array<IResourceChange>) {
```
petrjasek, line 130: not sure if I'm missing something but seems like it's always just single change, so why there is array?

client/planning-extension/src/assignments-overview/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+        const state = this.state;
+
+        if (state.loading === true) {
+            return;
+        }
+
+        const {assignments} = state;
+
+        const refetchDesks = changes.find(({resource}) => resource === 'desks');
```
petrjasek, line 139: I guess this only makes sense if new desk is created, but on update there is no need to fetch them all

client/planning-extension/src/assignments-overview/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+        const refetchContentTypes = changes.find(
+            ({resource, itemId}) => resource === 'vocabularies' && itemId === 'g2_content_type',
+        );
+        const refetchAssignments = changes.find(
+            ({changeType, resource, itemId}) =>
+                (resource === 'assignments' && (changeType === 'created' || changeType === 'deleted'))
+                || (
+                    resource === 'assignments'
+                    && changeType === 'updated'
+                    && assignments.find(({_id}) => _id === itemId) != null
```
petrjasek, line 149: here it should also fetch single one

client/planning-extension/src/extension_bridge.ts
```diff
@@ -0,0 +1,21 @@
...
+        };
+        components: {
+            SluglineComponent: React.ComponentType<{assignment: IAssignmentItem}>;
+            DueDateComponent: React.ComponentType<{assignment: IAssignmentItem}>;
+            StateComponent: React.ComponentType<{assignment: IAssignmentItem}>;
+        };
+    };
+}
+
+export const extensionBridge = (window as unknown as any)['extension_bridge'] as IExtensionBridge;
```
petrjasek, line 21: can you explain why do we need this?

client/planning-extension/src/assignments-overview/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+import {AssignmentsOverviewListItem} from './assignments-overview-list-item';
+
+const {addWebsocketMessageListener} = superdesk;
+
+const DropdownTree = superdesk.components.getDropdownTree<IAssignmentItem>();
+const {queryRawJson, findOne} = superdesk.dataApi;
+const {GroupLabel, IconBig, TopMenuDropdownButton} = superdesk.components;
+
+interface IState {
+    loading: false;
```
MarkLark86, line 22: What is the benefit of using the values `false` or `true` here instead of the typical `boolean`?

client/planning-extension/src/assignments-overview/index.tsx
MarkLark86: Would it be beneficial to have an interface in the `superdeskApi` to  **fetch and subscribe** to resources?

Where `resource` is required, and `itemId` is optional, and the implementation can reside in Superdesk (internally subscribing to websocket notifications)?

This would help later on when we add a caching store (or something the like of). The `superdeskApi` interface would not change, but all uses of it would automatically get cached results. Otherwise we would have to come back here and implemen caching here as well

client/planning-extension/src/assignments-overview/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+        const itemsCount = assignments.length;
+        const grouped = groupBy(assignments, (item) => item.assigned_to.desk);
+
+        return (
+            <DropdownTree
+                groups={Object.keys(grouped).map((deskId) => ({
+                    render: () => (
+                        <GroupLabel>
+                            <Badge type="highlight" text={grouped[deskId].length.toString()} />
+                            <span style={{marginLeft: 6}}>{desks[deskId].name}</span>
```
MarkLark86, line 207: Would a margin left of 8 pixels work? (in accordance with https://ui-framework.superdesk.org/#/design/layout-principles)
If so, it might be beneficial to use the classname `sd-margin-l--1`

client/planning-extension/src/assignments-overview/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+                                    onClick();
+                                }
+                            }}
+                            active={isOpen}
+                            disabled={itemsCount < 1}
+                            pulsate={false}
+                            data-test-id="toggle-button"
+                        >
+                            <Badge type="highlight" text={itemsCount.toString()}>
+                                <span style={{color: isOpen ? '#3783A2' : 'white'}}>
```
MarkLark86, line 226: Would this work better `{color: isOpen ? 'var(--sd-slugline-color)' : 'var(--color-bg-00)'}`. 

Using vars from the ui-framework allows the framework to control these colours (especially in the case of accessibility / dark-ui)

client/planning-extension/src/assignments-overview/index.tsx
```diff
@@ -0,0 +1,245 @@
...
+                }
+                renderItem={(key, assignment, closeDropdown) => (
+                    <AssignmentsOverviewListItem
+                        key={key}
+                        assignment={assignment}
+                        contentTypes={contentTypes}
+                        onClick={closeDropdown}
+                    />
+                )}
+                wrapperStyles={{whiteSpace: 'nowrap', padding: 15, paddingTop: 0}}
```
MarkLark86, line 241: To stay in line with the UI framework's Layout Principles, padding should be `16` here

client/planning-extension/src/index.css
```diff
@@ -0,0 +1,14 @@
...
+
+.assignments-overview--item {
+    display: block;
+    width: 100%;
+    padding: 0;
+    text-align: left;
+}
+
+.assignments-overview--item + .assignments-overview--item {
+    margin-top: 6px;
```
MarkLark86, line 11: Would `margin-top: 8px` work here, to stay in line with the UI framework's Layout Principles?

client/styles/index.scss
```diff
@@ -19,24 +19,36 @@
...
+    padding-left: 0!important;
+  }
+  
+  &.label-icon--warning {
+    .sd-list-item__time__schedule {
+        color: $orange !important;
+    }
+  
+    .label {
+        margin-left: 0.75rem;
```
MarkLark86, line 46: To stay in line with the UI framework's Layout Principles, this should be `margin-left: 0.8rem;`

client/utils/index.ts
```diff
@@ -212,9 +212,17 @@ export const createStore = (params = {}, app = planningApp) => {
     let _compose = compose;
 
     if (process.env.NODE_ENV !== 'production') {
-        // activate logs actions for non production instances.
-        // (this should always be the last middleware)
-        middlewares.push(createLogger());
+        const devtools = localStorage.getItem('devtools');
```
MarkLark86, line 215: What are you trying to achieve here? Will this remove the redux logs from the Browsers console?

client/planning-extension/src/assignments-overview/index.tsx
tomaskikutis: Good point. I'm not sure how a subscribe method would work, it seems it would need a react wrapper the same way redux needs one. I have instead added a few higher order components that fetch data and keep it up to date. When we have cache we can connect it to the implementation of these HOCs.

client/utils/index.ts
MarkLark86: Does this then require the redux extension to work? I like to use the Browsers console, and it also has the ability to search so you can find what you need.

Could we use an env. variable to control this instead?

tomaskikutis: It doesn't require a redux extension. It simply reads an array from local storage. Env var would mean that the config would need to be stored on the server and a network request would be required to get it, which seems like an overkill for such a simple thing.

Here's a line of code you can run in the console to enable it `localStorage.setItem('devtools', '["redux-logger"]')`

MarkLark86: In that case, could you add a section in this repos README.md file for redux-logs, and the instruction on enabling it please?

MarkLark86:
Superdesk fails to load in the browser, with the following error:

image

MarkLark86 (approved):

PR #1570, opened 2021-04-29
[SDESK-5905] Editor Bookmarks and embedded coverage form

MarkLark86:


MarkLark86:
@petrjasek @GyanP 
Assigning this as Draft for now while I work on other tickets. I will come back and write E2E tests for the new functionality

client/api/autosave.ts
```diff
@@ -0,0 +1,33 @@
+import {IEventOrPlanningItem, IPlanningAPI} from '../interfaces';
+import {planningApi} from '../superdeskApi';
+import * as actions from '../actions';
+
+function getAutosaveItemById(
+    type: IEventOrPlanningItem['type'],
+    id: IEventOrPlanningItem['_id']
+): Promise<IEventOrPlanningItem | null> {
+    return planningApi.redux.store.dispatch<any>(
```
petrjasek, line 9: one day you have to explain me why we're doing this instead of using redux connect, I guess we already talked about it, but I'm still curious (meaning I can't remember why :wink: )

client/api/editor/item_events.ts
```diff
@@ -0,0 +1,281 @@
...
+import {AddPlanningBookmark, AssociatedPlanningsBookmark} from '../../components/Editor/bookmarks';
+import {RelatedPlanningItem} from '../../components/fields/editor/EventRelatedPlannings/RelatedPlanningItem';
+
+
+export function getEventsInstance(type: EDITOR_TYPE): IEditorAPI['item']['events'] {
+    function getGroups() {
+        const {gettext} = superdeskApi.localization;
+        const profile = planningApi.events.getEditorProfile();
+
+        return {
```
petrjasek, line 27: wondering why not use a list for these, id is duplicating the dict key anyhow and there is extra index

petrjasek (approved):
looks good, but it's kind of big so I could have missed something

client/api/editor/item_events.ts
MarkLark86: This is to make it easier to determine if a bookmark should be rendered, in `getGroupsForItem` function (in the same file):

``` js
groups = getGroups();

return {
    bookmarks: [{
        id: 'location',
        disabled: !groups.location.fields.length,
    }],
    groups: Object.values(groups),
}
```

client/api/autosave.ts
MarkLark86: The ultimate goal is to use Redux **only** for state management. Originally we used Redux as a way to pass AngularJS services as well, such as `api`, `notify`, `$timeout`, `$location` etc. This required dispatch actions that sometimes didn't even change the state.

This made accessing these functions more cumbersome by requiring access to the Redux store to execute these functions, which are sometimes used outside of React components/lifecycle (such as responding to websocket notifications).

Now that we have the `superdeskApi`, we can migrate the business logic into pure JS functions. And later an improvement to make all Redux dispatches to only handle state management.

I had a problem with the `autosave` API here though, as re-writing it with the `superdeskApi` was causing errors. I decided to, for now, pass these onto the dispatch action. Later we can improve this part of the API to use the `superdeskApi`.

petrjasek: so rather than having actions which might or might not change state we have another set of functions which might or might not change state?

client/api/editor/item_events.ts
petrjasek: I see

client/api/autosave.ts
petrjasek: I understand that part with websocket where we need to have access to store, but that's in the client probably the only part where some action is not connected to a piece of ui, so it kind of makes sense to be handled extra.

petrjasek: anyhow I don't mind it, just would be good to have it somewhere documented so it's clear how we're diverging from redux way of doing things, so we're consistent later when using it in superdesk

MarkLark86: We can work on an implementation proposal, and maybe use Planning as a prototype before moving it into Superdesk?. 

Basically it's about separating out the different parts of the system:

* UI - Basically dumb react components, no internal business logic or state
* State - And state only, absolutely nothing else
* API (logic) - handling change requests (i.e. server request), performing side affects (i.e. updating URL), executing action dispatches

The dataflow from UI components will be:
* UI -> API: Request an action to be performed
* API -> Server
* API -> State
* API -> Browser
* State -> UI: Update the state of UI components to re-render

Using the API as the backbone of the system, to control all dataflow, should make it easier to debug. It should also allow use, if need be, to change how we store and use state in the system

petrjasek: that kind of reminds me of [flux](https://facebook.github.io/flux/docs/in-depth-overview#structure-and-data-flow)

PR #1571, opened 2021-05-05
[SDCP-529] Fix various UI translations

MarkLark86:
* Patch strings in constant files after superdeskApi has initialised
* update planning.pot

### DIFF
--- a/client/components/UI/List/PubStatus.tsx
+++ b/client/components/UI/List/PubStatus.tsx
@@ -2,9 +2,10 @@ import React from 'react';
 import {OverlayTrigger, Tooltip} from 'react-bootstrap';
 import PropTypes from 'prop-types';
 
+import {superdeskApi} from '../../../superdeskApi';
+
 import {Column} from './Column';
-import {isNotForPublication, gettext} from '../utils';
-import {TOOLTIPS} from '../constants';
+import {isNotForPublication} from '../utils';
 
 /**
  * @ngdoc react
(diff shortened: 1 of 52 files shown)


client/components/UI/List/PubStatus.tsx
```diff
@@ -2,24 +2,26 @@ import React from 'react';
...
  * @description Component to show published status of an item
  */
 export const PubStatus = ({item, isPublic}) => {
+    const {gettext} = superdeskApi.localization;
     let badge;
     let title = null;
 
     if (isPublic) {
-        title = gettext(TOOLTIPS.postedState);
+        title = gettext('Posted');
```
petrjasek, line 21: why not use the tooltips here? seems like those are translated

MarkLark86: That was a change from my first pass, before I got the Tooltips. I'll update that

po/planning.pot
```diff
@@ -4,1791 +4,4338 @@ msgstr ""
...
+#: client/components/Planning/FeaturedPlanning/FeaturedPlanningModal.tsx:364
+msgid "All featured planning items are currently selected"
+msgstr ""
+
+#: client/actions/agenda.ts:42
+msgid "An agenda with this name already exists"
+msgstr ""
+
+#: client/components/ConfirmationModal.tsx:108
+msgid "Are you sure ?"
```
petrjasek, line 376: extra space

petrjasek (approved):
I'm not sure about that approach with assigning the translated labels later, imo it's a bit error prone as you can use the labels before those are translated. we can give it a try, maybe we can set some placeholder there rather then the actual string so it would be obvious if it's not used correctly

MarkLark86:
These labels are used in React components, which are rendered by an AngularJS controller. So here we can guarantee they will be set for UI purposes.

I agree, it's not a good solution. Tomas has ideas we can discuss, maybe for 2.4?

petrjasek:
sure

PR #1573, opened 2021-05-05
SDESK-5791 ADD: Location should not be created from within search form.

dev-tsoni:


client/components/fields/editor/Location.tsx
```diff
@@ -18,6 +18,7 @@ export class EditorFieldLocation extends React.PureComponent<IEditorFieldProps>
                     label={this.props.label ?? gettext('Location')}
                     value={get(this.props.item, field, this.props.defaultValue)}
                     disableSearch={true}
+                    disableAddLocation={false}
```
MarkLark86, line 21: This component should allow this option to be controlled via an optional prop, and then in setting this prop to false in the `AdvancedSearch` component: https://github.com/superdesk/superdesk-planning/blob/develop/client/components/AdvancedSearch/index.tsx#L138

MarkLark86 (approved):

PR #1597, opened 2021-06-09
SDESK-6018 ADD: Clear button does not clear the advanced search form.

dev-tsoni:


client/components/Main/SearchPanel.tsx
```diff
@@ -23,6 +23,7 @@ interface IProps {
 
 interface IState {
     params: ISearchParams;
+    key?: string;
```
MarkLark86, line 26: `key` should be a required field as it will always have a value. Therefor the `?` suffix isn't required

MarkLark86 (approved):

PR #1598, opened 2021-06-10
[SDCP-545] Accessibility improvements

tomaskikutis:
Depends on: https://github.com/superdesk/superdesk-ui-framework/pull/558 and https://github.com/superdesk/superdesk-client-core#develop.

@MarkLark86 regarding the accessible three dot menu component see [the example here](https://github.com/superdesk/superdesk-planning/compare/develop...tomaskikutis:accessibility-improvements?expand=1#diff-bec33568d08b03266013cf46fa5817caba68f64e0fac536433d7953e8a62b204R131)


client/components/Planning/PlanningItem.tsx
```diff
@@ -29,6 +29,7 @@ import {
     getItemWorkflowState,
 } from '../../utils';
 import {renderFields} from '../fields';
+import {gettext} from 'superdesk-core/scripts/core/utils';
```
MarkLark86, line 32: It would be better to use the `superdeskApi` instead of importing from `superdesk-core` directly

client/components/UI/List/Group.tsx
```diff
@@ -7,35 +7,44 @@ import classNames from 'classnames';
...
-            }
-        )}
-        style={style}
-        onScroll={onScroll}
-        ref={refNode}
-    >
-        {children}
-    </div>
-);
+export const Group = (props) => {
```
MarkLark86, line 10: In my PR, there is now the `IProps` interface defined for this component:
https://github.com/superdesk/superdesk-planning/blob/99e16ac3c2529b744b72631e57a5f58b65d1c7c5/client/components/UI/List/Group.tsx

e2e/cypress/support/planning/advancedSearch.ts
```diff
@@ -37,15 +37,15 @@ export class AdvancedSearch {
             state: new SelectMetaTerms(getSearchPanel, '[data-test-id=field-state]'),
             only_posted: new ToggleInput(
                 getSearchPanel,
-                '[data-test-id=field-posted] .sd-toggle'
+                '[data-test-id=field-posted] .sd-switch__wrapper'
```
MarkLark86, line 40: Merging `develop` should fix these e2e issues (https://github.com/superdesk/superdesk-planning/pull/1605)

MarkLark86 (approved):

PR #1611, opened 2021-06-17
SDESK-5833 ADD: About link not accessible from planning section.

dev-tsoni:


client/components/WorkqueueContainer/index.tsx
```diff
@@ -16,8 +16,9 @@ export const WorkqueueComponent = ({
     unlockAndCloseEditor,
     openEditForm,
     openConfirmationModal,
+    workQueueOpen,
```
MarkLark86, line 19: You're using `workQueueOpen` everywhere, which the name doesn't make sense.
The name should indicate what the value represents, such as `mainMenuOpen`

MarkLark86 (approved):

PR #1612, opened 2021-06-18
SDBELGA-546 ADD: Don't show manage agenda action if user does not hav…

dev-tsoni:
…e privileges.

### DIFF
--- a/client/components/Main/ActionsSubnavDropdown.tsx
+++ b/client/components/Main/ActionsSubnavDropdown.tsx
@@ -44,18 +44,16 @@ const ActionsSubnavDropdownComponent = (props) => {
     }
 
     return (
-        <React.Fragment>
-            {items.length == 0 ? null : (
-                <Dropdown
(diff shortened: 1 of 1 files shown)


client/components/Main/ActionsSubnavDropdown.tsx
```diff
@@ -42,14 +44,18 @@ const ActionsSubnavDropdownComponent = (props) => {
...
     return (
-        <Dropdown
-            icon="icon-dots-vertical"
-            label={gettext('Actions')}
-            items={items}
-            alignRight={true}
-            tooltip={gettext('Actions')}
-            aria-label={gettext('Actions')}
-        />
+        <React.Fragment>
```
MarkLark86, line 47: There's no need to use `React.Fragment` here, you can simply have:
```ts
return items.length === 0 ? null : (
```

MarkLark86 (approved):

PR #1622, opened 2021-06-28
Sdesk 5833 about link not accessible from planning section

dev-tsoni:


client/controllers/PlanningController.tsx
```diff
@@ -63,7 +74,12 @@ export class PlanningController {
         if (!workspaceChanged) {
             return Promise.resolve();
         }
-
+        if (this.store) {
```
MarkLark86, line 77: You could call `this.onMenuChange(false)` instead of duplicating that code here

MarkLark86: A quick comment here could help as well, as to why we're manually setting the `MENU_OPEN` to false.
That way in the future we know why this code was added

dev-tsoni: Okay.

MarkLark86 (approved):

PR #1627, opened 2021-06-29
[SDBELGA-544] New Global Filters privilege

MarkLark86:


server/features/search_planning.feature
```diff
@@ -627,3 +627,58 @@ Feature: Planning Search
             {"_id": "planning_4"}
         ]}
         """
+
+    @auth @wip
```
GyanP, line 631: I don't think we need this @wip.

GyanP (approved):

PR #1634, opened 2021-07-12
[SDNTB-685] Move Planning ProdAPI out of core

MarkLark86:
Enhance item responses with HATEOAS links

Requires PR in core https://github.com/superdesk/superdesk-core/pull/2235


MarkLark86:
`CI-Server / prodapi` will fail, as `_links` was previously removed in `superdesk-core`.
We'll merge this one first, as `superdesk-core` is also failing as it requires this PR

petrjasek (approved):

PR #1636, opened 2021-07-16
SDNTB-687 ADD: Export assignments in an ICS file usable in google cal…

dev-tsoni:
…endar.

### DIFF
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -338,53 +338,59 @@ class AssignmentsService(superdesk.Service):
         else:
             event_item = None
 
-        # Create the ICS file to be added to the email usable in google calendar.
-        ical = Calendar()
-        app_name = app.config['APPLICATION_NAME']
-        org_name = app.config.get('ORGANIZATION_NAME_ABBREVIATION') or app.config['ORGANIZATION_NAME']
-        language = app.config['DEFAULT_LANGUAGE'].upper()
-        ical.add('PRODID', f'-//{app_name}//{org_name}//{language}')
-        ical.add('VERSION', '2.0')
+        # Allow to create the ICS object only if there is scheduled time in the assignment.
+        # This situation won't be applicable in the production but only for the test cases.
+        if not assignment['planning'].get('scheduled'):
+            logger.error('Assignment has no scheduled date, cannot create an ICS file')
+        else:
+            # Create the ICS object to be added to the email usable in google calendar.
+            ical = Calendar()
+            scheduled_time = assignment['planning']['scheduled']
+            app_name = app.config['APPLICATION_NAME']
+            org_name = app.config.get('ORGANIZATION_NAME_ABBREVIATION') or app.config['ORGANIZATION_NAME']
+            language = app.config['DEFAULT_LANGUAGE'].upper()
+            ical.add('PRODID', f'-//{app_name}//{org_name}//{language}')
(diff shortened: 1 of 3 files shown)


server/planning/assignments/assignments.py
```diff
@@ -336,6 +338,53 @@ def send_assignment_notification(self, updates, original=None, force=False):
         else:
             event_item = None
 
+        # Create the ICS file to be added to the email usable in google calendar.
+        ical = Calendar()
+        ical.add('BEGIN', 'VCALENDAR')
+        ical.add('PRODID', '-//Superdesk//NTB//EN')
```
MarkLark86, line 344: This should be using the system configured attributes here.
For example:

```python
app_name = app.config['APPLICATION_NAME']
org_name = app.config.get('ORGANIZATION_NAME_ABBREVIATION') or app.config['ORGANIZATION_NAME']
language = app.config['DEFAULT_LANGUAGE'].upper()
ical.add('PRODID', f'-//{app_name}//{org_name}//{language}')
```

server/planning/assignments/assignments.py
```diff
@@ -336,6 +338,53 @@ def send_assignment_notification(self, updates, original=None, force=False):
         else:
             event_item = None
 
+        # Create the ICS file to be added to the email usable in google calendar.
+        ical = Calendar()
+        ical.add('BEGIN', 'VCALENDAR')
```
MarkLark86, line 343: There's no need to add `BEGIN:VCALENDAR` specifically. The library handles this for us

server/planning/assignments/assignments.py
```diff
@@ -336,6 +338,53 @@ def send_assignment_notification(self, updates, original=None, force=False):
...
+        ical.add('PRODID', '-//Superdesk//NTB//EN')
+        ical.add('VERSION', '2.0')
+        ical.add('BEGIN', 'VEVENT')
```
MarkLark86, line 346: The `icalendar` library has an `Event` class that can be used to add subcomponents to the file (https://icalendar.readthedocs.io/en/latest/usage.html#subcomponents).

This removes the need to manually add `BEGIN/END` attributes for the event. It would be better to let the library handle this for us

server/planning/assignments/assignments.py
```diff
@@ -336,6 +338,53 @@ def send_assignment_notification(self, updates, original=None, force=False):
...
+        ical.add('CLASS', 'PUBLIC')
+
+        if assignment['planning']['scheduled']:
```
MarkLark86, line 349: An Assignment should always have a scheduled date.
In any case, an Event inside an ical file is not usable if the Event does not have a start/end date/time

server/planning/assignments/assignments.py
```diff
@@ -336,6 +338,53 @@ def send_assignment_notification(self, updates, original=None, force=False):
...
+
+        UID = str(assignment['_id'])
+        url = client_url + '#/workspace/assignments?assignment=' + UID
+        summary = get_assginment_name(assignment)
+        priority = assignment['priority']
+        created = assignment['_created']
+        updated = assignment['_updated']
+
+        ical.add('UID', UID)
+        ical.add('SUMMARY;LANGUAGE=EN-US', summary)
```
MarkLark86, line 362: This entry should be using the default language as defined in `app.config` (also used in the `PRODID`)

MarkLark86 (approved):

PR #1639, opened 2021-08-05
SDNTB-700 ADD: Updated event json blocks new events to be ingested on…

dev-tsoni:
… FTP

### DIFF
--- a/server/planning/feed_parsers/superdesk_event_json.py
+++ b/server/planning/feed_parsers/superdesk_event_json.py
@@ -39,13 +39,6 @@ class EventJsonFeedParser(FileFeedParser):
         with open(file_path, 'r') as f:
             superdesk_event = json.load(f)
 
-            events_service = get_resource_service('events')
-            existing_event = events_service.find_one(req=None, guid=superdesk_event.get('guid'))
-            if existing_event:
-                logger.warn(
-                    "An event already exists with exact same Id. Updating events is not supported yet."
-                )
-                return []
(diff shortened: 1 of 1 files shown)


server/planning/feed_parsers/superdesk_event_json.py
```diff
@@ -43,9 +42,10 @@ def parse(self, file_path, provider=None):
             events_service = get_resource_service('events')
             existing_event = events_service.find_one(req=None, guid=superdesk_event.get('guid'))
             if existing_event:
-                raise ParserError.parseMessageError(
+                logger.warn(
```
petrjasek, line 45: not sure why this check is here in the first place, imo it has nothing to do with the parser, so I would remove it from here and if needed we can log that in events service on update

MarkLark86: Karel and I made the decision not to currently support updating of existing Events through the ingest. Once an Event has been ingested, the management of that Event must be through the Superdesk UI and not through further ingests.

If an ingested Event were to be changed in Superdesk (rescheduled, cancelled, metadata changed etc), then an update of this Event through an ingest would undo those changes.

That might be okay with metadata, but we also have actions that may affect other items in Superdesk/Planning (i.e. cancelling an Event will also cancel associated Planning, Coverage and Assignments).

So basically the parser skips ingesting Events that already exists.

MarkLark86 (approved):

PR #1651, opened 2021-09-06
SDESK-6122 ADD: Recurring event is causing events to compound while i…

dev-tsoni:
…ngesting.

### DIFF
--- a/server/planning/events/events.py
+++ b/server/planning/events/events.py
@@ -190,7 +190,9 @@ class EventsService(superdesk.Service):
 
             # generates events based on recurring rules
             # If _created_externally is true, generate_recurring_events is restricted.
-            if event["dates"].get("recurring_rule", None) and not event["dates"]["recurring_rule"].get("_created_externally", False):
+            if event["dates"].get("recurring_rule", None) and not event["dates"]["recurring_rule"].get(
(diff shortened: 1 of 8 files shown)


server/planning/events/events.py
```diff
@@ -192,18 +192,21 @@ def on_create(self, docs):
...
-                # Set the current Event to the first Event in the new series
-                # This will make sure the ID of the Event can be used when
-                # using 'event' from here on, such as when linking to a Planning item
-                event = recurring_events[0]
-                # And set the Planning Item from the original
-                # (generate_recurring_events removes this field)
-                event["_planning_item"] = planning_item
+
+                # If _created_externally is true, generate_recurring_events is restricted.
+                if event["dates"]["recurring_rule"].get("_created_externally", None) and not event["dates"]["recurring_rule"]["_created_externally"]:
```
MarkLark86, line 197: This would need to go further up, line 192:
```python
if event["dates"].get("recurring_rule", None) and not event["dates"]["recurring_rule"].get("_created_externally", False):
```

There is also no need to double check `"_created_externally"`, the following will suffice:
```python
not event["dates"]["recurring_rule"].get("_created_externally", False)
```

PR #1657, opened 2021-09-16
SDESK-6110 fix: Clear filter button not appearing in search in planning

dev-tsoni:


client/apps/Planning/PlanningSubNav.tsx
```diff
@@ -94,7 +94,7 @@ export class PlanningSubNavComponent extends React.PureComponent<IProps> {
                         search={this.search}
                         activeFilter={this.props.currentView}
                     />
-                    {!this.props.isViewFiltered ? null : (
+                    {this.props.fullText && (
```
MarkLark86, line 97: This button should also appear if there are any filters applied in the `Advanced filters` panel, so the solution should be applied in the `selectors.main.isViewFiltered` selector

MarkLark86 (approved):

PR #1658, opened 2021-09-20
[SDESK-6183] Allow `Reassign` or `Edit Priority` of Picture Assignments

MarkLark86:
This reverts the enforcement added in SDESK-5010 (https://github.com/superdesk/superdesk-planning/commit/0275d6ff2c577e8b12fe5b0a16cb9b4d8ab2cb83)

We should allow these actions on a Picture Assignment, even if the state is controlled externally (i.e. via AAP DC).
@marwoodandrew: do you see any issue here?


dev-tsoni (approved):

marwoodandrew (approved):
I think that should be OK

PR #1659, opened 2021-09-23
SDESK-6173 fix: Redundant attachment section when adding attachment field in planning profile

dev-tsoni:
…ield in planning profile.

### DIFF
--- a/client/components/fields/editor/EventAttachments.tsx
+++ b/client/components/fields/editor/EventAttachments.tsx
@@ -11,7 +11,7 @@ import {Row, FileInput} from '../../UI/Form';
 
 import {getFileDownloadURL} from '../../../utils';
 
-interface IProps extends IEditorFieldProps{
+interface IProps extends IEditorFieldProps {
     files: Array<IFile>;
     uploadFiles(files: Array<Array<File>>): Promise<Array<IFile>>;
     removeFile(file: IFile): Promise<void>;
@@ -77,16 +77,22 @@ class EditorFieldEventAttachmentsComponent extends React.Component<IProps, IStat
         });
     }
 
+    focus() {
+        if (this.node.current != null) {
+            this.node.current.focus();
+        }
+    }
+
     render() {
         const {gettext} = superdeskApi.localization;
         const field = this.props.field ?? 'files';
         const value = get(this.props.item, field, this.props.defaultValue ?? []);
 
         return (
-            <Row ref={this.props.refNode}>
+            <Row>
                 <div className={this.state.uploading ? 'sd-loader' : ''}>
                     <label className="form-label">
-                        {gettext('Attached Files')}
+                        {this.props.label ?? gettext('Attached Files')}
                     </label>
                     {this.state.uploading ? null : (
                         <FileInput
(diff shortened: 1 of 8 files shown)


dev-tsoni:
It also includes the similar issue in links sections in Event profile.

client/components/fields/editor/EventAttachments.tsx
```diff
@@ -91,34 +91,23 @@ class EditorFieldEventAttachmentsComponent extends React.Component<IProps, IStat
         const value = get(this.props.item, field, this.props.defaultValue ?? []);
 
         return (
-            <ToggleBox
-                ref={this.props.refNode}
-                title={this.props.label ?? gettext('Attached Files')}
-                isOpen={false}
-                onOpen={this.onOpen}
-                scrollInView={true}
-                hideUsingCSS={true} // hideUsingCSS so the file data is kept on hide/show
```
MarkLark86, line 100: Come to think of it, this needs to be addressed as well.
If a user uploads a file then collapses the ToggleBox, then I believe the file data is lost.
Could you please test this with your solution please?

client/components/fields/editor/EventAttachments.tsx
```diff
@@ -91,34 +91,23 @@ class EditorFieldEventAttachmentsComponent extends React.Component<IProps, IStat
...
-                            files={this.props.files}
-                            createLink={getFileDownloadURL}
-                            onAddFiles={this.onAddFiles}
-                            onRemoveFile={this.onRemoveFile}
-                            readOnly={this.props.disabled}
-                        />
-                    )}
-                </div>
-            </ToggleBox>
+            <div className={this.state.uploading ? 'sd-loader' : ''}>
```
MarkLark86, line 94: It would be good to encapsulate this entire field under a `Row` component, and use the class `form-label` for the label. i.e.

```typescript
<Row ref={this.props.refNode}>
    <div className={this.state.uploading ? 'sd-loader' : ''}>
        <label className="form-label">
            {gettext('Attached Files')}
        </label>
    </div>
</div>
```

This should enable the `sd-loader` to function properly, and the label to use the same look as other field labels

client/components/fields/editor/EventLinks.tsx
```diff
@@ -59,44 +59,33 @@ export class EditorFieldEventLinks extends React.PureComponent<IEditorFieldProps
...
-                            style="hollow"
-                            expand={true}
-                            icon="plus-sign"
-                            onClick={this.addLink}
-                        />
-                    )}
-                </Row>
-            </ToggleBox>
+            <Row refNode={this.node}>
+                <label className="InputArray__label side-panel__heading side-panel__heading--big">
```
MarkLark86, line 63: Same here, use `form-label` for the label's class

MarkLark86:
This following function is no longer used, it would be good to remove it:

https://github.com/superdesk/superdesk-planning/blob/78c9ee8e31b785a4c0ee7f0b6417a01d4fb4199e/client/components/fields/editor/EventAttachments.tsx#L82

client/components/fields/editor/EventAttachments.tsx
```diff
@@ -91,34 +91,23 @@ class EditorFieldEventAttachmentsComponent extends React.Component<IProps, IStat
...
+                <label className="InputArray__label side-panel__heading side-panel__heading--big">
+                    {gettext('Attached Files')}
+                </label>
+                {this.state.uploading ? null : (
+                    <FileInput
+                        ref={this.node}
```
MarkLark86, line 100: An E2E test is failing for this bookmark type. It may be due to `this.props.refNode` is no longer connected to a component.

Try replacing this line with `ref={this.props.refNode}`, and removing `node` from this class attribute

MarkLark86 (approved):

PR #1668, opened 2022-01-25
[SDESK-6367] fix: Convert between Editor3 and plain text when copying metadata

MarkLark86:


MarkLark86:
Putting this in `draft` state so I can implement something similar to what we're doing in Newshub:
https://github.com/superdesk/newsroom-core/blob/develop/assets/utils.js#L458

MarkLark86:
@GyanP This is ready for review now

GyanP (approved):

PR #1675, opened 2022-04-04
add loading indicator to assignments view

petrjasek:
SDESK-6387


MarkLark86:
To provide a familiar loading experience to Monitoring, wouldn't it be better to use `ListItemLoader` from the UI-Framework (https://ui-framework.superdesk.org/#/react/list-items)

MarkLark86:
@petrjasek If you apply the changes from https://github.com/superdesk/superdesk-planning/pull/1676/commits/d0d60aa98e33b3e55c967936a75eeed82d8e870a

This should fix the e2e tests

MarkLark86:
I just tested this in the PR test instance, and further looked at the AssignmentController, and it looks like the UI is not rendering until the list of Assignments has loaded. I think we also need to improve the initial Assignment page load experience as well.

`AssignmentController.loadWorkspace:`
https://github.com/superdesk/superdesk-planning/blob/develop/client/controllers/AssignmentController.tsx#L80

`AssignmentController.onDeskChange:`
https://github.com/superdesk/superdesk-planning/blob/develop/client/controllers/AssignmentController.tsx#L127

Do you think it would improve the UX if we ignore the response of `this.onDeskChange` in the `loadWorkspace: return Promise.all[` statement by changing from:
```typescript
.then(() => this.onDeskChange()),
```
to
```typescript
.then(() => {
    this.onDeskChange();
}),
```

This way the `AssignmentController.render: ReactDOM.render` statement will be processed while the `onDeskChange` handler is being processed (i.e. waiting for network response etc).

MarkLark86 (approved):

PR #1676, opened 2022-04-05
[SDESK-6428] fix: Locks not loading for Planning items in the past

MarkLark86:


petrjasek (approved):

petrjasek:
we would need this also on 2.4 probably, if it affect that version @MarkLark86 ?

PR #1681, opened 2022-05-06
 Fix : Improve the error message when unposting a planning item which had a coverage story also published [SDESK-6434]

GyanP:


client/actions/assignments/notifications.ts
```diff
@@ -315,13 +315,14 @@ const onAssignmentDeleteFailed = (_e, data) => (
         if (get(data, 'items.length', 0) > 0 &&
                 get(data, 'user') === currentUserId &&
                 get(data, 'session') === sessionId) {
-            const msg = data.items.map((i) => gettext('Delete of {{ type }} assignment \'{{ slugline }}\' failed',
+            const msg = data.items.map((i) => gettext(
```
MarkLark86, line 318: Sometimes the error may be different then `state == 'in_progress'`, such as: https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning.py#L1009

Maybe we should handle the errors better here:
https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning.py#L1009

MarkLark86, line 318: Sometimes the error may be different then `state == 'in_progress'`, such as: https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning.py#L1009

Maybe we should handle the errors better here:
https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning.py#L1009

server/planning/planning/planning.py
```diff
@@ -1005,9 +1005,10 @@ def delete_assignments_for_coverages(self, coverages, notify=True):
                     }
                 )
             except SuperdeskApiError:
-                logger.error("Failed to delete assignment {}".format(assign_id))
+                logger.error("There is a assignment '{}' is in progress".format(assign_id))
```
MarkLark86, line 1008: Sorry, I meant that the `SuperdeskApiError` could be raised by many different places in our code (`planning`, `assignment`, `archive` or `delivery` services/resources).

It might be good if we pass the actual error to the end user, as to provide a more accurate reason why the action failed (still displaying them as warnings though, as the original action succeeded)

MarkLark86 (approved):

PR #1686, opened 2022-05-23
[SDESK-6439] New EventsML feed parser

MarkLark86:


server/planning/feed_parsers/events_ml.py
```diff
@@ -0,0 +1,171 @@
...
+            self.parse_concept(tree, item)
+            self.parse_event_details(tree, item)
+
+            if item.get("firstcreated"):
+                item["event_created"] = item["firstcreated"]
+            if item.get("versioncreated"):
+                item["event_lastmodified"] = item["versioncreated"]
+
+            item["firstcreated"] = utcnow()
+            item["versioncreated"] = utcnow()
```
petrjasek, line 97: not sure if we should override it, might be useful later when doing auto updating based on version

petrjasek (approved):

PR #1690, opened 2022-06-09
Planning: As a SD admin I want to ingest planning items [SDESK-6480]

GyanP:


server/planning/feed_parsers/superdesk_planning_xml.py
```diff
@@ -0,0 +1,194 @@
...
+            return itemClass_node.attrib.get("qcode", "") == "plinat:newscoverage"
+        except Exception:
+            return False
+
+    def parse(self, tree: Element, provider=None):
+        self.root = tree
+
+        try:
+            guid = tree.attrib["guid"]
+            item = {
```
MarkLark86, line 62: Should probably add a check for existing Planning item here, as supporting updates is troublesome and we need to come up with a plan for this.

Similar to EventsML: https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/feed_parsers/events_ml.py#L76

server/planning/feed_parsers/superdesk_planning_xml.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def parse_news_coverage_status(self, tree, item):
+        """Parse assert tag
+
+        :param tree: tree
+        :param item: planning item
+        """
+        assert_el = tree.find(self.qname("assert"))
+        if assert_el is not None:
+            news_coverage_status = assert_el.find(self.qname("newsCoverageStatus")).get("qcode")
```
MarkLark86, line 140: It might be a good idea to use the CV for this (especially for translations, name changes etc).

Similar to EventsML `eventoccurstatus`
https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/feed_parsers/events_ml.py#L97

server/planning/feed_parsers/superdesk_planning_xml.py
```diff
@@ -0,0 +1,194 @@
...
+    def parse_genre(self, planning_elt, planning):
+        """Parse Genre tag
+
+        :param planning_elt: planning element
+        :param planning: coverage planning item
+        """
+        genre_elt = planning_elt.find(self.qname("genre"))
+        if genre_elt is not None:
+            genre_name = genre_elt.find(self.qname("name"))
+            planning["genre"] = [{"qcode": genre_elt.get("qcode"), "name": genre_name.text}]
```
MarkLark86, line 156: Same with the `genre`, try to use the CV to populate the value.

server/planning/planning/planning.py
```diff
@@ -63,6 +64,19 @@
 class PlanningService(superdesk.Service):
     """Service class for the planning model."""
 
+    def post_in_mongo(self, docs, **kwargs):
```
MarkLark86, line 67: Why are changes in `post_in_mongo` and `patch_in_mongo` needed?

GyanP: Because we needed it [here](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/commands/update_ingest.py#L674).

PR #1691, opened 2022-06-14
[SDESK-6441] Ingest rule to auto post Events or Planning

MarkLark86:
Depends on:
- https://github.com/superdesk/superdesk-core/pull/2354
- https://github.com/superdesk/superdesk-client-core/pull/4108


petrjasek:
I'm wondering if we should restrict this somehow based on application (scope). or would it make sense to allow the current one to be used for planning? or this one for regular items?

MarkLark86:
> I'm wondering if we should restrict this somehow based on application (scope). or would it make sense to allow the current one to be used for planning? or this one for regular items?

I completely agree here, though this will require further changes to core. For example, at the moment we can create an ingest for ICS and select `text` as one of the content types. IMO, the content types for an ingest should be restricted to what the ingest feed parser can supply.

We can add this idea of `scope` to feed parsers (and their ingest configs), where a `scope` can handle certain content types. Then the rest of the ingest system can be based around this `scope`. Maybe for an improvement later (in the `RuleHandlers` I have applied content types to their `can_handler` method)

petrjasek (approved):

PR #1692, opened 2022-06-16
Planning: As a SD admin I like to ingest events from the Onclusive API [SDCP-627]

GyanP:


server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -0,0 +1,87 @@
...
+    NAME = "Onclusive_api"
+    label = "Onclusive api feed"
+    service = "events"
+
+    fields = []
+
+    HTTP_AUTH = False
+    session = requests.Session()
+
+    current_date = datetime.now()
```
petrjasek, line 24: if you store this on the class it will be the same from the application start till it's restarted, it should be set in in the _update method to be up to date

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -0,0 +1,87 @@
...
+            resp.raise_for_status()
+            if resp.status_code == 200 and resp.content:
+                data = resp.json()
+                if data.get("token") and data.get("refreshToken"):
+                    TOKEN = data["token"]
+                    REF_TOKEN = data["refreshToken"]
+
+        if TOKEN:
+            current_url = "{}/api/v2/events/date?date={}".format(URL, self.current_date.strftime("%Y%m%d"))
+            latest_url = "{}/api/v2/events/latest?date={}".format(URL, self.current_date.strftime("%Y%m%d"))
```
petrjasek, line 57: I think we need to call current_url only during first ingest run and after that it should only call that latest url to get the changes. now you get everything for current date and then you get also new and update since that date which should be pretty much the same. might need also some pagination to fetch all events in the future

petrjasek: maybe we should work with some upper limit to only fetch events in next 30 days or so and not everything, that db might be huge

GyanP: I see, Is there any way you remember to check for first ingest run?
Because I think for achieving this functionality we need to define a global variable something like count.

petrjasek: it's stored on provider which you should be able to get in the service via `self._provider` https://github.com/superdesk/superdesk-core/blob/develop/superdesk/io/ingest_provider_model.py#L109

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -15,13 +14,20 @@ class OnclusiveApiService(HTTPFeedingServiceBase):
...
+        {
+            "id": "url",
+            "type": "text",
+            "label": "Feed URL",
+            "placeholder": "Feed URL",
+            "required": True,
+            "default": "https://api.forwardplanner.com/",
+        },
+        {"id": "username", "type": "text", "label": "Username", "placeholder": "Username", "required": True},
+        {"id": "password", "type": "password", "label": "Password", "placeholder": "Password", "required": True},
```
petrjasek, line 27: we should use lazy gettext for labels/placeholders

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -33,48 +39,58 @@ def _update(self, provider, update):
...
+                resp = session.post(auth_url, body)
+
+                resp.raise_for_status()
+                if resp.status_code == 200 and resp.content:
+                    data = resp.json()
+                    if data.get("token") and data.get("refreshToken"):
+                        TOKEN = data["token"]
+                        REF_TOKEN = data["refreshToken"]
+            except Exception as e:
+                logger.error("Not Found URL {}".format(URL))
```
petrjasek, line 63: not sure if the error is correct, could be different issues if you catch Exception

GyanP: What if we handle it something like credentials auth errors or we can just log the exception:
```
except Exception:
            raise CredentialsAuthError(credentials)
```

GyanP: I think for catching HTTP exceptions the best way is:
```
except requests.HTTPError as http_error:
    logger.error("Authntication Failed {}".format(http_error))
except Exception as e:
    logger.error("Exception: {}".format(e))
```
This will provide the exact error message/exception message.

GyanP:
> lets always use some timeout when calling the session otherwise this might block for a long time

Yup, that sounds right.

server/planning/feeding_services/onclusive_api_service.py
petrjasek: yep but I actually wonder why catching it, you want the update to stop on such exception not continue right?

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        session = requests.Session()
+        current_date = datetime.now()
+        URL = provider["config"]["url"]
+        USERNAME = provider["config"]["username"]
+        PASSWORD = provider["config"]["password"]
+        TOKEN = ""
+        REF_TOKEN = ""
+        # authntication get token
+        if not TOKEN:
```
petrjasek, line 50: this will be always true we don't keep instances around but we always create a new one

petrjasek: but it might make sense to store it in db (probably not the token but the refresh one which lasts for a few days)

server/planning/feeding_services/onclusive_api_service.py
GyanP: Yes, we actually want to stop the update, but I see what you mean here, you mean to say that we should just raise the exception which will just stop the execution?

petrjasek: yep any exception will pretty much stop it. so unless we want to handle it in some specific way we can just raise that or rather not catch it

server/planning/feed_parsers/onclusive.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def parse_http(self, content, provider):
+        return self.parse(content, provider)
+
+    def parse(self, content, provider=None):
+        try:
+            for event in content:
+                guid = "urn:newsml:{}:{}".format(event["createdDate"], event["itemId"])
+
+                if get_resource_service("events").find_one(req=None, guid=guid):
```
petrjasek, line 35: I think we don't need this one, imo it should or will be supported right @MarkLark86 ?

server/planning/feed_parsers/onclusive.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def parse_item_meta(self, event, item):
+        item["versioncreated"] = self.datetime(event["createdDate"])
+        item["firstcreated"] = self.datetime(event["createdDate"])
+        item["_updated"] = self.datetime(event["lastEditDate"])
+        item["slugline"] = (
+            event["summary"] if (event["summary"] is not None and event["summary"] != "") else event["description"]
+        )
+        item["definition_short"] = event["description"]
+        item["links"] = event["website1"]
```
petrjasek, line 79: imo it should be a list, so rather `[event["website1"]]`

server/planning/feed_parsers/onclusive.py
```diff
@@ -0,0 +1,116 @@
...
+        else:
+            item["occur_status"] = [
+                x for x in eocstat_map.get("items", []) if x["qcode"] == "eocstat:eos5" and x.get("is_active", True)
+            ][0]
+            item["occur_status"].pop("is_active", None)
+
+    def parse_item_meta(self, event, item):
+        item["versioncreated"] = self.datetime(event["createdDate"])
+        item["firstcreated"] = self.datetime(event["createdDate"])
+        item["_updated"] = self.datetime(event["lastEditDate"])
```
petrjasek, line 74: `_updated` is rather internal, this should go into `versioncreated`

server/planning/feed_parsers/onclusive_tests.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self.assertEqual(item["occur_status"]["qcode"], "eocstat:eos5")
+
+        self.assertIn("https://www.canadianinstitute.com/anti-money-laundering-financial-crime/", item["links"])
+
+        self.assertEqual(item["dates"]["start"], "2022-06-15T00:00:00")
+        self.assertEqual(item["dates"]["end"], "2022-06-16T00:00:00")
+        self.assertEqual(item["dates"]["tz"], "EDT")
+
+        self.assertEqual(item["slugline"], "Annual Forum on Anti-Money Laundering and Financial Crime")
```
petrjasek, line 49: this should be rather `headline`

server/planning/feed_parsers/onclusive_tests.py
```diff
@@ -0,0 +1,52 @@
...
+        self.assertEqual(item["definition_short"], "Annual Forum on Anti-Money Laundering and Financial Crime")
```
petrjasek, line 50: I think we shouldn't copy it if there is only description, we would be showing same thing 2x everywhere, so if summary is empty use description to populate headline, otherwise use summary for headline and description for definition_short

server/planning/feed_parsers/onclusive_sample.json
```diff
@@ -0,0 +1,150 @@
...
+            "tagId": 148,
+            "tagName": "Trade Conferences",
+            "tagTypeId": 7,
+            "tagTypeName": null
+        },
+        {
+            "tagId": 228,
+            "tagName": "Cyber Security and Fraud",
+            "tagTypeId": 9,
+            "tagTypeName": null
```
petrjasek, line 132: pls also parse these as subjects with `scheme=onclusive_event_types`

server/planning/feed_parsers/onclusive_sample.json
```diff
@@ -0,0 +1,150 @@
...
+            "pressContactName": "",
+            "pressContactOffice": "American Conference Institute",
+            "pressContactTelephone": "1 212 352 3220",
+            "pressContactMobile": "",
+            "pressContactEmail": "customerservice@americanconference.com",
+            "pressContactTwitter": "",
+            "pressContactLinkedIn": null
+        }
+    ],
+    "categories": [
```
petrjasek, line 77: I would parse these as subject with scheme `onclusive_categories`

petrjasek: as it doesn't really match their categories

server/planning/feed_parsers/onclusive.py
MarkLark86: @petrjasek This is similar to Events ingest, until we come up with a strategy to allow management of the data from both the source and within Superdesk we decided to not allow updates for now.

IMO we can go forth with not allowing for now (so we can get the ingest system tested), and later on re-address the issue of updating certain metadata when ingesting an item with the same ID

petrjasek: ok

petrjasek:
it seems like this would actually fetch only current day right? we might need to have this configurable, in general the idea will be to ingest next x days or rather months, we're need to be able to ingest next 12 months for testing

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -74,24 +84,32 @@ def _update(self, provider, update):
                 current_event_response.raise_for_status()
```
petrjasek, line 84: not sure if we need this if part we later fetch it from current day anyhow?

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -74,24 +84,32 @@ def _update(self, provider, update):
...
+                print("loop started", offset)
+                between_url = "{}/api/v2/events/between?startDate={}&endDate={}&limit={}".format(
+                    URL, current_date.strftime("%Y%m%d"), end_date.strftime("%Y%m%d"), offset
+                )
+                between_event_response = session.get(url=between_url, headers=headers, timeout=TIMEOUT)
+                between_event_response.raise_for_status()
+                content.append(between_event_response.json())
+
+                if between_event_response.status_code == 401:
+                    TOKEN = self.renew_token(provider, session)
```
petrjasek, line 102: it should probably also call that api again otherwise this would skip this offset and go to next one

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -74,24 +84,32 @@ def _update(self, provider, update):
...
+
+            end_date = current_date + timedelta(days=int(provider["config"]["days"]))
+            for offset in range(100, ONCLUSIVE_MAX_OFFSET, 10000):
+                print("loop started", offset)
+                between_url = "{}/api/v2/events/between?startDate={}&endDate={}&limit={}".format(
+                    URL, current_date.strftime("%Y%m%d"), end_date.strftime("%Y%m%d"), offset
+                )
+                between_event_response = session.get(url=between_url, headers=headers, timeout=TIMEOUT)
+                between_event_response.raise_for_status()
+                content.append(between_event_response.json())
```
petrjasek, line 99: might be better to parse and yield from here instead of putting it all into array and doing it only after, so we don't need to store it all in memory

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -74,52 +73,43 @@ def _update(self, provider, update):
...
                     TOKEN = self.renew_token(provider, session)
+                    between_event_response = session.get(url=between_url, headers=headers, timeout=TIMEOUT)
+
+                content = between_event_response.json()
                 if between_event_response.json() == []:
                     break
 
-            else:
-                logger.warning("some items were not fetched due to the limit")
+                parser = self.get_feed_parser(provider)
```
petrjasek, line 96: not sure if we need new instance for each loop

server/planning/__init__.py
```diff
@@ -222,6 +223,7 @@ def init_app(app):
     app.client_config["planning_use_xmp_for_pic_assignments"] = get_planning_use_xmp_for_pic_assignments(app)
     app.client_config["planning_use_xmp_for_pic_slugline"] = get_planning_use_xmp_for_pic_slugline(app)
     app.client_config["planning_auto_close_popup_editor"] = get_planning_auto_close_popup_editor(app)
+    app.client_config["onclusive_max_offset"] = get_onclusive_max_offset(app)
```
petrjasek, line 226: do you need it on client?

GyanP: Imo, May be we need it in future.

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -80,8 +82,8 @@ def _update(self, provider, update):
             end_date = current_date + timedelta(days=int(provider["config"]["days_to_ingest"]))
             for offset in range(100, ONCLUSIVE_MAX_OFFSET, 1000):
```
petrjasek, line 83: since you have that LIMIT defined now you can use it here as the 3rd param so it's in sync as it should be

PR #1699, opened 2022-07-19
Fix : Planning: As a SD admin I like to ingest events from the Onclusive API [SDCP-627]

devketanpro:


server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -101,18 +101,13 @@ def _update(self, provider, update):
                 logger.info("Ingesting content: {} ...".format(str(between_event_response.content)[:4000]))
 
                 if hasattr(parser, "parse_http"):
```
petrjasek, line 103: are actually both options possible? imo we should limit onclusive to single parser as we do for rss

petrjasek, line 103: are there actually both options possible? imo we should limit onclusive to single parser as we do for rss

devketanpro: as much as i understand, that we have to registered a feedparser in the feeding service just like rss or we should do something like:
```
 if hasattr(parser, "parse_onclusive"):
      yield (parser.parse_onclusive(content, provider))
```

petrjasek: we can set the parser explicitly like here:
https://github.com/superdesk/superdesk-core/blob/1b3ff89a92d3cce8e1cbfe635f4c23c6e39f9a5c/superdesk/io/feeding_services/rss.py#L403
so then we know it will be using onclusive parser and the parser will have that `parse_http` method so that if is not needed anymore

petrjasek (approved):

PR #1702, opened 2022-07-20
Planning: Provide Ingest Source filter for 'Events & Planning' and 'Planning Only' views [SDESK-6528]

devketanpro:


MarkLark86 (changes requested):
There's one final missing piece, only Ingest Sources that have `Event` type enabled is showing up. We need to also include `Planning` type as well:
https://github.com/superdesk/superdesk-planning/blob/develop/client/services/PlanningStoreService.ts#L209

You can test this by editing the config of an ingest source:
image

client/services/PlanningStoreService.ts
```diff
@@ -207,7 +207,9 @@ export class PlanningStoreService {
                     ),
                     ingest: {
                         providers: get(data, 'ingest', [])
-                            .filter((p) => get(p, 'content_types', []).indexOf(ITEM_TYPE.EVENT) !== -1)
+                            .filter((p) => (
+                                get(p, 'content_types', []) === ITEM_TYPE.EVENT || ITEM_TYPE.PLANNING)
```
MarkLark86, line 211: This won't filter any of the source providers, because ` || ITEM_TYPE.PLANNING` statement is run separately and always returns `true`.

You'd need to check the `ITEM_TYPE.[EVENT|PLANNING]` existence in `content_types` separately

devketanpro: Can we use this as my second suggestion?, I think its works.
```
.filter((p) => (
           get(p, 'content_types', []).indexOf(ITEM_TYPE.EVENT) !== -1 ||
           get(p, 'content_types', []).indexOf(ITEM_TYPE.PLANNING) !== -1))
```

MarkLark86: @devketanpro Yes that will work

MarkLark86 (approved):

PR #1703, opened 2022-07-21
Fix Planning: As a SD admin I like to ingest events from the Onclusive API [SDCP-627]

devketanpro:


MarkLark86 (approved):

server/planning/feeding_services/onclusive_api_service.py
```diff
@@ -117,14 +123,21 @@ def authentication(self, TIMEOUT, session, provider):
...
         # Need to renew Token
         url = provider["config"]["url"] + "/api/v2/auth/renew"
         body = {"refreshToken": provider["config"]["refreshToken"]}
         renew_response = session.post(url=url, data=body, timeout=5)
-        renew_response.raise_for_status()
+        try:
+            renew_response.raise_for_status()
+        except Exception as e:
+            logger.error(e)
+        if renew_response.status_code == 400:
```
petrjasek, line 137: I wonder if we ever get non 400 status here?

devketanpro: IMO we had two different types of responses here. the first is 400 it occurs when our `refreshToken` is no longer valid; otherwise, we receive a 200 status.

PR #1704, opened 2022-07-27
[STTNHUB-169] fix(ingest): PlanningML missing coverage schedule and other metadata

MarkLark86:
Also added:
* `ingested` PlanningHistory
* setting: PLANNING_DEFAULT_COVERAGE_STATUS_ON_INGEST


server/planning/common.py
```diff
@@ -223,18 +223,52 @@ def remove_lock_information(item):
     item.update({LOCK_USER: None, LOCK_SESSION: None, LOCK_TIME: None, LOCK_ACTION: None})
 
 
-def get_coverage_cancellation_state():
+def get_default_coverage_status_qcode_on_ingest(current_app=None):
+    return (current_app or app).config.get("PLANNING_DEFAULT_COVERAGE_STATUS_ON_INGEST", "ncostat:int")
+
+
+def get_coverage_status_from_cv(qcode: str):
+    default_states = {
```
petrjasek, line 231: is the coverage state mandatory? if not I would rather keep it empty than using some hardcoded values, or we might need to add those to settings

server/planning/feed_parsers/superdesk_planning_xml_test.py
```diff
@@ -73,10 +73,54 @@ def test_content(self):
...
-        self.assertEqual(item["coverages"][0]["coverage_id"], "ID_TEXT_120190859")
-        self.assertEqual(item["coverages"][0]["planning"]["g2_content_type"], "text")
+
+        self.assertEqual(len(item["coverages"]), 2)
+
+        coverage = item["coverages"][0]
+        self.assertEqual(coverage["coverage_id"], "ID_TEXT_120190859")
+        self.assertEqual(coverage["workflow_status"], "draft")
+        self.assertEqual(coverage["firstcreated"], item["firstcreated"])
+        self.assertEqual(coverage["versioncreated"], datetime(2022, 5, 6, 2, 26, 27, tzinfo=tzoffset(None, 7200)))
```
petrjasek, line 95: is this converted to utc at some point?

MarkLark86: Yes, looks like Superdesk/Eve data-layer is converting it to UTC:
image

server/planning/common.py
MarkLark86: In Planning front-end we're always setting a value, and we have business logic in the back-end involving coverage state. So I would say yes it is mandatory.

I updated the code so it raises an error instead

petrjasek (approved):

PR #1705, opened 2022-08-10
[SDAAP-23] Data updates from planning are not being deployed

marwoodandrew:


MarkLark86:
Thanks for that @marwoodandrew 
Even though `data_updates` is inside the `planning` module, the reason why it isn't included is because there isn't an `__init__.py` inside that folder.

marwoodandrew:
@MarkLark86 I don't think core has a __init__.py in its folder.

MarkLark86 (approved):

PR #1710, opened 2022-08-20
Fix : Planning preview shows all Agenda items regardless [SDESK-6561]

devketanpro:


client/utils/vocabularies.ts
```diff
@@ -28,15 +28,14 @@ export function getVocabularyItemFieldTranslated(
...
+    valueField: string,
     nameField: keyof T,
     language: string
 ): Array<string> {
     if (!selected?.length) {
         return [];
     }
-
-    const values = selected.map((item) => item[valueField]);
+    const values = valueField === '_id' ? selected : selected.map((item) => item[valueField]);
```
MarkLark86, line 38: It is possible that we'd like to map items from `selected` using `valueField=='_id'`, which would be broken with these changes.

Maybe instead, if the types were changed to `valueField: keyof T | undefined` and check here if `valueField == null` to determine if we're to map `selected` or keep them as they are.

client/components/fields/preview/base/converters.ts
```diff
@@ -53,7 +53,7 @@ export function getG2ContentTypeString(
 
 export function getValuesFromCV(
     propName: string,
-    valueField: string = 'qcode',
+    valueField: string = propName === 'agendas' ? '_id' : 'qcode',
```
MarkLark86, line 56: Rather than adding a specific default `valueField` here based on the `propName`, it would be easier to maintain if the `valueField` was provided where it is used.
https://github.com/superdesk/superdesk-planning/blob/develop/client/components/fields/preview/index.ts#L40

MarkLark86 (approved):

client/utils/vocabularies.ts
devketanpro: > It is possible that we'd like to map items from `selected` using `valueField=='_id'`, which would be broken with these changes.
> 
I am not sure, but IMO its not possible because in `selected` we got only item `_id` not whole item

> Maybe instead, if the types were changed to `valueField: keyof T | undefined` and check here if `valueField == null` to determine if we're to map `selected` or keep them as they are.

I am little bit confusing in that. Would you kindly elaborate?

MarkLark86: This utility is designed to provide the translated values based on a given array of items.
By default, we're using `qcode` to get the list of values but it is plausible that we could use `_id` if given an array of objects (separate to the usage by Agendas).

Therefor, we should support supplying `valueField='_id'` while `selected` is an array of Objects.
We can support the above use case as well as an array of strings if this function is supplied `valueField=null`

MarkLark86 (approved):

PR #1712, opened 2022-08-29
[SDESK-6569] Add config to allow sending self assignment email notific…

marwoodandrew:
…ations


MarkLark86 (approved):
@marwoodandrew Can you also please update the README file with this new config
https://github.com/superdesk/superdesk-planning/blob/develop/README.md#assignments-config

PR #1714, opened 2022-09-01
[SDESK-6852] Correct date formats in assignment ical attachment

marwoodandrew:


marwoodandrew:
The resulting ical now validates against https://icalendar.org/validator.html
Change the status field to be consistent with the RFC 5545

server/planning/assignments/assignments.py
```diff
@@ -402,9 +402,10 @@ def send_assignment_notification(self, updates, original=None, force=False):
                     )
                     event["LOCATION"] = formatted_location
 
-            event["CREATED"] = created
-            event["LAST-MODIFIED"] = updated
-            event["STATUS"] = assigned_to["state"]
+            event.add("CREATED", created)
+            event.add("LAST-MODIFIED", updated)
+            event.add("DTSTAMP", updated)
+            event.add("STATUS", "CANCELED" if assigned_to["state"] == "canceled" else "CONFIRMED")
```
MarkLark86, line 408: It might be good to use `planning.common.ASSIGNMENT_WORKFLOW_STATE` named tuple when checking the state

marwoodandrew: Good point, will do!

MarkLark86 (approved):

PR #1719, opened 2022-09-07
Fix : Language field of ingested event is populated but missing language blocks posting the event [SDCP-638]

devketanpro:


MarkLark86 (commented):
As `language` is a generic field it would be beneficial if it was populated in the `events` service, if not already defined, and not be specific to any particular ingest service

server/planning/events/events.py
```diff
@@ -119,6 +119,10 @@ def _enhance_event_item(self, doc):
         if not doc.get("original_creator"):
             doc.pop("original_creator", None)
 
+        # SDCP-638
+        if not doc.get("language"):
```
MarkLark86, line 123: Almost right. The `on_fetched` and `on_fetched_item` functions modify the item from a `GET` request, which these changes are not stored in the DB.

We want to modify the item on initial creation, which would be the `on_create` function

petrjasek:
we should probably do the same for planning items to be consistent

server/features/search_events.feature
```diff
@@ -174,10 +174,12 @@ Feature: Event Search
         ]}
         """
         When we get "/events_planning_search?repo=events&only_future=false&language=en"
-        Then we get list with 1 items
+        Then we get list with 3 items
```
MarkLark86, line 177: This change doesn't really test the `language=en` filter now that the system default is applied on creation.
It might be good to change this Event (at the top of this file) and this URL to use a language different to the default (such as `fr-CA`) instead

server/features/search_planning.feature
```diff
@@ -260,10 +260,15 @@ Feature: Planning Search
         ]}
         """
         When we get "/events_planning_search?repo=planning&only_future=false&language=en"
-        Then we get list with 1 items
+        Then we get list with 6 items
```
MarkLark86, line 263: Same with this one, changing the Planning item's language to a different value than the system default and changing this URL to reflect.

MarkLark86 (approved):

PR #1720, opened 2022-09-09
SDAAP-36 Stop assignment crashing when there is no user

marwoodandrew:


client/components/Assignments/AssignmentHistory.tsx
```diff
@@ -108,7 +112,9 @@ class AssignmentHistoryComponent extends React.Component {
     }
 
     render() {
-        const getHistoryActionUserName = (userId) => getItemInArrayById(this.props.users, userId).display_name;
+        const getHistoryActionUserName = (userId) =>
```
MarkLark86, line 115: You could use Typescript's `Optional Chaining` and `Nullish Coalescing` operators so that `getItemInArrayById` isn't called twice.
i.e.
```typescript
getItemInArrayById(this.props.users, userId)?.display_name ?? ''
```

marwoodandrew: Very nice, will give it a go!

MarkLark86 (approved):

PR #1724, opened 2022-09-29
[STTNHUB-179] ingest(eventsml): Fix date parsing and add new Event fields

MarkLark86:
Depends on https://github.com/superdesk/superdesk-core/pull/2383

### DIFF
--- a/server/planning/feed_parsers/events_ml_test.py
+++ b/server/planning/feed_parsers/events_ml_test.py
@@ -1,7 +1,9 @@
+from typing import Optional
 from os import path
-from xml.etree import ElementTree
 from datetime import datetime
 from dateutil.tz import tzoffset
+from pytz import utc
+from superdesk.etree import etree
 
 from superdesk.metadata.item import (
     ITEM_TYPE,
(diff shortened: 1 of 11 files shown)


server/planning/feed_parsers/events_ml_test.py
```diff
@@ -104,5 +110,130 @@ def test_content(self):
...
+                tz=self.app.config["DEFAULT_TIMEZONE"],
+            ),
+        )
+
+        # Only start date & time supplied, with time NOT midnight in local time
+        self.assertEqual(
+            get_item_dates("2022-07-05T18:00:00+03:00"),
+            dict(
+                start=datetime(2022, 7, 5, 15, 0, tzinfo=utc),
+                end=datetime(2022, 7, 5, 16, 0, tzinfo=utc),
```
petrjasek, line 162: I'm not sure if we should interpret this as 1h event, imo would be better to have explicit end=None, once ingested we would loose that information

petrjasek (approved):

server/planning/feed_parsers/events_ml_test.py
MarkLark86: We have a config in the system for the default Event duration (the system expects start & end dates for all Events).

https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/feed_parsers/events_ml.py#L133
This same config is used in the front end as well as the back end.

petrjasek: yep but that doesn't really reflect the reality if it's unknown, or do we show in the UI it's using the default and we don't know the actual length?

MarkLark86: We're always storing an end date

MarkLark86: It's untested if having end=None will be supported at this time

petrjasek: ok np, if that's how it works now we should use that

PR #1725, opened 2022-09-30
Fix : Event links are overlapping in the preview [SDESK-6558]

devketanpro:


MarkLark86 (changes requested):
Rather than adding more custom styles (we try to avoid them if possible), we can set `noLabel={false}` to the `<LineInput`
https://github.com/superdesk/superdesk-planning/blob/develop/client/components/UI/Form/LinkInput.tsx#L165

This would solve the problem without any extra custom styles

MarkLark86 (approved):

PR #1732, opened 2022-11-14
[SDESK-6698] fix spike event date not showing in the UI

thecalcc:


client/utils/events.ts
```diff
@@ -419,9 +419,9 @@ const getDateStringForEvent = (event, dateOnly = false, useLocal = true, withTim
     // !! Note - expects event dates as instance of moment() !! //
     const dateFormat = appConfig.planning.dateformat;
     const timeFormat = appConfig.planning.timeformat;
-    const start = get(event.dates, 'start');
-    const end = get(event.dates, 'end');
-    const tz = get(event.dates, 'tz');
+    const start = get(event, 'start');
```
MarkLark86, line 422: This function is used in multiple places, passing in the entire Event item.
In this case it might be good to change those usages of `getDateStringForEvent` as well.

client/utils/events.ts
```diff
@@ -419,9 +419,9 @@ const getDateStringForEvent = (event, dateOnly = false, useLocal = true, withTim
     // !! Note - expects event dates as instance of moment() !! //
```
MarkLark86, line 419: It might also help to add types to the signature, such as `eventSchedule: IEventItem['dates']` so it is clear what this function is expecting

MarkLark86:
Looks like the usage of `EventScheduleSummary` and `getDateStringForEvent` is not consistent (some pass in the entire Event, some pass in the Event schedule). The interface/usage between these all will have to be fixed up.

image
image

MarkLark86:
@thecalcc There are also invalid arguments pass to `getDateStringForEvent` in many places, some passing the whole Event while others passing only the Event Schedule. These will need to be fixed up as well

client/components/Events/EventScheduleSummary/index.tsx
```diff
@@ -6,17 +6,30 @@ import {RepeatEventSummary} from '../RepeatEventSummary';
...
-    if (!schedule)
+interface IProps {
+    schedule: Partial<IEventItem>,
+    noPadding?: boolean,
+    forUpdating?: boolean,
+    useEventTimezone?: boolean
+}
+
+export const EventScheduleSummary = ({
+    schedule: event,
```
MarkLark86, line 20: If we're now passing in the entire Event, it would make more sense to rename this prop to `event` or `item` instead (as passing the entire Event into a prop named `schedule` can be misleading)

MarkLark86 (approved):

PR #1733, opened 2022-11-15
Planning: As a SD admin I want to have a flag 'no end time' stored for ingested events that arrive without an end time set [SDESK-6646]

devketanpro:


client/components/Events/EventDateTime.tsx
```diff
@@ -72,13 +75,16 @@ export class EventDateTime extends React.PureComponent<IProps> {
                     date={start}
                     {...commonProps}
                 />
+                {!noEndTime && (
```
petrjasek, line 78: the date part might be still relevant from the end timestamp, would it still display it?

petrjasek:
can you also handle the `dates.all_day` flag?

client/components/Events/EventDateTime.tsx
devketanpro: Yes, end date is also relevant to end time stamp. It would not display when no_end_time flag is true.
image
image (1)

petrjasek: the end date should be visible for multiday events even if `no_end_time` is set

client/components/Events/EventScheduleSummary/index.tsx
```diff
@@ -49,11 +59,35 @@ export const EventScheduleSummary = ({schedule, noPadding, forUpdating, useEvent
         currentDateLabel = gettext('Current Date (Based on Event timezone)');
     }
 
+    const start = moment(eventSchedule.start);
+    const end = moment(eventSchedule.end);
```
petrjasek, line 63: in case there is no end time or for start date when it's all day event it needs to use `moment.utc()` instead of `moment()` to interpret the dates as it is and not trying to convert those to local timezone.
same like here https://github.com/superdesk/newsroom-core/pull/231

client/components/UI/DateTime.tsx
```diff
@@ -12,37 +12,82 @@ import './style.scss';
...
+        eventEndDate = null;
+        setHideDash && setHideDash(true);
+    } else if (allDay && multiDay) {
+        eventEndDate = moment(date).format(dateTimeFormat)
+            .slice(0, 5) || '';
+    } else if (noEndTime && multiDay) {
+        eventEndDate = moment(date).format(dateTimeFormat)
+            .slice(0, 5) || '';
+    } else if (allDay && !multiDay) {
+        setHideDash && setHideDash(true);
```
petrjasek, line 56: I don't see the need for this setHideDash call, doesn't the parent component have all the info?

client/components/UI/DateTime.tsx
```diff
@@ -12,37 +12,82 @@ import './style.scss';
...
+    } else {
+        eventEndDate = moment(date).format(dateTimeFormat);
+    }
+
+    let eventStartDate;
+
+    if (allDay && multiDay) {
+        eventStartDate = moment(date).format(dateTimeFormat)
```
petrjasek, line 64: when ignoring the time there should be always `moment.utc` otherwise this will be converted to local timezone

client/components/UI/DateTime.tsx
devketanpro: We have to use setHideDash because parent component doesn't have that info on which we have to hide dash. We are hiding dash when we don't want to show end date and the whole logic resides in child component. That's why i thought it would be better to use this here.

client/components/Events/EventDateTime.tsx
```diff
@@ -50,9 +62,15 @@ export class EventDateTime extends React.PureComponent<IProps> {
             );
         }
 
+        const noEndTime = item.dates?.no_end_time;
+        const allDay = item.dates?.all_day;
```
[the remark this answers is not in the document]
devketanpro: Yes, I thought the same but isAllDay flag is not same as allDay. isAllDay is build according to previous needs when we didn't have no_end_time falg.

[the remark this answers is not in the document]
petrjasek: yep I know, but we should harmonize that probably because both should have the same effect, if not we should change the names because it's just confusing

client/components/UI/DateTime.tsx
petrjasek: given that the child component gets all the info from the parent it makes me think that the parent has all the info

devketanpro: For the start and finish dates, we do, however, use a child component twice in the parent component, with a dash in between. Date logic is included in the child component, and since the end date is hidden from the logic of the child component, we must use it.

client/utils/events.ts
```diff
@@ -445,6 +446,38 @@ const getDateStringForEvent = (event, dateOnly = false, useLocal = true, withTim
...
+    const isFullDay = event?.dates?.all_day;
+    const noEndTime = event?.dates?.no_end_time;
+
+    const multiDay = !isEventSameDay(start, end);
+
+    if (isFullDay && !multiDay) {
+        return datesToShow = start.format(dateFormat);
+    } else if (noEndTime && !multiDay) {
+        if (withTimezone) {
+            if (!useLocal && (tz || tz == null)) {
```
petrjasek, line 458: not sure about that `tz` condition, it can be set or null, so what are we actually checking there?

client/utils/events.ts
```diff
@@ -445,6 +446,38 @@ const getDateStringForEvent = (event, dateOnly = false, useLocal = true, withTim
...
+                // eslint-disable-next-line max-len
+                timezoneForEvents = `(${getDateTimeString(start, dateFormat, timeFormat, ' @ ', true, tz ? tz : 'utc')})`;
+            } else {
+                timezoneForEvents = getDateTimeString(start, dateFormat, timeFormat, ' @ ', true);
+            }
+        }
+        return datesToShow = timezoneForEvents;
```
petrjasek, line 465: not sure in general why there is `return` with `=` a few times, wouldn't just return do?

client/components/UI/DateTime.tsx
```diff
@@ -12,37 +12,81 @@ import './style.scss';
...
+        .join('\u00a0'); // &nbsp;
+
+    let eventStartDate;
+
+    if (isFullDay && multiDay) {
+        eventStartDate = moment.utc(date).format(dateFormat);
+    } else if (noEndTime && !multiDay || multiDay) {
+        eventStartDate = moment(date).format(dateFormat) + ' ' + moment(date).format(timeFormat);
+    } else if (isFullDay && !multiDay) {
+        eventStartDate = moment.utc(date).format(dateFormat);
```
petrjasek, line 50: seems like this is the same for `isFullDay == true` no matter what multiday is

client/components/UI/DateTime.tsx
```diff
@@ -12,37 +12,81 @@ import './style.scss';
...
+    ]
+        .filter((d) => d)
+        .join('\u00a0'); // &nbsp;
+
+    let eventStartDate;
+
+    if (isFullDay && multiDay) {
+        eventStartDate = moment.utc(date).format(dateFormat);
+    } else if (noEndTime && !multiDay || multiDay) {
+        eventStartDate = moment(date).format(dateFormat) + ' ' + moment(date).format(timeFormat);
```
petrjasek, line 48: why not `moment(date).format(dateTimeFormat)` ?

client/components/UI/DateTime.tsx
```diff
@@ -12,37 +12,81 @@ import './style.scss';
...
     return (
-        <time
-            className={!padLeft ? 'Datetime' : null}
-            title={date.toString()}
-        >
-            {moment(date).format(dateTimeFormat)}
+        <time className={!padLeft ? 'Datetime' : null} title={date.toString()}>
+            {!isEndEventDateTime && eventStartDate}
+            {isEndEventDateTime &&
+        (eventEndDate || eventEndDate == null) &&
```
petrjasek, line 73: when the `eventEndDate` is `null` should this be still there?

client/utils/events.ts
devketanpro: In (noEndTime && !multiDay) scenario when there is utc `tz `we are getting null so for that we have to add this extra condition. For other time zones it is working properly.

client/utils/events.ts
devketanpro: my bad , I forget to remove it.

client/components/UI/DateTime.tsx
devketanpro: in some scenarios we are getting only time 09:00 instead of 27.11 09:00 date and time. So, for getting desired result we have to use it

client/utils/events.ts
petrjasek: my point is that `(tz || tz == null)` should be pretty much always true, or is there a case when it's not?

client/components/UI/DateTime.tsx
petrjasek: what scenarios?

client/utils/events.ts
devketanpro: It is `true` when `withTimezone` is `true`.If flag `withTimezone` is true we should have `tz `but in `utc` scenario we are getting null that's why we are assigning `utc` only if `withTimezone` is true and then `tz` is `null`

petrjasek: I see, but then this condition `(tz || tz == null)` is not needed, is it?

client/components/UI/DateTime.tsx
devketanpro: When `noEndTime` is `true` or `multiday` is `true` or `false` sometimes we are getting only time through `dateTimeFormat` so to make sure we are getting desired output we are using this

client/utils/events.ts
devketanpro: IMO it is needed when there is `UTC` timezone we have to show it in preview but we are getting null. please let me know if I am missing something.

client/components/UI/DateTime.tsx
petrjasek: can you add an example? I still don't get it

devketanpro: image (2)
image (3)
So, here is an examples. I consoled 
`moment(date).format(dateTimeFormat)`
Case 1:-  `allDay = true` and `multiDay = false`, as you can see even if we are using `dateTime` format we are only getting time "05 : 30".
Case 2:- `noEndTime= true` and `multiDay = false`, we are getting only time not date and time. like:- 18:29, 11:00, 23:59.
We are only getting date and time with `moment(date).format(dateTimeFormat)` when `multiDay` is true.

petrjasek: still not sure I get this, my question was why are we using
```
eventStartDate = moment(date).format(dateFormat) + ' ' + moment(date).format(timeFormat);
```
instead of
```
eventStartDate = moment(date).format(dateTimeFormat);
```
?

client/components/Events/EventDateTime.tsx
```diff
@@ -72,15 +88,17 @@ export class EventDateTime extends React.PureComponent<IProps> {
                     date={start}
                     {...commonProps}
                 />
-                &ndash;
+                {hideDash && <>&ndash;</>}
```
petrjasek, line 91: just wondering if it's called `hideDash` shouldn't the dash be rather removed? maybe this should be rather called `showDash`

client/utils/events.ts
petrjasek: I think that:
```
if (!useLocal) {
```
would do because that `(tz || tz == null)` is always true in
```
if (!useLocal && (tz || tz == null)) {
```

client/components/UI/DateTime.tsx
```diff
@@ -12,37 +12,78 @@ import './style.scss';
...
         withDate ? format : null,
         withTime ? newTimeFormat : null,
-    ].filter((d) => d).join('\u00a0'); // &nbsp;
+    ]
+        .filter((d) => d)
+        .join('\u00a0'); // &nbsp;
+
+    let eventStartDate;
+
+    if (isFullDay && (multiDay || !multiDay)) {
```
petrjasek, line 45: this `(multiDay || !multiDay)` looks pretty useless

client/components/UI/DateTime.tsx
```diff
@@ -12,37 +12,78 @@ import './style.scss';
...
+        eventStartDate = moment.utc(date).format(dateFormat);
+    } else if (noEndTime && (!multiDay || multiDay)) {
```
petrjasek, line 47: same here

petrjasek (approved):

PR #1735, opened 2022-11-17
Fix Menu component rendering after UI framework changes

thecalcc:
Wrapped all Menu instances with `<div>` because `<div>` was removed from Menu in UI framework


MarkLark86:
@thecalcc I think these changes are meant for the `beta/2.6` branch and not `develop`

PR #1743, opened 2022-12-05
Planning: As a SD admin I want to have a flag 'no end time' stored for ingested events that arrive without an end time set [SDESK-6646]

devketanpro:


client/components/UI/DateTime.tsx
```diff
@@ -56,18 +56,18 @@ function DateTime({
...
-    const isEventEndDate = isEndEventDateTime && eventEndDate;
+    const startDate = !isEndEventDateTime && eventStartDate;
+    const endDate = isEndEventDateTime && eventEndDate;
     const tz = timeUtils.getTimeZoneAbbreviation(date.format('z')) + ' ';
 
     return (
-        <time className={!padLeft ? 'Datetime' : null} title={tz + (isEventStartDate || isEventEndDate)}>
-            {isEventStartDate}{isEventEndDate}
+        <time className={!padLeft ? 'Datetime' : null} title={tz + (startDate || endDate)}>
+            {startDate}{endDate}
```
petrjasek, line 70: would we ever display both here? if not it could be rather like:
```
const displayDate = isEndEventDateTime ? eventEndDate : eventStartDate;
```

petrjasek (approved):

PR #1745, opened 2022-12-12
[SDAAP-51] Populating the completed assignments list on the assignmen…

marwoodandrew:
…ts page is slow

### DIFF
--- a/client/actions/assignments/api.ts
+++ b/client/actions/assignments/api.ts
@@ -162,6 +162,7 @@ const query = ({
     dateFilter = null,
     size = null,
     ignoreScheduledUpdates = false,
+    max_results = null,
 }) => (
     (dispatch, getState, {api}) => {
         const filterByValues = {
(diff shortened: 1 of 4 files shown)


marwoodandrew:
@MarkLark86 Not sure I'm in love with the front end part of this, It may not be required and the infinite scrolling on the assignments page is crap, It clears an redraws the whole pane when updating with the next batch. I think the server bit might be enough.

MarkLark86:
> @MarkLark86 Not sure I'm in love with the front end part of this, It may not be required and the infinite scrolling on the assignments page is crap, It clears an redraws the whole pane when updating with the next batch. I think the server bit might be enough.

@marwoodandrew I think this should be okay for your use case, as 100 items should cover most scenarios and there is also the search bar if users are looking for a specific assignment.

Nonetheless we will need to fix the infinite scroll in another ticket.

MarkLark86 (approved):

marwoodandrew:
Closed for rebase

PR #1756, opened 2023-01-16
Fix : Planning: Language in event templates is not kept [SDESK-6792]

devketanpro:


server/planning/events/events_template.py
```diff
@@ -39,6 +39,7 @@ class EventsTemplateResource(Resource):
     }
     _event_fields = {
         "slugline": {"type": "string", "required": False, "readonly": True},
+        "language": {"type": "string", "required": False, "readonly": True},
```
MarkLark86, line 42: It might be easier to maintain EventTemplates if we don't restrict the resource schema to specific fields, but rather allow any fields to be stored under the `data` attribute.

server/planning/events/events_template.py
```diff
@@ -179,6 +180,7 @@ def _fill_event_template(self, doc):
             "event_contact_info",
             "subject",
             "occur_status",
+            "language",
```
MarkLark86, line 183: Similar here, it would be better IMO to not specifically provide what fields to copy, but rather what fields to exclude. We have something in place already for Event Duplicate action:
https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/events/events_duplicate.py#L63

Might be good to move the declaration for the list of fields to remove into a common file as well.

server/planning/events/events_template.py
```diff
@@ -94,7 +95,7 @@ class EventsTemplateResource(Resource):
             embeddable=False,
             required=True,
         ),
-        "data": {"type": "dict", "schema": _event_fields},
+        "data": {"type": "dict", "schema": _event_fields, "allow_unknown": True},
```
MarkLark86, line 98: `schema` can be an empty dict, and we can delete the `_event_field` attribute on the class, otherwise the underlying framework (Eve) will validate the `data` field against this schema (which does not add any value here)

server/planning/common.py
```diff
@@ -135,6 +135,27 @@ class PostStates(NamedTuple):
 TO_BE_CONFIRMED_FIELD = "_time_to_be_confirmed"
 TO_BE_CONFIRMED_FIELD_SCHEMA = {"type": "boolean"}
 
+EVENT_IGNORED_FIELDS = {
```
MarkLark86, line 138: The name `EVENT_IGNORED_FIELDS` does not provide context as to when these fields are ignored.
It would provide more meaning if the name indicates that these are the fields that are ignored when duplicating an Event

MarkLark86 (approved):

PR #1757, opened 2023-01-30
Fix : closing without changes introduced brings Save changes modal[SDESK-6691]

devketanpro:


client/reducers/featuredPlanning.ts
```diff
@@ -73,7 +73,7 @@ function updateListIds(prevState: IFeaturedPlanningState): IFeaturedPlanningStat
 
     const dirty = (removePlanningIds?.length ?? 0) > 0 || (
         !existingItem &&
-        selectedPlanningIds?.length &&
+        !selectedPlanningIds &&
```
MarkLark86, line 76: This doesn't seem like the required fix. The modal/form will be dirty if `selectedPlanningIds` is not empty.
Does the form become dirty when moving Planning items around?

devketanpro: Yeah, when `selectedPlanningIds` is not null, it becomes dirty. however, saving that item without any changes does not operate properly.

MarkLark86: If you add any item to a `featured story` then visit the FeaturedStories modal, there are unsaved changes because that newly added item has not been saved into said `featured story` list of items (only been flagged).

MarkLark86: We eve have E2E tests to make sure this happens. Not sure if this is even a bug really but a intended feature: https://github.com/superdesk/superdesk-planning/blob/develop/e2e/cypress/e2e/planning/featured_planning.cy.ts#L84

PR #1760, opened 2023-02-15
As a Planning editor I want to filter the events by language[SDESK-6741]

devketanpro:


server/planning/content_profiles/profiles/advanced_search.py
```diff
@@ -157,6 +150,13 @@
                 "search_enabled": True,
                 "filter_enabled": True,
             },
+            "language": {
+                "enabled": True,
+                "index": 1,
+                "group": "language",
```
MarkLark86, line 156: I don't think we need a specific group for 1 input, does this work if changing `enabled` to `True` (with no other changes)?

devketanpro: >  does this work if changing enabled to True (with no other changes)?
 yes its works

devketanpro: > I don't think we need a specific group for 1 input,

IMO , It is mentioned in the ticket to create a new language group.

MarkLark86 (approved):

MarkLark86 (approved):

PR #1763, opened 2023-02-23
Fix : Event creation and update notifications are not appearing [SDESK-6024]

devketanpro:


client/actions/main.ts
```diff
@@ -236,7 +236,8 @@ const save = (original, updates, withConfirmation = true) => (
 
         switch (itemType) {
         case ITEM_TYPE.EVENT:
-            promise = dispatch(eventsUi.save(original, updates, confirmation));
+            confirmation = false;
```
MarkLark86, line 239: Confirmation **may** be required when saving an Event (that's where `withConfirmation` comes into play).

client/actions/main.ts
```diff
@@ -237,6 +237,7 @@ const save = (original, updates, withConfirmation = true) => (
         switch (itemType) {
         case ITEM_TYPE.EVENT:
             promise = dispatch(eventsUi.save(original, updates, confirmation));
+            confirmation = false;
```
MarkLark86, line 240: Looking further at the code of this function, this solution won't work either.
If the user clicks `cancel` in the Ignore/Cancel/Save modal.
See the comment here: https://github.com/superdesk/superdesk-planning/blob/1d07aa9d09927e29c726741808a56fd3c9ccef2c/client/actions/main.ts#L258

MarkLark86 (approved):

PR #1766, opened 2023-02-28
add buffer to onclusive api call when fetching updates

petrjasek:
also change the api error handling and retry on 400 which happens sometimes

CPNHUB-198


petrjasek:
also changed the way of checking for new items so it's not reingesting autoposted events

MarkLark86 (approved):

PR #1767, opened 2023-03-02
ignore embargoed events from onclusive

petrjasek:
CPNHUB-212

### DIFF
--- a/server/planning/feed_parsers/onclusive.py
+++ b/server/planning/feed_parsers/onclusive.py
@@ -18,6 +18,10 @@ from flask import current_app as app
 logger = logging.getLogger(__name__)
 
 
+class EmbargoedException(RuntimeError):
+    pass
+
+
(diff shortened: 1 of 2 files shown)


MarkLark86 (approved):
@petrjasek I'm assuming for now we simply won't ingest these at all, but my question would be at what stage would these Event's make it to Newshub?

If an Event is embargoed, then these Event's will never make it to Newshub.

petrjasek:
yep we will handle it eventually, atm those events won't make it at all (that's confirmed by client, it's pretty rare probably).
the tricky part is when we should be sending notifications for those events.

MarkLark86:
> yep we will handle it eventually, atm those events won't make it at all (that's confirmed by client, it's pretty rare probably). the tricky part is when we should be sending notifications for those events.

Understood, all good then.

PR #1768, opened 2023-03-03
[SDESK-6737] feature: Support multilingual text fields on Events & Planning items

MarkLark86:


server/planning/events/events_schema.py
```diff
@@ -311,4 +315,15 @@
...
     },
     "extra": metadata_schema["extra"],
+    "translations": {
+        "type": "list",
+        "mapping": {
+            "type": "nested",
+            "properties": {
+                "field": not_analyzed,
+                "language": not_analyzed,
+                "value": not_analyzed,
```
petrjasek, line 325: I think the value in general should be analyzed if it's for data like headline/description

petrjasek (approved):

devketanpro (approved):

PR #1771, opened 2023-03-14
use timezone info when checking embargoed time from onclusive

petrjasek:
CPNHUB-212

### DIFF
--- a/server/planning/feed_parsers/onclusive.py
+++ b/server/planning/feed_parsers/onclusive.py
@@ -12,6 +12,7 @@ from superdesk.metadata.item import (
     CONTENT_STATE,
 )
 from superdesk.errors import ParserError
+from superdesk.utc import utcnow
 from planning.common import POST_STATE
 from flask import current_app as app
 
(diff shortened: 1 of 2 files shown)


MarkLark86 (approved):

MarkLark86:
@petrjasek We're mandating here that `embargoTime` and `timezone.timezoneOffset` are both required, is there ever a case when Onclusive will provide only `embargoTime` without a UTC offset?

MarkLark86:
Merging from `develop` should fix the tests

petrjasek:
> @petrjasek We're mandating here that `embargoTime` and `timezone.timezoneOffset` are both required, is there ever a case when Onclusive will provide only `embargoTime` without a UTC offset?

yep that's what they confirmed

> Merging from `develop` should fix the tests

is there anything that shouldn't make it to the 2.6.1 release?

MarkLark86:
> > @petrjasek We're mandating here that `embargoTime` and `timezone.timezoneOffset` are both required, is there ever a case when Onclusive will provide only `embargoTime` without a UTC offset?
> 
> yep that's what they confirmed
> 
> > Merging from `develop` should fix the tests
> 
> is there anything that shouldn't make it to the 2.6.1 release?

All good then. Sorry, I was incorrect, rebase against the hotfix/2.6.1 branch and will fix failing test

PR #1775, opened 2023-03-21
Fix : When multilingual is not checked the language cancellation leads to a blank page [SDESK-6856]

devketanpro:


client/components/fields/editor/Language.tsx
```diff
@@ -64,7 +64,7 @@ export class EditorFieldLanguageComponent extends React.PureComponent<IProps> {
             this.props.setMainLanguage != null
         );
         const selectedLanguages = this.props.languages
-            .filter((entry) => values.includes(entry.value.qcode))
+            .filter((entry) => values?.includes(entry.value.qcode))
```
MarkLark86, line 67: I'm curious how an exception can be raised here and not [here](https://github.com/superdesk/superdesk-planning/blob/8346cbf6842f2850757fa11d14c96a76f2389132/client/components/fields/editor/Language.tsx#L63) 
`values.length > 1 &&`

```typescript
const showMainLanguageChoice = (
    isMultilingual === true &&
    values.length > 1 &&
    this.props.setMainLanguage != null
);
```

devketanpro: because here we got `isMultilingual === false.`.

MarkLark86 (approved):

PR #1781, opened 2023-04-01
As a Planning editor I want to see all items with Coverages assigned to a specific editor [SDESK-6745].

devketanpro:


client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -51,11 +57,16 @@ class PlanningListSubNavComponent extends React.Component<IProps, IState> {
     sortFieldOptions: Array<{label: string, onSelect(): void}>
     container?: HTMLDivElement;
     resizeObserver: ResizeObserver;
+    userList: Array<{label: string, onSelect(): void}>
 
     constructor(props) {
         super(props);
+        this.state = {
```
MarkLark86, line 64: Rather than storing this locally, it would be better to pass in this search param from a higher up component, like we're already doing with the other search params in this component

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -51,11 +57,16 @@ class PlanningListSubNavComponent extends React.Component<IProps, IState> {
...
+            name: ''
+        };
         const {gettext} = superdeskApi.localization;
 
+        this.props.users.unshift({_id: null, display_name: 'All'});
```
MarkLark86, line 69: Props should **never** be modified, no matter what the circumstances.
Instead you create a new array in the render method, combining this `All` choice and the list of users

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -171,6 +197,19 @@ class PlanningListSubNavComponent extends React.Component<IProps, IState> {
                     <ButtonGroup align="inline">
                         <FilterSubnavDropdown viewSize={this.state.viewSize} />
                     </ButtonGroup>
+                    {this.props.activefilter === 'EVENTS' ? ' ' : (
+                        <div>
+                          All items and Coverages assigned to :
```
MarkLark86, line 202: `gettext` is missing.
This label is also very long, and would not work on smaller screens.

client/utils/planning.ts
```diff
@@ -974,7 +974,8 @@ function getCoverageDateTimeText(coverage: IPlanningCoverageItem) {
  */
 function getCoverageIcon(
     type: IG2ContentType['qcode'] | IG2ContentType['content item type'],
-    coverage?: DeepPartial<IPlanningCoverageItem>
+    coverage?: DeepPartial<IPlanningCoverageItem>,
+    users?: Array<IUser>
```
MarkLark86, line 978: `users` argument is not used, needs to be removed

client/apps/Planning/PlanningListSubNav.tsx
devketanpro: my bad I forgot to remove this.

client/apps/Planning/PlanningListSubNav.tsx
MarkLark86: `state` is still being used. It is better to pass in the search param instead of storing it locally

client/api/combined.ts
```diff
@@ -24,7 +24,9 @@ function convertCombinedParams(params: ISearchParams): Partial<ISearchAPIParams>
         calendars: cvsToString(params.calendars),
         agendas: arrayToString(params.agendas),
         include_associated_planning: params.include_associated_planning,
-        source: cvsToString(params.source, 'id')
+        source: cvsToString(params.source, 'id'),
+        coverage_user_id: params.coverage_user_id,
+        coverage_user_name: params.coverage_user_name,
```
MarkLark86, line 29: There is no need to pass `coverage_user_name` as a search param, as it doesn't do anything.

Instead you would need to pass in the current `coverage_user_id` param into the `PlanningListSubNavComponent` using the `mapStateToProps`. Then in this component, you would get the name of the User based on the above prop and the `users` prop.

See how this is currently done with other params in the same component:
* https://github.com/superdesk/superdesk-planning/blob/develop/client/apps/Planning/PlanningListSubNav.tsx#L39
* https://github.com/superdesk/superdesk-planning/blob/develop/client/selectors/main.ts#L89

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -22,14 +22,17 @@ interface IProps {
     currentInterval: 'DAY' | 'WEEK' | 'MONTH';
     sortOrder: SORT_ORDER;
     sortField: SORT_FIELD;
-
+    userList:Array<{label: string, onSelect(): void}>;
```
MarkLark86, line 25: This prop is never used, and can be removed (along with it's assignment inside the `mapStateToProps` function

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -22,14 +22,17 @@ interface IProps {
...
+    users: any;
```
MarkLark86, line 26: Avoid using `any` with types, it is more meaningful and provides better type checking if you use specific types.

Here it would be: `users: Array<IUser>`

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -38,6 +41,10 @@ const mapStateToProps = (state) => ({
     listViewType: selectors.main.getCurrentListViewType(state),
     sortOrder: selectors.main.getCurrentSortOrder(state),
     sortField: selectors.main.getCurrentSortField(state),
+    users: selectors.general.users(state),
+    groups: selectors.main.itemGroups(state),
+    activefilter: selectors.main.activeFilter(state),
```
MarkLark86, line 46: `activefilter` and `currentSearch` is not required. Instead create a new selector for the search param you require (in this case, it would be a selector to get `coverage_user_id`)

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -131,7 +139,19 @@ class PlanningListSubNavComponent extends React.Component<IProps, IState> {
         planningApi.ui.list.search({sort_field: field});
     }
 
+    filterCoverageUser(item) {
+        planningApi.ui.list.search({coverage_user_id: item._id, coverage_user_name: item.display_name});
+    }
+
     render() {
+        let newOption = {_id: null, display_name: 'ALL'};
+        let List = [newOption, ...this.props.users];
```
MarkLark86, line 148: The naming convention we use in Typescript is CamelCase with the first character to be lower case

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -22,14 +22,17 @@ interface IProps {
     currentInterval: 'DAY' | 'WEEK' | 'MONTH';
     sortOrder: SORT_ORDER;
     sortField: SORT_FIELD;
-
+    userList:Array<{label: string, onSelect(): void}>;
+    users: any;
     setStartFilter(value?: moment.Moment): void;
     jumpTo(interval: 'TODAY' | 'BACK' | 'FORWARD'): void;
     setJumpInterval(interval: 'DAY' | 'WEEK' | 'MONTH'): void;
+    activefilter?:string;
```
MarkLark86, line 30: We have an enum `FILTER_TYPE` in the `interfaces.ts` file, which can be used here. This would provide better type checking when comparing this value

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -22,14 +22,17 @@ interface IProps {
...
+    currentSearch : any;
 }
 
 interface IState {
-    viewSize: SUBNAV_VIEW_SIZE;
+    viewSize?: SUBNAV_VIEW_SIZE;
```
MarkLark86, line 35: Why was this one changed to optional?

client/components/Coverages/CoverageIcon.tsx
```diff
@@ -85,7 +85,7 @@ export class CoverageIcon extends React.PureComponent<IProps> {
                         planningUtils.getCoverageIcon(
                             planningUtils.getCoverageContentType(this.props.coverage, this.props.contentTypes) ||
                             this.props.coverage.planning?.g2_content_type,
-                            this.props.coverage
+                            this.props.coverage,
```
MarkLark86, line 88: This change can be removed, as it does nothing to the code and there are no other changes in this file

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -22,10 +23,12 @@ interface IProps {
     currentInterval: 'DAY' | 'WEEK' | 'MONTH';
     sortOrder: SORT_ORDER;
     sortField: SORT_FIELD;
-
+    users: Array<IUser>
     setStartFilter(value?: moment.Moment): void;
     jumpTo(interval: 'TODAY' | 'BACK' | 'FORWARD'): void;
     setJumpInterval(interval: 'DAY' | 'WEEK' | 'MONTH'): void;
+    activefilter: PLANNING_VIEW;
+    coverageUser : string;
```
MarkLark86, line 31: As this param may be `undefined`, it should be optional in the props.
It would also be more meaningful to use `IUser['_id']` here so it describes what type of data it is and the context

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -171,6 +189,20 @@ class PlanningListSubNavComponent extends React.Component<IProps, IState> {
                     <ButtonGroup align="inline">
                         <FilterSubnavDropdown viewSize={this.state.viewSize} />
                     </ButtonGroup>
+                    {this.props.activefilter == PLANNING_VIEW.EVENTS ? ' ' : (
+                        <div>
+                            {gettext('Assigned Coverages Items :')}
+                            <Dropdown items={userList}>
+                                <span className="sd-margin-l--1 sd-margin-r--3">
+                                    {this.props.users.find(
+                                        (user) => user._id == this.props.coverageUser)?.display_name ?? 'ALL'}
```
MarkLark86, line 198: `gettext` is missing on `ALL`

client/selectors/main.ts
```diff
@@ -96,6 +96,10 @@ export const getCurrentSortField = createSelector(
     (params) => params.sortField ?? SORT_FIELD.SCHEDULE,
 );
 
+export const getCoverageUser = createSelector(
+    [currentSearch],
+    (params) => params.coverage_user_id ?? 'ALL',
```
MarkLark86, line 101: No need to return `ALL`, `undefined` will suffice (indicating that this param has no value)

MarkLark86 (approved):

PR #1788, opened 2023-04-21
Fix : When using multilingual option for event fields, all required language fields should be validated [SDESK-6869]

devketanpro:


client/validators/profile.ts
```diff
@@ -38,6 +40,14 @@ export const formProfile = ({field, value, profile, errors, messages}) => {
     } else if (schema.required && (typeof fieldValue === 'number' ? !fieldValue : isEmpty(fieldValue))) {
         errors[field] = gettext('This field is required');
         messages.push(gettext('{{ name }} is a required field', {name: fieldLabel}));
+    } else if (schema.required && schema.multilingual) {
+        const errorMessage = gettext(`${fieldLabel} is a required field`);
+
+        if (field !== 'language' && diff?.languages?.length !== field_value?.length || field_value.some((obj
+        ) => obj.value === '')) {
+            errors[field] = gettext('This field is required');
```
MarkLark86, line 48: Do these errors appear under the field that is required? i.e. if the Canadian French version of `name` is not filled, is the error displayed underneath this field (might need to store the error under the `field` and `language` from the `translations` attribute entry

client/validators/profile.ts
```diff
@@ -11,6 +11,8 @@ export const formProfile = ({field, value, profile, errors, messages}) => {
 
     const fieldValue = (typeof value === 'string') ? value.trim() : value;
 
+    const field_value = (diff?.translations || []).filter((e) => e.field === field);
```
MarkLark86, line 14: By convention we use camel case for variable names in Typescript/Javascript.
Also, as this value is not used anywhere except inside the `if ` statement for multilinugal, might be better to place it there (otherwise it's filtering translations when it may not need to be used).

MarkLark86:
Play around with those errors, as they still aren't showing up in the form (tried this locally).
The following should be visible upon errors:
image

See how the field in the form is highlighted red, with a message underneath stating that this field is required.
It would be good to have the same for multilingual fields too.

I believe this is done using the `errors` variable, and the key should be `<field>.<language>` for multilingual errors.
i.e.
`errors['name.en'] = gettext('This field is required')`
See the following attributes passed (shown via React dev tools)
image

MarkLark86 (approved):

PR #1789, opened 2023-04-24
[SDESK-6851] Refactor locking mechanism

MarkLark86:


client/actions/assignments/api.ts
```diff
@@ -605,21 +541,21 @@ const removeAssignment = (assignment) => (
...
+function unlink(assignment: IAssignmentItem) {
+    return (dispatch, getState, {api, notify}) => (
         api('assignments_unlink').save({}, {
             assignment_id: assignment._id,
             item_id: get(assignment, 'item_ids[0]'),
         })
             .then(() => {
                 notify.success(gettext('Assignment reverted.'));
-                return dispatch(self.unlock(assignment));
+                return planningApi.locks.unlockItem(assignment);
```
petrjasek, line 552: I'm not sure about those changes in calling the actions in general, what's the motivation for that?

MarkLark86: We originally used Redux not just for state management, but as a way to pass AngularJS services from client-core as well. Sometimes we were dispatching an action that never, directly, updated the Redux state. This added an extra layer that didn't add any benefits.

I would rather see Redux being used for state management only, and use the new pure-js API provided by client-core to access functionality provided by client-core. This would mean, eventually, this `actions` folder would only contain dispatch actions that update state and no other side affects.

Therefor actions themselves can be implemented in this internal Planning API. This internal API can be executed from anywhere without acquiring the Redux store, and if need be this API will dispatch the appropriate actions to update the store.

petrjasek: I'm just not sure if implicit access to dispatch in this internal api makes things simpler or more complex at the end if it can be called from anywhere anytime. not against it per se rather wondering if there can be any side effects.

client/actions/events/api.ts
```diff
@@ -23,7 +22,7 @@ import {
     getTimeZoneOffset,
 } from '../../utils';
 
-import planningApi from '../planning/api';
+import planningApis from '../planning/api';
```
petrjasek, line 25: would be good to have some more descriptive naming for this vs `planningApi`, it's not obvious what's the difference

client/api/locks.ts
```diff
@@ -0,0 +1,319 @@
...
+        },
+        (error) => {
+            notify.error(getErrorMessage(error, gettext('Failed to load item locks')));
+
+            return Promise.reject(error);
+        }
+    );
+}
+
+function addLockToStore(data: IWebsocketMessageData['ITEM_LOCKED']): void {
```
petrjasek, line 79: not sure if we should mention the store in the method name, that made me think it's related to localStorage or something, maybe `setLockedState` would do

petrjasek (approved):
I don't consider that naming to be a blocker, but would be good to look at it at some point imo

MarkLark86:
> I don't consider that naming to be a blocker, but would be good to look at it at some point imo

I completely agree. In this PR I changed all imports from `actions/planning/api` to use `planningApis`, and all internal API to use `planningApi`. Where as before it was a mixed bag between the two. So in the future should be able to simply replace `planningApis` with whatever makes sense

PR #1794, opened 2023-04-25
Fix : Some events ingested from Onclusive do not contain the complete location info [SDCP-692]

devketanpro:


server/planning/feed_parsers/onclusive_sample.json
```diff
@@ -13,15 +13,9 @@
...
         "timezoneOffset": -4.00
     },
-    "venue": "One King West Hotel & Residence, 1 King St W, Toronto",
+    "venue": "Karuizawa",
     "tbcVenue": false,
-    "venueData": [
-        {
-            "venueId": 20848,
-            "locationLat": 43.648940,
-            "locationLon": -79.378086
```
petrjasek, line 22: can you also parse those into `location.location.lat` and `location.location.lon`?

devketanpro: but as I checked the event response which you shared, there is not any data in `venueData`

devketanpro: okay

server/planning/feed_parsers/onclusive.py
```diff
@@ -159,6 +159,7 @@ def parse_location(self, event, item):
                     "name": event["venue"],
                     "qcode": "onclusive-venue:{}".format(venue_data.get("venueId")),
                     "address": self.parse_address(event),
+                    "location": {"lat": venue_data.get("locationLat"), "lon": venue_data.get("locationLon")},
```
petrjasek, line 162: only populate `location` this if there is some data, afaik elastic would fail to save the item if there is `None`

petrjasek (approved):

PR #1795, opened 2023-04-27
Add confirmation when removing coverage or scheduled update [SDESK-4865]

devketanpro:


client/components/UI/Form/InputArray/index.tsx
```diff
@@ -58,10 +60,23 @@ export class InputArray extends React.PureComponent<IProps> {
...
     remove(index: number) {
-        this.props.onChange(
-            this.props.field,
-            (this.props.value ?? []).filter((value, i) => i !== index)
-        );
+        const {gettext} = superdeskApi.localization;
+        const {confirm} = superdeskApi.ui;
+
+        confirm(
+            gettext('Remove Coverage', {
```
MarkLark86, line 67: 2nd argument is not required here, as there are no variables to interpolate

client/components/UI/Form/InputArray/index.tsx
```diff
@@ -6,6 +6,8 @@ import {get} from 'lodash';
 import {Button} from '../../';
 import {Row, LineInput} from '../';
 import './style.scss';
+import {notify} from 'superdesk-core/scripts/core/notify/notify';
```
MarkLark86, line 9: `notify` can be imported from `superdeskApi` (in the same way `confirm` is imported)

MarkLark86 (approved):

PR #1800, opened 2023-05-05
[SDESK-6744] Improve displaying of coverage icons & avatars

dzonidoo:


client/components/Coverages/CoverageIcons.tsx
```diff
@@ -0,0 +1,260 @@
+import * as React from 'react';
+import moment from 'moment-timezone';
+import {getUserInitials} from './../../components/UserAvatar';
+import * as config from 'appConfig';
+import {IPlanningCoverageItem, IG2ContentType, IContactItem, IPlanningConfig} from '../../interfaces';
+import {IUser, IDesk} from 'superdesk-api';
+import {gettext} from 'superdesk-core/scripts/core/utils';
```
MarkLark86, line 7: `gettext` should be used from the `superdeskApi.localization` interface

client/components/Coverages/CoverageIcons.tsx
```diff
@@ -0,0 +1,260 @@
...
+            kind: 'plus-button',
+            icon: icon,
+        };
+
+        return placeholder;
+    } else {
+        const avatar: Omit<IPropsAvatar, 'size'> = {
+            initials: getUserInitials(user.display_name),
+            imageUrl: user.picture_url,
+            tooltip: user.display_name?.length > 0 ? user.display_name : null,
```
MarkLark86, line 80: If `display_name` is not available on the user, should maybe we use a combination of `first_name` and `last_name`?

client/components/Coverages/CoverageIcons.tsx
```diff
@@ -0,0 +1,260 @@
...
+    }
+}
+
+export class CoverageIcons extends React.PureComponent<IProps> {
+    render() {
+        const {coverages, users} = this.props;
+
+        return (
+            <WithPopover
+                placement="bottom-end"
```
MarkLark86, line 94: Will this still work if the list item as at the bottom of the Browser window?

client/components/Coverages/coverage-icons.scss
```diff
@@ -0,0 +1,23 @@
+.coverages-popup {
+    background-color: var(--color-dropdown-menu-Bg);
+    border-radius: 5px;
```
MarkLark86, line 3: To provide a consistent and predictable interface, I would suggest we use variables defined in the ui-framework, including radius, padding etc.

* [border-radius: var(--b-radius--medium)](https://github.com/superdesk/superdesk-ui-framework/blob/develop/app/styles/design-tokens/_design-tokens-general.scss#L13)
* [sd-font-size--x-small](https://github.com/superdesk/superdesk-ui-framework/blob/develop/app/styles/_helpers.scss#L957)
* [Paddings & margins](https://ui-framework.superdesk.org/#/components/spacing)

I could not find any CSS var for `line-height`, so maybe changing to using `rem` instead of `px`.

All these are important to be able to support different themes for Superdesk. It may be a requirement for an accessibility theme to provide bigger margins, less rounded corners etc, and hard coding these values will not support this.

client/components/Planning/PlanningItem.tsx
```diff
@@ -244,7 +244,7 @@ class PlanningItemComponent extends React.Component<IProps, IState> {
                             </span>
                         )}
                     </Row>
-                    <Row>
+                    <Row style={{overflow: 'visible'}}> {/** overflow is needed for coverage icons */}
```
MarkLark86, line 247: We already have a `classes` prop for this component, you could use `sd-overflow--visible` here instead

client/components/Coverages/CoverageIcons.tsx
tomaskikutis: yep, it will switch to `top-end` if there's not enough space at the bottom

client/components/Coverages/CoverageIcons.tsx
tomaskikutis: Good catch. I'll update it. I didn't import this myself - simply kept it from the previous component which I have refactored :upside_down_face:

client/components/Coverages/CoverageIcons.tsx
tomaskikutis: I think it's never unavailable. When user gets created, username is set as display name and when you change first/last names display name gets updated as well.

client/components/Planning/PlanningItem.tsx
tomaskikutis: ok, I'll update it.

client/components/Coverages/coverage-icons.scss
tomaskikutis: I've updated it. I'd say we should use more variables separately rather than adding a long list of classes to HTML elements. One reason is to avoid repetition like in this file. Another is that working this way you can't see all style declarations in one place for a particular classname which makes working on it a lot slower.

petrjasek (approved):

PR #1803, opened 2023-05-15
Fix: Update "Some Coverage Assigned" Filter to Include Items with At Least One Coverage is not Assigned [SDESK-6743]

devketanpro:


server/planning/search/queries/planning.py
```diff
@@ -280,6 +280,13 @@ def search_coverage_assignment_status(params: Dict[str, Any], query: elastic.Ela
                     query=elastic.bool_query(must=[elastic.exists(field="coverages.assigned_to.assignment_id")]),
                 )
             )
+            query.must.append(
+                elastic.nested(
+                    path="coverages",
+                    query=elastic.bool_query(must_not=[elastic.exists(field="coverages.assigned_to.assignment_id")]),
```
petrjasek, line 286: you can probably combine it with the previous query if one is must and another is must_not

devketanpro: I tried previously, but that will not worked actually we want item which has atleast one coverage is not assgined and some coverage is assigned (one or two).

MarkLark86:
@devketanpro Can you please change the pull request title so that it describes what you're fixing here

server/planning/search/queries/planning.py
devketanpro: the combined approach does not work since it checks both conditions on a single coverage and never matches.

petrjasek (approved):

PR #1806, opened 2023-05-17
Improvement the behavior to remove the date label and day browser on small screens [SDESK-6930]

devketanpro:


client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -203,7 +203,7 @@ class PlanningListSubNavComponent extends React.Component<IProps, IState> {
                         </div>
                     )}
 
-                    <ButtonGroup align="end">
+                    <ButtonGroup className = "btnGroup" align="end">
```
petrjasek, line 206: not sure why this `ButtonGroup` has `btnGroup` class but the other one not, it should be either more specific or set on all of them

client/apps/style.scss
```diff
@@ -195,3 +195,7 @@ a:hover {
 .sd-filters-panel--fixed {
     flex-shrink: 0;
 }
+
+.sd-page-content--slide-in--open .btnGroup .icn-btn, .sd-page-content--slide-in--open .btnGroup .btn, .sd-page-content--slide-in--open .btnGroup > span{
```
petrjasek, line 199: not sure how this is active only on small screens? would expect some media query here

client/apps/Planning/PlanningListSubNav.tsx
devketanpro: we're applying that `btnGroup` because the modifications are only needed here. If we alter the `ButtonGroup`, the changes may be reflected elsewhere.

petrjasek: I get this, but then we should use some more specific class name here instead of `btnGroup` which would make sense to be on all of those and not just here

fritzSF: There is no need for creating a separate button group here, it can go in the same as the firsat dropdown.

fritzSF (changes requested):
The behaviour is still not OK. There are four different layout scenarios to consider here:

- Just the list view open
- List view + preview open
- List view + editor open
- List view + preview + editor open

I suggest using a container query here instead of a media query. This would be much simpler, having all these options in mind.

A few other things:
There is no space in front of the "Assigned to:" label.
Image 2023-05-17 at 8 08 12 PM
Was it a request that the "Day, Week, Month" dropdown remains instead of the calendar dropdown like it was before?
Image 2023-05-17 at 8 28 32 PM

"Assigned to:" should have a span with a class around it. Like this:
`<span class="sd-margin-l--1 sd-opacity--75 ">Assigned to:</span>`

client/apps/style.scss
```diff
@@ -195,3 +195,9 @@ a:hover {
 .sd-filters-panel--fixed {
     flex-shrink: 0;
 }
+
+@media (max-width: 1400px) {
```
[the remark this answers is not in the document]
fritzSF: A container query would be much simpler here as there are four layot options to consider.

fritzSF (changes requested):
This is not good.
`.sd-page-content--slide-in--open` is not contained in the subnav, so the query can't work.

Try this instead:
`@container subnav (max-width: 800px) {
        .hideOnMobile {
            .icn-btn, .btn, .subnav-calendar  {
              display: none;
            }
        }
}`

Please test before pushing the changes.

client/apps/Planning/PlanningListSubNav.tsx
```diff
@@ -191,7 +191,7 @@ class PlanningListSubNavComponent extends React.Component<IProps, IState> {
                     </ButtonGroup>
                     {this.props.activefilter == PLANNING_VIEW.EVENTS ? ' ' : (
                         <div>
-                            {gettext('Assigned to: ')}
+                            <span className="sd-margin-l--1 sd-opacity--75 ">{gettext('Assigned to :')}</span>
```
petrjasek, line 194: we should keep keep that string `gettext('Assigned to:')` without space before `:` otherwise this will be a new string for transifex

devketanpro: yep

fritzSF (approved):

PR #1807, opened 2023-05-19
Update the Desk validation on the coverage assignment form and the Coverage assignment filter based on the key 'planning_auto_assign_to_workflow' [SDESK-6743]

devketanpro:


server/planning/search/queries/planning.py
```diff
@@ -270,7 +270,16 @@ def search_coverage_assignment_status(params: Dict[str, Any], query: elastic.Ela
             query.must_not.append(
                 elastic.nested(
                     path="coverages",
-                    query=elastic.bool_query(must=[elastic.exists(field="coverages.assigned_to.assignment_id")]),
+                    query=elastic.bool_query(
```
MarkLark86, line 273: This change is not related to Coverage Desk validation, and would be better suited to a separate PR

devketanpro: okay I revert these changes and create a separate PR for this.

client/components/Assignments/AssignmentEditor.tsx
```diff
@@ -182,6 +186,14 @@ export class AssignmentEditorComponent extends React.Component {
                 desk: value,
                 filteredUsers: getUsersForDesk(value, this.props.users),
             });
+
+            // Remove the validation error for the "desk" field if PLANNING_AUTO_ASSIGN_TO_WORKFLOW is false
+            if (!appConfig.planning_auto_assign_to_workflow) {
```
MarkLark86, line 191: Directly above this we're calling `this.onChange`, inside this function we're calling `setState` with the new list of errors.

Adding in a `setState` here would mean 2 state updates, it is preferable if we can do this in 1 state update

MarkLark86: Actually looking further into this, you might not even need any changes here at all.
Instead, what you were to check the config and apply validation here: https://github.com/superdesk/superdesk-planning/blob/develop/client/validators/assignments.tsx#L4

devketanpro: Done, Thanks Mark!!

MarkLark86 (approved):

PR #1812, opened 2023-06-14
Fix event template

thecalcc:
SDESK-6892


tomaskikutis (changes requested):
ensure gettext is used everywhere

tomaskikutis (approved):

PR #1815, opened 2023-06-20
Fix: update styles for fix chrome specific issue in monitoring view [SDESK-6963]

devketanpro:


client/apps/style.scss
```diff
@@ -196,7 +196,7 @@ a:hover {
     flex-shrink: 0;
 }
 
-.subnav {
+.subnav.subnav--light {
```
tomaskikutis, line 199: could you improve naming? The styles here aren't making the subnav light. It'd be better to add an extra classname for this and name it according to the function of these styles.

devketanpro: yep

PR #1824, opened 2023-07-12
fixing language in vocabulary select (multilingual mode)

dzonidoo:
SDESK-6999

### DIFF
--- a/client/components/CustomVocabulariesFields.tsx
+++ b/client/components/CustomVocabulariesFields.tsx
@@ -3,8 +3,6 @@ import {get} from 'lodash';
 
 import {IVocabulary} from 'superdesk-api';
 import {superdeskApi} from '../superdeskApi';
-
-import {getUserInterfaceLanguageFromCV} from '../utils/users';
 import {SelectMetaTermsInput, Field} from './UI/Form';
 
 interface IProps {
@@ -27,13 +25,12 @@ export default class CustomVocabulariesFields extends React.PureComponent<IProps
             popupProps,
             popupContainer,
             testId,
+            language,
         } = this.props;
 
         const {
             errors,
         } = fieldProps;
-
-        const language = this.props.language ?? getUserInterfaceLanguageFromCV();
 
         return customVocabularies
             .map((cv) => (
--- a/client/components/CustomVocabulariesFields.tsx
+++ b/client/components/CustomVocabulariesFields.tsx
@@ -18,7 +18,6 @@ interface IProps {
 }
 
 export default class CustomVocabulariesFields extends React.PureComponent<IProps> {
-
     render() {
         const {gettext} = superdeskApi.localization;
         const {
--- a/client/components/CustomVocabulariesFields.tsx
+++ b/client/components/CustomVocabulariesFields.tsx
@@ -17,6 +17,7 @@ interface IProps {
 }
 
 export default class CustomVocabulariesFields extends React.PureComponent<IProps> {
+
     render() {
         const {gettext} = superdeskApi.localization;
         const {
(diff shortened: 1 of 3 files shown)


client/components/CustomVocabulariesFields.tsx
```diff
@@ -32,7 +33,10 @@ export default class CustomVocabulariesFields extends React.PureComponent<IProps
             errors,
             diff,
         } = fieldProps;
-        const language = get(diff, 'language') || getUserInterfaceLanguageFromCV();
+
+        const language = MAIN_LANGUAGE ?? get(diff, 'language') ?? getUserInterfaceLanguageFromCV();
```
tomaskikutis, line 37: Hi @MarkLark86 I want to consult regarding the approach to fix this ticket. In multilingual mode there is the [component](https://github.com/superdesk/superdesk-planning/assets/2076953/17a9ab04-43b8-4803-99ae-2578d5021a34) to select the main language. When the language is changed, it saves it to the state. BUT when we are rendering a vocabulary select, we use `item.language`. Should we fix it by always updating `item.language` when the value of language select component changes? Or should we rather not touch `item.language` and use that state variable that stores the main language?

MarkLark86: Hi @tomaskikutis & @dzonidoo 

This `Main Language` selector is purely for view purposes and nothing to do with the data layer, therefor my recommendation would be to use the Editor's state.

This component is rendered by the Editor, using the [EditorFieldCustomVocabularies](https://github.com/superdesk/superdesk-planning/blob/develop/client/components/fields/editor/CustomVocabularies.tsx) component. The Editor passes in the `language` that should be used for rendering (see [IEditorFieldProps](https://github.com/superdesk/superdesk-planning/blob/develop/client/interfaces.ts#L1494) for the interface).

My suggested solution here would be to:
* Remove any language logic from within this component
* Add a `language` prop to this component
* Pass through the `language` prop in the `EditorFieldCustomVocabularies` component

This way the language used in this component is controlled elsewhere, in this case the Editor itself.

client/components/CustomVocabulariesFields.tsx
```diff
@@ -31,12 +31,9 @@ export default class CustomVocabulariesFields extends React.PureComponent<IProps
 
         const {
             errors,
-            diff,
         } = fieldProps;
 
-        const language = MAIN_LANGUAGE ?? get(diff, 'language') ?? getUserInterfaceLanguageFromCV();
-
-        console.log(diff, 'diff');
+        const language = this.props.language ?? getUserInterfaceLanguageFromCV();
```
tomaskikutis, line 36: move `getUserInterfaceLanguageFromCV()` out of this file - apply it before passing language prop to this file

client/components/UI/Form/SelectMetaTermsInput/SelectFieldPopup.tsx
```diff
@@ -280,10 +280,16 @@ export class SelectFieldPopup extends React.Component<IProps, IState> {
 
         const valueNoCase = val.toLowerCase();
         const searchKey = this.props.searchKey ?? 'name';
-        let searchResults = this.props.options.filter((opt) => (
-            opt[searchKey].toLowerCase().substr(0, val.length) === valueNoCase ||
-                opt[searchKey].toLowerCase().indexOf(valueNoCase) >= 0
-        ));
+
+        let searchResults = this.props.options.filter((opt) => {
```
tomaskikutis, line 284: try removing code duplication here: use a conditional for setting a function that would get a label in a correct language, but reuse the same label comparison logic without duplicating code

tomaskikutis:
@dzonidoo remember to remove draft state of a PR when it's ready for review [image](https://github.com/superdesk/superdesk-planning/assets/2076953/5bb60b02-e3a5-438a-8338-af62ff837c41)

I've pushed one commit to clean up your function. In your version, filtering algorithm was conditional(via if/else) when in fact it the same one was being used in both cases, the only difference being the label. I've made the code match the intention - label is conditional and filtering algorithm is the same.

Also in your version, the function name was called `itemTranslation` when the function itself wasn't doing anything specific to translations.

tomaskikutis (approved):

PR #1826, opened 2023-07-14
copy language and translation when creating a planning from event and vice versa

tomaskikutis:
SDESK-7001

### DIFF
--- a/client/actions/agenda.ts
+++ b/client/actions/agenda.ts
@@ -248,54 +248,64 @@ const createPlanningFromEvent = (
     event: IEventItem,
     planningDate: Moment = null,
     agendas: Array<string> = []
-) => (
-    (dispatch) => (
-        dispatch(planning.api.save({}, {
-            event_item: event._id,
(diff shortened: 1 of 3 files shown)


MarkLark86 (approved):

client/actions/agenda.ts
```diff
@@ -248,54 +248,64 @@ const createPlanningFromEvent = (
...
-                'ednote',
-                event.ednote
-            ),
-            agendas: agendas,
-            language: event.language,
-        }))
-    )
-);
+) => {
+    const newPlanningItem: Partial<IPlanningItem> = {
```
MarkLark86, line 252: I'm wondering if we should extract this out into a util function, and used both here and https://github.com/superdesk/superdesk-planning/blob/develop/client/api/editor/item_events.ts#L83, so we're only defining this conversion once

tomaskikutis: I agree, but it seems it's being done in a different way in `createPlanningFromEvent` and `addPlanningItem`. Is it possible to drop one and use the remaining one in both places?

PR #1830, opened 2023-07-24
Make it possible to set the name field in the events and planning items to be expandable [SDESK-7010]

devketanpro:


server/planning/content_profiles/profiles/event.py
```diff
@@ -31,7 +31,7 @@ class EventSchema(BaseSchema):
     language = LanguageField()
     links = schema.ListField()
     location = schema.StringField()
-    name = StringField(required=True)
+    name = TextField(required=True, field_type="multi_line")
```
MarkLark86, line 34: `field_type="single_line"` might be a better default here, as names would most often than not be single line.

It might also be good to do this same change for `PlanningSchema.name` field as well

devketanpro: sounds good Thanks !!

MarkLark86 (approved):

PR #1833, opened 2023-07-27
improve naming of definition_short

tomaskikutis:
When label was "Description" it matches planning item's "Description" label and creates confusion why translations in multi-lingual mode were not copied over.

SDESK-7001

### DIFF
--- a/client/components/fields/resources/events.ts
+++ b/client/components/fields/resources/events.ts
@@ -21,7 +21,7 @@ registerEditorField(
     'definition_short',
     EditorFieldMultilingualText,
     () => ({
-        label: superdeskApi.localization.gettext('Description'),
+        label: superdeskApi.localization.gettext('Short Description'),
         field: 'definition_short',
     }),
     null,


client/components/fields/resources/events.ts
```diff
@@ -21,7 +21,7 @@ registerEditorField(
     'definition_short',
     EditorFieldMultilingualText,
     () => ({
-        label: superdeskApi.localization.gettext('Description'),
+        label: superdeskApi.localization.gettext('Short Description'),
```
MarkLark86, line 24: I would keep this as `Description` and fix the copying metadata between Event & Planning item translations (as this currently works with single language configs).

We're already copying `event.definition_short` to `planning.description_text`. Therefor I believe the fix here should be to copy `event.translations.definition_short` (if defined) to `planning.translations.description_text`.

See:
* https://github.com/superdesk/superdesk-planning/blob/develop/client/api/editor/item_events.ts#L94
* https://github.com/superdesk/superdesk-planning/blob/develop/client/actions/agenda.ts#L279

PR #1845, opened 2023-08-22
[SDESK-7029] fix(ui): Show validation for CustomVocabulary fields

MarkLark86:
- Removed redundant `CustomVocabulariesFields` component
- Created specific validation function for custom vocabularies
- Render to `SelectMetaTermsInput` directly instead of going through `Field` first
- Fix css to show proper background colour when invalid
- Fix `EditorFieldVocabulary` with showing invalid state


MarkLark86:
@tomaskikutis It would be ideal to replace all input fields with the UI-Framework equivalent. But I think this task should wait till v2.8 development so I continued using the custom select input for now.

client/validators/profile.ts
```diff
@@ -85,3 +90,39 @@ export const formProfile = ({field, value, profile, errors, messages, diff}) =>
...
+    }
+
+    planningApi.vocabularies.getCustomVocabularies()
+        .filter((cv) => schema.vocabularies?.includes(cv._id))
+        .forEach((cv) => {
+            const values = ((diff[cv.schema_field || 'subject'] || []) as Array<ISubject>).filter((item) => (
+                item.scheme === cv._id
+            ));
+            const fieldName = `custom_vocabularies.${cv._id}`;
+            const fieldLabel = gettext(cv.display_name).toUpperCase();
```
tomaskikutis, line 119: only static strings should be passed to gettext. I think we don't support translating CV names at the moment and when we do support it should come from DB as it now works with translatable vocabulary values.

client/validators/profile.ts
```diff
@@ -85,3 +90,39 @@ export const formProfile = ({field, value, profile, errors, messages, diff}) =>
...
+    const schema = profile.schema?.[field];
+
+    if (schema == null || schema.type !== 'list' || profile.editor[field]?.enabled !== true) {
+        return;
+    } else if (schema.required !== true) {
+        return;
+    }
+
+    planningApi.vocabularies.getCustomVocabularies()
+        .filter((cv) => schema.vocabularies?.includes(cv._id))
```
tomaskikutis, line 113: `.filter((cv) => (schema.vocabularies ?? []).includes(cv._id))`

a default would be better here - it's easier to quickly verify that output will be a boolean we won't get surprises by returning `undefined` from callback function

tomaskikutis (approved):

PR #1847, opened 2023-08-31
Unlink as coverage

thecalcc:
SDESK-7017


tomaskikutis:
can you fix CI before assigning for review? :) [image](https://github.com/superdesk/superdesk-planning/assets/2076953/f1a934c9-312b-46f1-ade2-add962ae342c)

client/planning-extension/src/extension.ts
```diff
@@ -112,6 +112,32 @@ const extension: IExtension = {
             contributions: {
                 entities: {
                     article: {
+                        getActions: (item) => [
+                            {
+                                label: 'Unlink as Coverage',
```
tomaskikutis, line 117: use gettext

tomaskikutis (commented):
CI is still failing :)

thecalcc:

tomaskikutis:
which ticket is this handling?

thecalcc:
> which ticket is this handling?

SDESK-7017

client/planning-extension/src/extension.ts
```diff
@@ -105,14 +105,40 @@ function onSendBefore(superdesk: ISuperdesk, items: Array<IArticle>, desk: IDesk
...
                     article: {
+                        getActions: (item) => [
+                            {
+                                label: gettext('Unlink as Coverage'),
+                                groupId: 'planning-actions',
+                                icon: 'cut',
+                                onTrigger: () => {
+                                    const superdeskArticle = superdesk.entities.article;
+
+                                    if (
```
tomaskikutis, line 125: Did you take these conditions from somewhere else? In order not to duplicate it, we should add `canUnlinkAscoverage` function and expose it via extensions bridge.

thecalcc: The thing is, we can't add it 1-to-1, because in the angular world permissions work under the hood, so if we do add it it'll still be chopped up. Personally I don't think it's necessary when this is the case, so we can keep it like this

tomaskikutis: add the comment about keeping it up to date

tomaskikutis (approved):

tomaskikutis:
one more thing, add `TAG: AUTHORING-ANGULAR` where it's registered on the angular side. I've started adding the tags too so when time comes to remove angular based authoring, we can search for tags and easily find all related code that needs to be removed as well.

index.ts
```diff
@@ -103,6 +103,8 @@ function configurePlanning(superdesk) {
                         !['killed', 'recalled', 'unpublished', 'spiked', 'correction'].includes(item.state);
                 }],
         })
+
+        // TAG: AUTHORING-ANGULAR
```
tomaskikutis, line 107: image

I'm wondering about that `external-app` action.

image

It looks like the same action also appears in the topbar. Could you double check whether we also have it in authoring-react?

tomaskikutis:
did you see my last comment?

index.ts
tomaskikutis: @thecalcc have you looked into this yet?

thecalcc: I'm doing so yep, but I plan to address it in a different PR. haven't forgotten about it

thecalcc: > I'm doing so yep, but I plan to address it in a different PR. haven't forgotten about it

In short, looks like this is some filtering based on which we calculate whether to show the action in the three dots menu or not

PR #1848, opened 2023-08-31
Fulfil assignment

thecalcc:
SDESK-7015

What this PR achieves:
1. Makes fulfil assignment available in both authoring react and angular through superdesk `getActions` api
2. Removes the old angular registration of this action


client/planning-extension/src/extension.ts
```diff
@@ -13,6 +13,9 @@ import {AssignmentsList} from './assignments-overview';
 import {IPlanningExtensionConfigurationOptions} from './extension_configuration_options';
 import {AutopostIngestRuleEditor} from './ingest_rule_autopost/AutopostIngestRuleEditor';
 import {AutopostIngestRulePreview} from './ingest_rule_autopost/AutopostIngestRulePreview';
+import ng from 'superdesk-core/scripts/core/services/ng';
```
tomaskikutis, line 16: Extensions must not import anything neither from core nor planning. Instead either use superdesk api or planning api via bridge.

tomaskikutis (commented):
fix CI

thecalcc:

thecalcc:
Address last comment from https://github.com/superdesk/superdesk-planning/pull/1847

client/planning-extension/src/extension.ts
```diff
@@ -135,6 +137,28 @@ const extension: IExtension = {
                                     ) {
                                         const event = new CustomEvent('planning:unlinkfromcoverage', {detail: {item}});
 
+                                        window.dispatchEvent(event);
+                                    }
+                                },
+                            },
+                            {
```
tomaskikutis, line 151: I'm wondering whether this would appear in authoring-angular as well?

thecalcc: if you're asking wether there would be a conflict thus two of the same actions rendering because of being registered in two places yes. I plan to address this in another PR, together with the `external-app` thingy, as this is something that affects all actions so we need a generic solution. 

I think doing a simple collision detection mechanism would suffice for now. If we have an action with the same id coming from extensions as well, we don't push it to the e.g. `allActions` array. We skip that one, while prioritising existing ones. In react we don't show the angular registered ones, so this should be enough to ignore the react side for now. Later when we remove the angular implementation we will remove the filtering

tomaskikutis: Extensions API have access to `authoringReactViewEnabled` variable. What about using it to conditionally push the action?

Alternativelly, what about if we dropped the old action and kept only the new one? I did that with [planning details widget](https://github.com/superdesk/superdesk-planning/pull/2078) - dropped the angular version of a widget - and the react one works in both - angular and react based authoring.

client/planning-extension/src/extension.ts
```diff
@@ -123,12 +123,12 @@ const extension: IExtension = {
...
                             {
                                 label: gettext('Unlink as Coverage'),
-                                groupId: 'planning-actions',
+                                groupId: planningActionsGroupId,
                                 icon: 'cut',
                                 onTrigger: () => {
                                     const superdeskArticle = superdesk.entities.article;
 
-                                    // keep in sync with client/planning-extension/src/extension.ts:123
+                                    // keep in sync with index.ts:108
```
tomaskikutis: better keep relative path. There are lots of index files

client/planning-extension/src/extension.ts
```diff
@@ -140,9 +140,38 @@ const extension: IExtension = {
...
+                            {
+                                label: superdesk.localization.gettext('Fulfil assignment'),
+                                groupId: planningActionsGroupId,
+                                icon: 'calendar-list',
+                                onTrigger: () => {
+                                    const itemStates = ['killed', 'recalled', 'unpublished', 'spiked', 'correction'];
+                                    const {isContentLinkToCoverageAllowed} = extensionBridge.assignments.utils;
 
-                                        window.dispatchEvent(event);
+                                    // keep in sync with index.ts:79
```
tomaskikutis: link relative path

tomaskikutis (approved):

client/planning-extension/src/extension.ts
thecalcc: If you mean the absolute it would be something like superdesk-planning/index.ts

If relative, then this is it, and coincidentally index.ts really is the path & file name, relative to the project.

tomaskikutis: I meant relative to the project. You get it by right clicking open file tab and choosing "copy relative path"

tomaskikutis:
why did you drop "keep in sync" comments?

thecalcc:
> why did you drop "keep in sync" comments?

Because I've removed the registrations from angular

tomaskikutis (approved):

PR #1849, opened 2023-09-05
Coverages created via Event creation form should be directly posted alongside with the associated Planning item [SDESK-7051]

devketanpro:


client/components/fields/resources/profiles.ts
```diff
@@ -34,6 +34,17 @@ registerEditorField(
     true
 );
 
+registerEditorField(
+    'schema.planning_auto_publish',
+    EditorFieldCheckbox,
+    () => ({
+        label: superdeskApi.localization.gettext('Planning Auto Publish'),
```
petrjasek, line 41: the label should be rather something like `Post planning items with Event`

server/planning/events/events_post.py
```diff
@@ -203,13 +203,22 @@ def publish_event(self, event, version):
...
         docs = []
+        if new_post_state != POST_STATE.CANCELLED:
+            if event_profile_schema.get("related_plannings", {}).get("planning_auto_publish"):
+                docs = [
+                    {
+                        "planning": planning[config.ID_FIELD],
+                        "etag": planning.get("etag"),
+                        "pubstatus": POST_STATE.USABLE,
+                    }
+                    for planning in plannings
```
petrjasek, line 218: we would like to only post the planning items which were updated since the last time, maybe that's the case and it's handled via that `.post` call, would be good to add some tests for this scenario

client/api/editor/item_events.ts
```diff
@@ -137,6 +137,10 @@ export function getEventsInstance(type: EDITOR_TYPE): IEditorAPI['item']['events
             return;
         }
 
+        if (updates && original?.version == original?.posted_version) {
+            original['version'] = original?.version ? original['version'] + 1 : 1;
```
petrjasek, line 141: isn't there any version field that would be handling it already? like `_current_version` ?

petrjasek: I see, maybe we can use the `versioncreated` then and have `versionposted` so we wouldn't have to maintain the version manually, there might be more places when it should be updated and it might get out of sync, like when it would get updated via ingest

devketanpro: Okay let me try , Thanks !

PR #1859, opened 2023-10-09
Planning: When starting to work on an Assignment prefill the article with the fields matching the language of the Coverage [SDESK-7061]

devketanpro:


server/planning/archive.py
```diff
@@ -38,6 +38,12 @@ def create_item_from_template(doc, extra_fields_to_override=None):
     # and apply them if any found
     updates = {key: val for key, val in doc.items() if key in fields_to_override}
 
+    # override translated values if any
+    translated_value = {
+        entry["field"]: entry["value"] for entry in translations or [] if entry["language"] == doc.get("language")
+    }
+    updates = {key: val for key, val in translated_value.items() if key in fields_to_override}
```
MarkLark86, line 45: This is overwriting not extending what the original `updates` value is. This is probably why you're getting failing tests.

server/planning/archive.py
```diff
@@ -38,6 +38,12 @@ def create_item_from_template(doc, extra_fields_to_override=None):
     # and apply them if any found
     updates = {key: val for key, val in doc.items() if key in fields_to_override}
 
+    # override translated values if any
+    translated_value = {
```
MarkLark86, line 42: Why not just run this code when `if translations is not None:`.

MarkLark86 (approved):

PR #1862, opened 2023-10-12
Planning: If multilingual Events or Planning items are filtered by language - show them in this language in the list views [SDESK-7060]

devketanpro:


client/components/Main/ListGroupItem.tsx
```diff
@@ -265,3 +274,5 @@ export class ListGroupItem extends React.Component<IProps, IState> {
         return null;
     }
 }
+
+export const ListGroupItem = connect(mapStateToProps)(ListGroupComponent);
```
MarkLark86, line 278: It's not a good idea to connect list items to the Redux store. If the list gets big, then  there will be 1 store listener per list item.

It would be better for performance to add the selector/param to a higher up component

client/components/fields/index.tsx
```diff
@@ -66,6 +66,23 @@ export function renderFields(fields, item, props = {}) {
     });
 }
 
+/**
+ * Get translated field value based on languagei
+ * @param {String} language
+ * @param {Object} item
+ * @param {String} fieldName
+ */
+export function getTranslatedValue(language, item, fieldName) {
```
MarkLark86, line 75: It would be good to use typescript types here, and not rely on jsdoc types

client/components/fields/headline.tsx
```diff
@@ -1,9 +1,12 @@
 import PropTypes from 'prop-types';
+import {getTranslatedValue} from '.';
 
-export const headline = ({item}) => item.headline || null;
+export const headline = ({item, filterLanguage}) => getTranslatedValue(filterLanguage, item, 'headline') ||
```
MarkLark86, line 4: When modifying existing functions, it would be good to add typescript types. We have a lot of code from when the source was javascript, so adding types when we modify the code is one way to add types everywhere (eventually)

MarkLark86 (changes requested):

client/components/fields/headline.tsx
devketanpro: yep

client/apps/Planning/PlanningList.tsx
```diff
@@ -206,6 +210,7 @@ export class PlanningListComponent extends React.PureComponent<IProps> {
                     listViewType={listViewType}
                     sortField={sortField}
                     indexItems
+                    searchParams= {currentSearch.advancedSearch}
```
MarkLark86, line 213: space after `=` not desirable for code style

client/components/fields/name.tsx
```diff
@@ -1,9 +1,13 @@
 import PropTypes from 'prop-types';
+import {getTranslatedValue} from '.';
+import {IFieldsProps} from '../../interfaces';
 
-export const name = ({item}) => item.name || null;
+export const name = ({item, filterLanguage}: IFieldsProps) => getTranslatedValue(filterLanguage, item, 'name') ||
```
MarkLark86, line 5: We're already sending a `language` field when rendering list item fields. (see https://github.com/superdesk/superdesk-planning/blob/develop/client/components/fields/index.tsx#L48). I believe this was done to show CVs items in users language.

These now seem like competing implementations. Rather than add a `filterLanguage` to individual components, it might be better to add a `language` prop to these components.

This could be done by:
1. Add a `language` argument to the `renderFields` function
2. Remove `const language = getUserInterfaceLanguage();` from within the `renderFields` function
3. In `EventItem` / `PlanningItem` components, pass in `filterLanguage || item.language || getUserInterfaceLanguageFromCV()` to the call of `renderFields`
4. Add `language` prop to the list field components where you want to handle this language (such as `name`, `slugline`, `headline` etc)

MarkLark86 (changes requested):

client/apps/Planning/PlanningList.tsx
devketanpro: my bad !, I forgot to remove it

MarkLark86 (approved):

PR #1863, opened 2023-10-12
coverage search for user

dzonidoo:
SDBELGA-752


client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
```diff
@@ -18,6 +18,7 @@ import {Row, SelectUserInput} from '../../../UI/Form';
 import {EditorFieldNewsCoverageStatus} from '../NewsCoverageStatus';
 import * as config from 'appConfig';
 import {getLanguagesForTreeSelectInput} from '../../../../selectors/vocabs';
+import {SelectUser} from 'superdesk-core/scripts/core/ui/components/SelectUser';
```
MarkLark86, line 21: It would be ideal not to directly import anything from `superdesk-core`, instead we can import components from the SuperdeskAPI (see: https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/superdesk-api.d.ts#L1810).

MarkLark86:
@dzonidoo E2E tests need to be fixed

tomaskikutis (changes requested):
@dzonidoo fix that import Mark pointed out

tomaskikutis (approved):

PR #1864, opened 2023-10-12
log each event id we process from onclusive

petrjasek:
so we can check logs in case some event is missing

CPCN-380


MarkLark86 (approved):

MarkLark86:
@petrjasek Might need to update the superdesk-core requirement to v2.6.5: https://github.com/petrjasek/superdesk-planning/blob/fix-onclusive-add-logging/server/requirements.txt#L22 (due to PyYAML issues https://github.com/petrjasek/superdesk-planning/blob/fix-onclusive-add-logging/server/requirements.txt#L22)

PR #1868, opened 2023-10-20
Merge hotfix 2.6.3 develop

MarkLark86:


MarkLark86:
@petrjasek The only thing I'm not 100% sure about is the onclusive ingest, and the group dates (front end). Could you double check those please. Everything else seemed straightforward

petrjasek:
checked those, looks good

petrjasek (approved):

PR #1869, opened 2023-10-20
add setup for e2e server docker

petrjasek:
to run use:

```bash
$ cd e2e
$ docker-compose build
$ docker-compose up
```

### DIFF
--- a/e2e/server/core-requirements.txt
+++ b/e2e/server/core-requirements.txt
@@ -1,0 +1,3 @@
+gunicorn==19.7.1
+honcho==1.0.1
+git+https://github.com/superdesk/superdesk-core.git@develop#egg=superdesk-core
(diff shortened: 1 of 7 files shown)


e2e/server/core-requirements.txt
```diff
@@ -0,0 +1,3 @@
+gunicorn==19.7.1
+honcho==1.0.1
+git+https://github.com/superdesk/superdesk-core.git@develop#egg=superdesk-core
```
MarkLark86, line 3: We should probably move this requirement to `setup.py`, so it's not required here nor in `server/requirements.txt`

MarkLark86 (approved):

e2e/server/core-requirements.txt
petrjasek: I actually wanted to have it separate so it can be cached by docker

PR #1873, opened 2023-10-31
Planning compatability fixes with client-core post-broadcasting

thecalcc:
SDESK-7102


client/components/UI/TermsList.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import {gettext} from 'superdesk-core/scripts/core/utils';
+import {gettext} from 'superdesk-core/scripts/core/utils.tsx';
```
MarkLark86, line 2: This import can be replaced with the SuperdeskAPI, and use `localization.gettext` in the render function

thecalcc: We won't need this PR since Petr decided that we use post-broadcasting branch instead. So I'm closing this PR.

PR #1876, opened 2023-11-08
Update Assignment article preview to support multilingual features [SDESK-7061]

devketanpro:


server/planning/planning/planning.py
```diff
@@ -735,6 +735,27 @@ def _create_update_assignment(
...
+        if translations is not None:
+            translated_value.update(
+                {
+                    entry["field"]: entry["value"]
+                    for entry in translations or []
+                    if entry["language"] == doc.get("planning", {}).get("language")
+                }
+            )
+
+            translated_name = translated_value.get("name", translated_value.get("headline"))
```
MarkLark86, line 749: `translated_name` will need to be defined before this if block, otherwise checking this variable later will result in a `UnboundLocalError: local variable 'translated_name' referenced before assignment` error.

I think this is why the E2E tests are failing

MarkLark86 (approved):

PR #1877, opened 2023-11-09
fix: Prevent unintended synchronization of multilingual text fields[SDESK-7061]

devketanpro:
Here we resolved an issue where typing in one multilingual text field unintentionally caused the first letter of the word to appear in the default language field.

issue snap-video : https://ctrlv.tv/EJHV

### DIFF
--- a/client/components/fields/editor/base/multilingualText.tsx
+++ b/client/components/fields/editor/base/multilingualText.tsx
@@ -20,7 +20,7 @@ export class EditorFieldMultilingualText extends React.Component<IMultilingualTe
     constructor(props: IMultilingualTextProps) {
         super(props);
 
-        this.state = {fields: this.getFieldStates()};
+        this.state = {fields: this.getFieldStates(true)};
         this.onChange = this.onChange.bind(this);
     }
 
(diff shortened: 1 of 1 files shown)


client/components/fields/editor/base/multilingualText.tsx
```diff
@@ -53,7 +53,7 @@ export class EditorFieldMultilingualText extends React.Component<IMultilingualTe
         // If the entry for the default language has not been assigned to the item
         // then this item may have been created before multilingual functionality
         // therefor show the value from the parent field
-        if (translations[multilingualConfig.defaultLanguage] == null) {
+        if (translations[multilingualConfig.defaultLanguage]) {
```
MarkLark86, line 56: I'm not sure how this solves the problem.

Example of the differing logic:
* **field:** name
* **defaultLanguage:** en

**Original logic:**
```
if item.translations.name[en] is not defined:
    item.translations.name[en] = item.name
```
This provides `name[en]` a value (from `item.name`) if it does not have one

**Updated logic:**
```
if item.translations.name[en] has a value:
    item.translations.name[en] = item.name
```
This overrides `name[en]` with `item.name` every time state updates

**Suggestion:**
Instead what we can do is set `name[en] = item.name` when the component first mounts, by providing a boolean argument to `getFieldStates`. When this arg is true, then we can perform the original logic to set `name[en]`

MarkLark86 (changes requested):

MarkLark86 (approved):

PR #1892, opened 2023-12-22
[SDBELGA-759] Feature: Sync Event metadata to Planning & Coverages

MarkLark86:
Also moves the EmbeddedPlanning from a front-end feature to be a back-end feature (so that metadata sync and associated planning can be handled with the 1 request, and the 1 DB change).


MarkLark86:
@petrjasek The code is ready to be reviewed. I have placed this PR in draft state while I add tests

petrjasek (approved):
looks good, what's your take on using dataclasses vs typed dicts, should we use both, only one, or have some kind of guidelines?

MarkLark86:
> looks good, what's your take on using dataclasses vs typed dicts, should we use both, only one, or have some kind of guidelines?

I don't mind the use of dataclasses, as it provides:
* both compile and runtime type checking 
* dot notation (which I find easier to read vs bracket notation).
* strict typing (for cases where you know the exact structure of the data)

On the other hand, dataclasses should **never** be used with data that will be used in a:
* response to a request from the API
* request for a job/task through celery

This is because the data would have to be serialised, which means converting it to a dict first, in which case a dict should be used.

Thoughts @petrjasek?

petrjasek:
makes sense

PR #1895, opened 2024-01-10
Selection of Planning filter is lost on scrolling down and up

dzonidoo:
SDESK-7153


tomaskikutis:
why was it happening?

tomaskikutis (approved):

PR #1898, opened 2024-01-17
As a planning user I want to see up to 5 templates recently used by me in the Event creation pull down menu [SDBELGA-763]

devketanpro:


client/components/Main/CreateNewSubnavDropdown.tsx
```diff
@@ -16,14 +16,42 @@ interface IProps {
...
+    recentTemplatesId: Array<string>;
 }
 
 class CreateNewSubnavDropdownFn extends React.PureComponent<IProps> {
+    constructor(props: IProps) {
+        super(props);
+        this.getRecentTemplates = this.getRecentTemplates.bind(this);
+    }
+
+    getRecentTemplates() {
```
MarkLark86, line 28: This is better off as a selector, using `createSelector` ([example](https://github.com/superdesk/superdesk-planning/blob/develop/client/selectors/vocabs.ts#L29)). Then this can be passed directly as a prop.

Don't forget the ts types

client/components/Main/CreateNewSubnavDropdown.tsx
```diff
@@ -16,14 +16,42 @@ interface IProps {
...
+            addPlanning,
+            createPlanningOnly,
+            privileges,
+            createEventFromTemplate,
+            recentTemplatesId,
+            eventTemplates
+        } = this.props;
         const items: Array<IDropdownItem> = [];
 
+        const recentTemplates = this.getRecentTemplates().sort(
```
MarkLark86, line 51: Why are the recent templates sorted? I would have though the sort order would be the order of last used to least used

MarkLark86 (changes requested):

client/components/Main/CreateNewSubnavDropdown.tsx
devketanpro: to maintain this order **last used to least used**, we used sort here.

MarkLark86 (approved):

PR #1900, opened 2024-01-18
Hide the text 'Agenda:No Agenda Assigned.' in the related planning section [SDESK-7136]

devketanpro:
PR based on latest comment :- https://sofab.atlassian.net/browse/SDESK-7136?focusedCommentId=126957

### DIFF
--- a/client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
@@ -8,7 +8,7 @@ import {
     IProfileSchemaTypeList,
     ISearchProfile
 } from '../../../../interfaces';
-import {superdeskApi} from '../../../../superdeskApi';
+import {planningApi, superdeskApi} from '../../../../superdeskApi';
 
 import {ButtonGroup, Button} from 'superdesk-ui-framework/react';
 import {Row} from '../../../UI/Form';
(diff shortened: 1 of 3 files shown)


client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
```diff
@@ -32,6 +32,7 @@ interface IProps extends IEditorFieldProps {
 export class EditorFieldEventRelatedPlannings extends React.PureComponent<IProps> {
     render() {
         const {gettext} = superdeskApi.localization;
+        const isAgendaEnabled = planningApi.planning.getEditorProfile().editor.agendas.enabled;
```
MarkLark86, line 35: `getEditorProfile` may not be required here, can you check if `profile` is provided in IProps (see https://github.com/superdesk/superdesk-planning/blob/develop/client/interfaces.ts#L1527)

devketanpro: I already checked this, but here we got event profile, and in event profile we did not get `agenda`, so that's why we fetch planning profile here.

MarkLark86: Understood

MarkLark86 (approved):

PR #1901, opened 2024-01-19
test unit tests fails

devketanpro:


tomaskikutis:
fixed by https://github.com/superdesk/superdesk-planning/pull/1904

PR #1821, opened 2023-06-30
Add to planning action refactoring

thecalcc:
SDESK-6267

PR #1907, opened 2024-02-07
Add to planning refactoring

thecalcc:
SDESK-6267

Continuation of [this](https://github.com/superdesk/superdesk-planning/pull/1821) auto-closed PR


tomaskikutis (changes requested):
fix conflicts, address comments from autoclosed PR

thecalcc:

client/planning-extension/src/extension.ts
```diff
@@ -121,6 +121,31 @@ const extension: IExtension = {
                 entities: {
                     article: {
                         getActions: (item) => [
+                            {
```
tomaskikutis: If conditions aren't satisfied - the action should not be displayed. The way you implemented it - it is still displayed, but clicking it doesn't do anything :smile:

tomaskikutis (commented):
Did you find out what is this about?

```
{
    action: 'external-app',
    type: 'addto-planning',
},
```

thecalcc:
> Did you find out what is this about?
> 
> ```
> {
>     action: 'external-app',
>     type: 'addto-planning',
> },
> ```

I'll address it in another PR, it's a filter for the actions. Basically responsible for displaying groups of actions.

client/planning-extension/src/extension.ts
thecalcc: I've come to the same realisation, that's why in the fulfil assignment action I added an `else` clause to notify the user they can't execute the action. For the future I can adjust the condition easily.

tomaskikutis: What I don't get is why don't you modify `getActions` function so that if conditions aren't satisfied - it would return an empty array and no actions would be displayed to a user as a result?

thecalcc:
Based on a search across all repos I can confirm the impact of refactoring this part of the logic won't affect any customers:
image

tomaskikutis:
If `external-apps` extension point was used only in one place and we removed that usage is it safe to drop the entire extension point? If so, let's do it.

tomaskikutis (approved):

thecalcc:
> If `external-apps` extension point was used only in one place and we removed that usage is it safe to drop the entire extension point? If so, let's do it.

Don't mistake `external-app` for `external-appS`. There's only a single place in the code related to `external-app` which I haven't removed, as I removed all the registered objects matching this whole `external-app` code. Although I can remove the "connection" itself as well

tomaskikutis:
yeah, I meant removing `external-app` extension point. I think we can remove the entire `sdItemActionByIntent` directive from core.

thecalcc:
> yeah, I meant removing `external-app` extension point. I think we can remove the entire `sdItemActionByIntent` directive from core.

yep, i just removed it in client-core

PR #1909, opened 2024-02-08
Templates modal implementation

thecalcc:
SDBELGA-764


thecalcc:
image

tomaskikutis:
image

Add a gap horizontally between select and filter input. Make the two have equal height.

fritzSF:
Hi @thecalcc, the design is not OK. You have the HTML code here: https://ui-framework.superdesk.org/#/playgrounds/react/test-ground (click on the "PLANNING TEMPLATES" button at the top). No need for improvisation.

tomaskikutis:
@fritzSF we agreed we'll be [removing dropdown](https://sofab.atlassian.net/browse/SDUF-97) component from the framework, so I asked Konstantin not to use it in new code. We'll try to make it looks the same.

client/components/PlanningTemplatesModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+        const TemplatesListView: React.FC<ICloseModalProps> = ({closeModal}) => {
+            const calendarFilter = (template: IEventTemplate) => this.state.activeCalendarFilter
+                ? template.data.calendars.map(({qcode}) => qcode).includes(this.state.activeCalendarFilter)
+                : template;
+            /**
+             * Groups the templates by calendar, filters the templates that match the current search query,
+             * filters the templates by the current selected calendar if one is selected,
+             * filters out the groups that don't have any templates
+             */
+            const filteredTemplates = calendars.map((calendar) => ({
```
tomaskikutis, line 45: It's not very readable. Maybe it would be cleaner to take templates, filter them, preferrably using 1 filter and only in last step to group it?

client/components/PlanningTemplatesModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+                calendarName: calendar.name,
+                templates: eventTemplates
+                    .filter((template) => template.template_name.includes(this.state.searchQuery))
+                    .filter((template) => template.data.calendars.map(({qcode}) => qcode).includes(calendar.qcode))
+                    .filter(calendarFilter)
```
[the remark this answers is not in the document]
thecalcc: If I return a template then a template gets returned, so the filter is practically fake. I did it for the purpose of not splitting the code based on `activeCalendarFilter` being or not being set.

client/components/PlanningTemplatesModal/TemplatesListView.tsx
```diff
@@ -0,0 +1,75 @@
...
+    searchQuery,
+    createEventFromTemplate,
+}: ITemplatesListViewProps) => {
+    /**
+     * Groups the templates by calendar,
+     * filters the templates that match the current search query,
+     * if a calendar is selected filters the groups that match the calendar are filtered,
+     * if not the groups that that don't have any templates are filtered out.
+     */
+    const filteredTemplates = calendars
```
tomaskikutis, line 29: image


This is very inefficient. It will probably not cause significant performance issues, but try getting used to writing code in an efficient way when it's not any harder like here. One thing to keep in mind is minimize iterations. You will not remove iterations entirely - but when you iterate, make sure you are iterating over less and less items in every iteration. Another rule is to avoid coupling which you did well here.

here's an example how you can do it more efficiently

```
const templatesMatchingSearchQuery = eventTemplates.filter((template) => template.template_name.includes(searchQuery));

const templatesByCalendar = {}; // only those matching search query

for (const template of templatesMatchingSearchQuery) {
    for (const calendar of template.data.calendars ?? []) {
        if (templatesByCalendar[calendar.qcode] == null) {
            templatesByCalendar[calendar.qcode] = [];
        }

        templatesByCalendar[calendar.qcode].push(template);
    }
}

const filteredCalendars = activeCalendarFilter == null
    ? props.calendars
    : [props.calendars.find(({qcode}) => qcode === activeCalendarFilter)];

const filteredCalendarsWithTemplates = filteredCalendars.map((calendar) => ({
    calendar: calendar,
    templates: templatesByCalendar[calendar.qcode],
}));
 ```
 
It's also not only about of number of iterations but also how performance intensive is each iteration. For example, I could have first grouped templates by calendars and only then performed filtering by template name on matching templates. If one calendar is selected in UI that text comparison would be done only on templates in one calendar.

**Review by dzonidoo — APPROVED** [review]

File: client/components/PlanningTemplatesModal/PlanningTemplatesModal.tsx [hunk 4]
```diff
@@ -0,0 +1,105 @@
...
+
+    render(): React.ReactNode {
+        const {gettext} = superdeskApi.localization;
+        const {closeModal, createEventFromTemplate, calendars, eventTemplates} = this.props;
+        const allCalendarsLabel = gettext('All Calendars');
+        const calendarDropdownItems = [];
+        const activeCalendarName = this.props.calendars
+            .find((cal) => cal.qcode === this.state.activeCalendarFilter)?.name;
+        const dropdownLabel = this.state.activeCalendarFilter
+            ? `${gettext('Calendar')}: ${activeCalendarName}`
```
petrjasek, line 37: gettext doesn't work well with string literals, better use:
```ts
gettext('Calendar: {{activeCalendarName}}', {activeCalendarName});
```

petrjasek (approved):

thecalcc:
@MarkLark86 @devketanpro can you please check the failing test on your side locally? I can't get it to run properly thus can't fix it. From my POV the input hasn't changed so it should be working correctly, but please check it if you have the e2e setup working correctly locally.

PR #1911, opened 2024-02-09
[SDBELGA-760] feature: Ability to update a series of planning items

MarkLark86:


client/api/events.ts
```diff
@@ -144,6 +144,7 @@ function create(updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
                 internal_note: coverage.planning.internal_note,
             })),
         })),
+        update_method: updates.update_method?.value ?? updates.update_method
```
petrjasek, line 147: I wonder where is this `.value` coming from?

petrjasek (approved):

client/api/events.ts
MarkLark86: @petrjasek It's coming from the [modal forms](https://github.com/superdesk/superdesk-planning/tree/develop/client/components/ItemActionConfirmation/forms). To fix that required more work outside the scope of this change, so I figured it would probably be better to change that separately in a different PR

PR #1917, opened 2024-02-15
Fix e2e testcases

devketanpro:


client/components/Main/CreateNewSubnavDropdown.tsx
```diff
@@ -74,7 +80,8 @@ class CreateNewSubnavDropdownFn extends React.PureComponent<IProps> {
                     });
                 });
 
-            if (sortedTemplates.length > MORE_TEMPLATES_THRESHOLD) {
+            if (recentTemplates.length < MORE_TEMPLATES_THRESHOLD ||
+                sortedTemplates.length > MORE_TEMPLATES_THRESHOLD) {
```
thecalcc, line 84: Why do you need to check if `recentTemplates` < 5? Regardless of the count of `recentTemplates` if `sortedTemplates` > 5 we need to show that button.

client/components/Main/CreateNewSubnavDropdown.tsx
```diff
@@ -63,7 +63,13 @@ class CreateNewSubnavDropdownFn extends React.PureComponent<IProps> {
             const sortedTemplates = eventTemplates
                 .sort((templ1, templ2) => templ1.template_name.localeCompare(templ2.template_name));
 
-            recentTemplates
+            /**
+             * If recent templates has empty then in this case show eventtemplates.
+             */
```
thecalcc, line 68: I think the code is self explanatory here. No need for a comment

client/components/Main/CreateNewSubnavDropdown.tsx
```diff
@@ -63,7 +63,13 @@ class CreateNewSubnavDropdownFn extends React.PureComponent<IProps> {
...
+
+            const templates = recentTemplates.length == 0 ? sortedTemplates : recentTemplates;
```
thecalcc, line 70: Use `===` instead of `==`

client/components/Main/CreateNewSubnavDropdown.tsx
devketanpro: okay

client/components/Main/CreateNewSubnavDropdown.tsx
devketanpro: my bad

client/components/Main/CreateNewSubnavDropdown.tsx
devketanpro: We have to show that more button in both cases, like if recentTemplates < 5 or sortedTemplates > 5.
Let's assume we have `sortedTemplates ==4` and in `recentTemplates` we have `3` templates, but the user wants to use that one template, which is not in `recentTemplates`.

thecalcc: Okay makes total sense, let's keep it then

thecalcc (approved):

PR #1919, opened 2024-02-20
Add Belga user search using credentials to the reassign modal [SDBELGA-770]

devketanpro:


client/components/Assignments/AssignmentEditor.tsx
```diff
@@ -17,6 +17,7 @@ import {
     SelectUserInput,
 } from '../UI/Form';
 import {ContactsPreviewList, SelectSearchContactsField} from '../Contacts';
+import {SelectUser} from 'superdesk-core/scripts/core/ui/components/SelectUser';
```
tomaskikutis, line 20: Don't import things from superdesk-core. Instead consume it from superdesk api

`const {SelectUser} = superdeskApi.components;`

client/components/Assignments/AssignmentEditor.tsx
```diff
@@ -297,15 +298,18 @@ export class AssignmentEditorComponent extends React.Component {
...
-                        field={this.FIELDS.USER}
-                        label={gettext('User')}
-                        value={this.state.user}
-                        onChange={this.onUserChange}
-                        users={this.state.filteredUsers}
-                        popupContainer={popupContainer}
-                        readOnly={disableUserSelection}
-                    />
+                    <Row style={{padding: '2rem 0', margin: '0 0 1.8em 0'}}>
+                        <SelectUser
```
tomaskikutis, line 302: Check the props. It looks to me that you're passing some e.g. field that don't exist in the interface.

client/components/Assignments/AssignmentEditor.tsx
```diff
@@ -297,15 +298,18 @@ export class AssignmentEditorComponent extends React.Component {
                         />
                     </Row>
                 ) : (
-                    <SelectUserInput
```
tomaskikutis, line 300: Besides this, only one usage remains of this component. Can you replace it too and remove this component completelly?

client/components/Assignments/AssignmentEditor.tsx
devketanpro: I tried, but when we run testcases, we got  `SelectUser`  is `undefined`

tomaskikutis: It will not work if you try to define that constant at the top level. Try definining it inside the component.

client/components/Assignments/AssignmentEditor.tsx
```diff
@@ -297,15 +297,18 @@ export class AssignmentEditorComponent extends React.Component {
...
+                    <Row style={{padding: '2rem 0', margin: '0 0 1.8em 0'}}>
+                        <SelectUser
+                            selectedUserId = {this.state.userId}
+                            onSelect={(user) => {
+                                this.onUserChange(null, user);
+                            }}
+                            autoFocus={false}
+                            horizontalSpacing={true}
+                            clearable={true}
+                            data-test-id= {this.FIELDS.USER}
```
tomaskikutis, line 309: Component doesn't have `data-test-id` prop defined in the interface. It already has "select-user-dropdown" data-test-id that you can use in tests when needed.

client/components/Coverages/CoverageAddAdvancedModal.tsx
```diff
@@ -276,18 +278,26 @@ export class CoverageAddAdvancedModal extends React.Component<IProps, IState> {
...
+                                                    >
+                                                        <SelectUser
+                                                            selectedUserId = {coverage.user?._id}
+                                                            onSelect={(user) => {
+                                                                this.onUserChange(coverage, user);
+                                                            }}
+                                                            autoFocus={false}
+                                                            horizontalSpacing={true}
+                                                            clearable={true}
+                                                            data-test-id= {'user'}
```
tomaskikutis, line 297: Component doesn't have `data-test-id` prop defined in the interface. It already has "select-user-dropdown" data-test-id that you can use in tests when needed.

client/components/Assignments/AssignmentEditor.tsx
tomaskikutis: image

Then you need to find where is it set, investigate why is it not set when running tests and resolve the issue.

tomaskikutis (changes requested):

client/components/Assignments/AssignmentEditor.tsx
tomaskikutis: I saw you removed usages, but also go ahead and remove the component itself.

tomaskikutis (approved):

PR #1936, opened 2024-03-21
set recurrence_id based on event _id

petrjasek:
so we can later check which event is the initial one

CPCN-695

### DIFF
--- a/server/planning/events/events.py
+++ b/server/planning/events/events.py
@@ -223,7 +223,7 @@ class EventsService(superdesk.Service):
                 recurring_events = generate_recurring_events(event)
                 generated_events.extend(recurring_events)
                 # remove the event that contains the recurring rule. We don't need it anymore
-                docs.remove(event)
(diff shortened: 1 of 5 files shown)


server/planning/events/events.py
```diff
@@ -223,7 +223,7 @@ def on_create(self, docs):
                 recurring_events = generate_recurring_events(event)
                 generated_events.extend(recurring_events)
                 # remove the event that contains the recurring rule. We don't need it anymore
-                docs.remove(event)
+                docs.remove(event)  # todo: why we remove that event and not update it?
```
MarkLark86, line 226: I think we can fix this by skipping the first event inside `generate_recurring_events` function, that way the passed in Event will not need to be removed.

MarkLark86 (approved):

server/planning/events/events.py
petrjasek: I see, will avoid further changes at this stage but might fix it later

PR #1940, opened 2024-04-01
Fix : Some coverage items are not displaying in Newspro [SDCP-755]

devketanpro:


server/planning/assignments/assignments.py
```diff
@@ -1299,17 +1299,12 @@ def _publish_planning(item):
             try:
                 # check if the planning item is locked
                 lock_service.validate_relationship_locks(planning_item, "planning")
-                use_published_planning = False
             except SuperdeskApiError as ex:
                 # planning item is already locked.
-                use_published_planning = True
-                logger.exception(str(ex))
+                lock_service = get_component(LockService)
+                lock_service.unlock(planning_item, user_id, get_auth()["_id"], "planning")
```
petrjasek, line 1305: I wonder if we need to unlock or deal with locks at all when updating assignment

petrjasek (approved):

MarkLark86:
@petrjasek @devketanpro Does this cause any issues for the person that had the Planning item locked in the first place?
i.e.

1. [UserA] Create a Planning item with a text Coverage
2. [UserA] Add the text Coverage to Workflow
3. [UserA] Make changes to the Planning and Coverage metadata (DO NOT SAVE)
4. [UserB] Create a Text Content item
5. [UserB] Link Text Content to Assignment (via the `Fulfil Assignment` action in monitoring)
6. [UserA] Save the Planning item

Question: Are the changes UserA made to the Planning & Coverage kept? Does the Planning Editor update with the changes from the linking to Assignment, is are these changes kept or lost? I believe this is why that lock checking was added in the first place

PR #1941, opened 2024-04-02
Validation of planning items when posted alongside events [SDESK-7202]

devketanpro:


petrjasek (commented):
would be good to have it covered with tests

devketanpro:
> would be good to have it covered with tests

will add tests for this !

petrjasek (approved):

PR #1957, opened 2024-04-08
Update notifications data for belga specfic templates [SDBELGA-781]

devketanpro:
https://github.com/superdesk/superdesk-belga/pull/444

### DIFF
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -455,8 +455,8 @@ class AssignmentsService(superdesk.Service):
             assignment["planning"]["ics_data"] = ical.to_ical()
 
         # get formatted contacts and event date time for email templates
-        formatted_contacts = get_formatted_contacts(event_item)
-        fomatted_event_date = get_event_formatted_dates(event_item)
+        formatted_contacts = get_formatted_contacts(event_item) if event_item else []
+        fomatted_event_date = get_event_formatted_dates(event_item) if event_item else ""
 
         # The assignment is to an external contact or a user
         if assigned_to.get("contact") or assigned_to.get("user"):
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -479,7 +479,7 @@ class AssignmentsService(superdesk.Service):
                         is_link=True,
                         contact_id=assigned_to.get("contact"),
                         contacts=formatted_contacts,
-                        location=event["LOCATION"],
+                        location=event.get("LOCATION", ""),
                         event_date_time=fomatted_event_date,
                     )
                     # notify the desk
@@ -498,7 +498,7 @@ class AssignmentsService(superdesk.Service):
                             omit_user=True,
                             is_link=True,
                             contacts=formatted_contacts,
-                            location=event["LOCATION"],
+                            location=event.get("LOCATION", ""),
                             event_date_time=fomatted_event_date,
                         )
 
@@ -537,7 +537,7 @@ class AssignmentsService(superdesk.Service):
                             is_link=True,
                             contact_id=assigned_to.get("contact"),
                             contacts=formatted_contacts,
-                            location=event["LOCATION"],
+                            location=event.get("LOCATION", ""),
                             event_date_time=fomatted_event_date,
                         )
                     else:
@@ -562,7 +562,7 @@ class AssignmentsService(superdesk.Service):
                             is_link=True,
                             contact_id=original.get("assigned_to").get("contact"),
                             contacts=formatted_contacts,
-                            location=event["LOCATION"],
+                            location=event.get("LOCATION", ""),
                             event_date_time=fomatted_event_date,
                         )
                         # notify the assignee
@@ -588,7 +588,7 @@ class AssignmentsService(superdesk.Service):
                             is_link=True,
                             contact_id=assigned_to.get("contact"),
                             contacts=formatted_contacts,
-                            location=event["LOCATION"],
+                            location=event.get("LOCATION", ""),
                             event_date_time=fomatted_event_date,
                         )
             else:  # A new assignment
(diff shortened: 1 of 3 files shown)


server/planning/utils.py
```diff
@@ -1,10 +1,81 @@
...
+
+
+def time_short(datetime: datetime):
+    if datetime:
+        return format_time(parse_date(datetime), "HH:mm")
+
+
+def date_short(datetime: datetime):
+    if datetime:
+        return format_datetime(parse_date(datetime), "dd/MM/yyyy")
```
petrjasek, line 61: imo we will need the format configurable

server/planning/utils.py
```diff
@@ -1,10 +1,81 @@
...
+                            c
+                            for c in [
+                                contact.get("first_name"),
+                                contact.get("last_name"),
+                            ]
+                            if c
+                        ]
+                    ),
+                    "organisation": contact.get("organisation", ""),
+                    "email": ", ".join(contact.get("contact_email", [])),
```
petrjasek, line 35: would keep that data as lists, it's easy to join it in the template, but might be harder to split it later

server/planning/utils.py
```diff
@@ -1,10 +1,81 @@
...
 
 
 def try_cast_object_id(value: str) -> Union[ObjectId, str]:
     try:
         return ObjectId(value)
     except InvalidId:
         return value
+
+
+def get_formatted_contacts(event: Dict[str, Any]) -> List[Dict[str, Union[str, List[str]]]]:
```
petrjasek, line 17: would be useful to define a typed dict for the return value

server/planning/assignments/assignments.py
```diff
@@ -473,6 +478,9 @@ def send_assignment_notification(self, updates, original=None, force=False):
                         event=event_item,
                         is_link=True,
                         contact_id=assigned_to.get("contact"),
+                        contacts=formatted_contacts,
+                        location=event.get("LOCATION", ""),
```
petrjasek, line 482: do we actually use capital `LOCATION` ?

devketanpro: yes

server/planning/utils.py
```diff
@@ -1,10 +1,85 @@
...
+from typing import Union, List, Dict, Any
 from bson.objectid import ObjectId
 from bson.errors import InvalidId
+from datetime import timedelta, datetime
+from flask_babel import format_time, format_datetime, lazy_gettext
+from eve.utils import str_to_date
+import arrow
+from flask import current_app as app
+
+ContactDict = Dict[str, Union[str, List[str]]]
```
petrjasek, line 10: what I ment is something like:
```py3
class FormattedContact(TypedDict):
    name: str
    organisation: str
    email: List[str]
    ...
```

server/planning/utils.py
```diff
@@ -1,10 +1,85 @@
...
+        return format_time(parse_date(datetime), app.config.get("TIME_FORMAT_SHORT", "HH:mm"))
+
+
+def date_short(datetime: datetime):
+    if datetime:
+        return format_datetime(parse_date(datetime), app.config.get("DATE_FORMAT_SHORT", "dd/MM/yyyy"))
+
+
+def get_event_formatted_dates(event: Dict[str, Any]) -> str:
+    DAY_IN_MINUTES = 24 * 60 - 1
```
petrjasek, line 69: what's this `-1`?

server/planning/utils.py
```diff
@@ -1,10 +1,85 @@
...
+    start = event.get("dates", {}).get("start")
+    end = event.get("dates", {}).get("end")
+
+    if start + timedelta(minutes=DAY_IN_MINUTES) < end:
```
petrjasek, line 73: I think it would be more readable like:
```py
if end == start:
   ...
if end - start > timedelta(days=1):
   ...
if end - start == timedelta(days=1):
   ...

File: server/planning/utils.py [hunk 8]
```diff
@@ -1,10 +1,85 @@
...
+    if datetime:
+        return format_time(parse_date(datetime), app.config.get("TIME_FORMAT_SHORT", "HH:mm"))
+
+
+def date_short(datetime: datetime):
+    if datetime:
+        return format_datetime(parse_date(datetime), app.config.get("DATE_FORMAT_SHORT", "dd/MM/yyyy"))
+
+
+def get_event_formatted_dates(event: Dict[str, Any]) -> str:
```
petrjasek, line 68: this func can use some unit tests

server/planning/utils.py
devketanpro: the -1 is subtracting one minute to avoid counting the very last minute of the day. So instead of counting 
`24 hours * 60 minutes = 1440` minutes in a day, it's counting 1439 minutes to exclude the last minute. This could be to avoid overlapping with the start time of the next day or to maintain consistency in time calculations

petrjasek: it might make events ending on `23:59:59` handled wrong, I would then expect it to work with seconds and not minutes

server/planning/utils.py
```diff
@@ -17,6 +17,10 @@ class FormattedContact(TypedDict):
     website: str
 
 
+ALL_DAY_SECONDS = 86460
+MULTI_DAY_SECONDS = 86400
```
petrjasek, line 21: it still looks strange to have multi day being lower than all day :sweat_smile: , it can use some doc string

devketanpro: you are right! Thank you for pointing it out! 😊

server/planning/utils.py
```diff
@@ -1,10 +1,97 @@
...
+    name: str
+    organisation: str
+    email: List[str]
+    phone: List[str]
+    mobile: List[str]
+    website: str
+
+
+ALL_DAY_SECONDS = 89999  # Number of seconds for an all-day event
+MULTI_DAY_SECONDS = 90000  # Number of seconds for an multi-day event
```
petrjasek, line 21: it could be like:
```py
MULTI_DAY_SECONDS = 24 * 60 * 60
ALL_DAY_SECONDS = MULTI_DAY_SECONDS - 1
```
otherwise I'm not really sure where are those numbers coming from

petrjasek (approved):

PR #1958, opened 2024-04-08
Implement Bulk add to workflow, improve SlidingToolbar types

thecalcc:
SDESK-7209

### DIFF
--- a/client/api/planning.ts
+++ b/client/api/planning.ts
@@ -200,15 +200,14 @@ function setDefaultValues(
     );
 }
 
-function bulkAddCoverageToWorkflow(planningItems: Array<IPlanningItem>): Array<Promise<IPlanningItem>> {
+function bulkAddCoverageToWorkflow(planningItems: Array<IPlanningItem>): Promise<Array<IPlanningItem>> {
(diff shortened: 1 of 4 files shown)


petrjasek:
which ticket is that @thecalcc ?

thecalcc:
> which ticket is that @thecalcc ?

SDESK-7209

client/api/planning.ts
```diff
@@ -200,6 +200,38 @@ function setDefaultValues(
...
+
+    return planningItemsToUpdate.map((plan) => {
+        const updates = {coverages: cloneDeep(plan.coverages)};
+
+        updates.coverages = plan.coverages
+            .map((coverage) => planningUtils.getActiveCoverage(coverage, coverageStatuses));
+
+        return planning.update(plan, updates)
+            .then((updatedPlan) => {
+                notify.success(gettext('Coverage added to workflow.'));
```
petrjasek, line 219: I wonder if we should show it per item, that could be a lot of notifications, might be good to do some summary at the end

petrjasek (approved):

PR #1963, opened 2024-04-16
add new should_udpate func [SDCP-742]

devketanpro:
[https://github.com/superdesk/superdesk-core/pull/2558](https://github.com/superdesk/superdesk-core/pull/2558)

### DIFF
--- a/server/planning/events/events.py
+++ b/server/planning/events/events.py
@@ -797,6 +797,16 @@ class EventsService(superdesk.Service):
             if events_using_file.count() == 0:
                 files_service.delete_action(lookup={"_id": file})
 
+    def should_update(self, old_item):
+        if (
+            old_item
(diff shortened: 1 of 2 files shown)


server/planning/events/events.py
```diff
@@ -797,6 +797,16 @@ def delete_event_files(self, updates, original):
...
 
+    def should_update(self, old_item, new_item, provider):
+        if (
+            old_item
+            and old_item.get("version_creator")
+            or old_item.get("pubstatus") == "cancelled"
+            or old_item.get("state") == "killed"
+        ):
+            return False
+        return True
```
petrjasek, line 808: would be more readable like:
```py
return old_item is None or old_item.get("version_creator") is None or ...
```
instead of using `if ...: return False or return True`

PR #1969, opened 2024-04-22
optimize validation of related planning items when posted alongside events [SDESK-7202]

devketanpro:
PR based on [Comment](https://sofab.atlassian.net/browse/SDESK-7202?focusedCommentId=129158)
Previously, when related planning item validation failed, the API returned an error, which prevented the Redux state from updating and consequently not reflect the event status in the UI
so I've implemented a workaround here, as when posting an Event, if validation of Event passes and the Event is posted, we respond back to the client with a success. If associated Planning items fail validation, we catch all planning posting errors, and return those to the user along with the response

### DIFF
--- a/client/actions/main.ts
+++ b/client/actions/main.ts
@@ -472,9 +472,7 @@ const post = (original, updates = {}, withConfirmation = true) => (
                     failedPlanningIds?.map((failedItem) => notifyError(
                         notify,
                         failedItem,
-                        gettext(
-                            'Related item {{ itemId }} Post Validation Failed',
-                            {itemId: failedItem?._id})));
+                        gettext('Some related planning item post validation failed')));
 
                     return Promise.resolve(item);
                 },
--- a/client/actions/main.ts
+++ b/client/actions/main.ts
@@ -452,7 +452,17 @@ const post = (original, updates = {}, withConfirmation = true) => (
         return promise
             .then(
                 (rtn) => {
-                    if (!confirmation && rtn) {
+                    let itemId, failedPlanningIds, item;
+
+                    if (Array.isArray(rtn) && rtn.length === 2) {
+                        [item, {itemId, failedPlanningIds}] = rtn;
+                    } else {
+                        itemId = rtn._id;
+                        item = rtn;
+                        failedPlanningIds = [];
+                    }
+
+                    if (!confirmation && item) {
                         notify.success(
                             gettext(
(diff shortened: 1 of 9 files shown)


server/planning/events/events_post.py
```diff
@@ -58,10 +59,15 @@ def create(self, docs):
...
             else:
-                ids.extend(self._post_recurring_events(doc, event, update_method))
-        return ids
+                event_id, planning_ids = self._post_recurring_events(doc, event, update_method)
+
+            ids.append(event_id)
+            if planning_ids:
+                failed_planning_ids.extend(planning_ids)
+
+        return [{"item_id": event_id, "failed_planning_ids": failed_planning_ids}] if failed_planning_ids else ids
```
MarkLark86, line 70: It would be good to have some behave tests around this

client/actions/main.ts
```diff
@@ -452,16 +452,31 @@ const post = (original, updates = {}, withConfirmation = true) => (
...
                                 'The {{ itemType }} has been posted',
                                 {itemType: typeString}
                             )
                         );
                     }
+                    failedPlanningIds?.map((failedItem) => notifyError(
+                        notify,
+                        failedItem,
+                        gettext(
+                            'Related item {{ itemId }} Post Validation Failed',
```
petrjasek, line 476: not sure if the id in message is useful for the user, might be better to just say that some related items failed validation and make it somehow visible in the UI which ones those were

server/features/events_post.feature
```diff
@@ -1185,3 +1185,75 @@ Feature: Events Post
...
+        "etag": "#EVENT1._etag#",
+        "pubstatus": "usable",
+        "update_method": "single"
+    }
+    """
+    Then we get OK response
+    Then we get updated response
+    """
+    {
+     "_id": {
```
petrjasek, line 1252: I wouldn't use `_id` field for this, that's expected to be string in general, can you rather add another field to the response with this data?

client/actions/main.ts
devketanpro: This default message remains hidden from users under normal conditions. It will only be displayed if there's a scenario outside the usual conditions :- https://github.com/superdesk/superdesk-planning/blob/d3d0eb0cc496d12d9475c95317deb396dc6ed5c1/client/utils/index.ts#L250-L276

petrjasek (approved):

PR #1970, opened 2024-04-23
Onclusive parse new key for event status mapping [SDCP-749]

devketanpro:
https://github.com/superdesk/superdesk-cp/pull/233

### DIFF
--- a/server/planning/feed_parsers/onclusive.py
+++ b/server/planning/feed_parsers/onclusive.py
@@ -66,12 +66,10 @@ class OnclusiveFeedParser(FeedParser):
                 GUID_FIELD: guid,
                 ITEM_TYPE: CONTENT_TYPE.EVENT,
                 "state": CONTENT_STATE.INGESTED,
-                # used for event status mapping
-                "is_provisional": event.get("isProvisional", False),
             }
 
             try:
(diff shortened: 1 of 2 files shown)


server/planning/feed_parsers/onclusive.py
```diff
@@ -66,6 +66,8 @@ def parse(self, content, provider=None):
                 GUID_FIELD: guid,
                 ITEM_TYPE: CONTENT_TYPE.EVENT,
                 "state": CONTENT_STATE.INGESTED,
+                # used for event status mapping
+                "is_provisional": event.get("isProvisional", False),
```
petrjasek, line 70: I'm actually wondering why not implement the logic in `set_occur_status` method and only do it in the other repo?

devketanpro: You mean, instead of doing that logic in CP, we should apply that logic here?

petrjasek: right

devketanpro: Actually, I assumed that it was CP-specific, so I didn't apply it here.

petrjasek (approved):

PR #1982, opened 2024-05-07
FIX : Planning filters: 'This week' and 'Next Week' date filter malfunctions [SDESK-7264]

devketanpro:


server/planning/search/queries/common.py
```diff
@@ -394,6 +394,10 @@ def construct_search_query(
 
         # Set `only_future` to False as `construct_query` with request params will add this if neccessary
         filter_params["only_future"] = False
+
+        # Set `time_zone` and start_of_week,  SDESK - 7264
+        filter_params["time_zone"] = params.get("time_zone")
+        filter_params["start_of_week"] = params.get("start_of_week")
```
petrjasek, line 400: I wonder why passing this via client if we have it configured on the server?

devketanpro: may be in the future, we might need this configuration in the UI, which is why we're passing it via the client.

petrjasek:
would be good to have that config tested to make sure it still works in future

petrjasek (approved):

PR #1987, opened 2024-05-15
Fix this_week filter [SDESK-7264]

devketanpro:
Fix items from previous week along with items from this week are also displayed

### DIFF
--- a/server/planning/search/queries/combined.py
+++ b/server/planning/search/queries/combined.py
@@ -22,8 +22,12 @@ def construct_combined_view_data_query(
 
     query = elastic.ElasticQuery()
 
-    if len(search_filter["params"]):
-        search_dates(search_filter["params"], query)
+    filter_params = common.get_params_from_search_filter(search_filter)
+    if len(filter_params):
+        filter_params["time_zone"] = params.get("time_zone")
+        filter_params["start_of_week"] = params.get("start_of_week")
+
+        search_dates(filter_params, query)
 
     search_dates(params, query)
 
--- a/server/planning/search/queries/combined.py
+++ b/server/planning/search/queries/combined.py
@@ -26,7 +26,7 @@ def construct_combined_view_data_query(
     filter_params = common.get_params_from_search_filter(search_filter)
     if len(filter_params):
         filter_params["time_zone"] = params.get("time_zone") or app.config.get("DEFAULT_TIMEZONE")
-        filter_params["start_of_week"] = params.get("start_of_week") or app.config.get("START_OF_WEEK", 0)
+        filter_params["start_of_week"] = params.get("start_of_week", app.config.get("START_OF_WEEK", 0))
 
         search_dates(filter_params, query)
 
--- a/server/planning/search/queries/combined.py
+++ b/server/planning/search/queries/combined.py
@@ -1,6 +1,7 @@
 from typing import Dict, Any, List, Callable
 
 from planning.search.queries import elastic, events, planning, common
+from flask import current_app as app
 
 
 def construct_combined_view_data_query(
@@ -24,8 +25,8 @@ def construct_combined_view_data_query(
 
     filter_params = common.get_params_from_search_filter(search_filter)
     if len(filter_params):
-        filter_params["time_zone"] = params.get("time_zone")
-        filter_params["start_of_week"] = params.get("start_of_week")
+        filter_params["time_zone"] = params.get("time_zone") or app.config.get("DEFAULT_TIMEZONE")
+        filter_params["start_of_week"] = params.get("start_of_week") or app.config.get("START_OF_WEEK", 0)
 
         search_dates(filter_params, query)
 
--- a/server/planning/search/queries/elastic.py
+++ b/server/planning/search/queries/elastic.py
@@ -121,18 +121,18 @@ class ElasticRangeParams:
 
 def start_of_this_week(start_of_week=0, date=None):
     start = get_start_of_next_week(date, start_of_week) - timedelta(days=7)
-    return start.strftime("%Y-%m-%d") + "||/d"
+    return start.strftime("%Y-%m-%d")
 
 
 def start_of_next_week(start_of_week=0, date=None):
-    return get_start_of_next_week(date, start_of_week).strftime("%Y-%m-%d") + "||/d"
+    return get_start_of_next_week(date, start_of_week).strftime("%Y-%m-%d")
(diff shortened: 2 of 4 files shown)


server/planning/search/queries/elastic.py
```diff
@@ -276,7 +279,7 @@ def range_this_week(query: ElasticRangeParams):
             time_zone=query.time_zone,
             value_format=query.value_format,
             gte=start_of_this_week(query.start_of_week),
-            lt=start_of_next_week(query.start_of_week),
+            lt=end_of_this_week(query.start_of_week),
```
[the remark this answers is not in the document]
devketanpro: no this is not ignoring events on that last date see :-
image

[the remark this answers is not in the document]
petrjasek: but looking at the test the query would be like `<2024-05-19T00:00:00` which seems wrong, question is why that event is visible there but I can't see the timestamps, might be due to the timezone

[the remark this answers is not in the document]
devketanpro: yep we have timezone issue,and when we search from sidenav bar filters work perfectly but if we using events&planning filters we got wrong items

server/planning/search/queries/combined.py
```diff
@@ -22,8 +22,12 @@ def construct_combined_view_data_query(
 
     query = elastic.ElasticQuery()
 
-    if len(search_filter["params"]):
-        search_dates(search_filter["params"], query)
+    filter_params = common.get_params_from_search_filter(search_filter)
+    if len(filter_params):
+        filter_params["time_zone"] = params.get("time_zone")
+        filter_params["start_of_week"] = params.get("start_of_week")
```
petrjasek, line 28: I wonder how this will be populated, especially when using those search filters for scheduled exports. imo those values are not stored in search filters atm

MarkLark86: What about add a `time_zone` input for scheduled exports form, and use app config for `start_of_week` on search?

server/planning/search/queries/elastic.py
```diff
@@ -120,12 +120,15 @@ def __init__(
...
-    end = get_start_of_next_week(date, start_of_week) - timedelta(days=7)
-    start = end - timedelta(days=7)
-
+    start = get_start_of_next_week(date, start_of_week) - timedelta(days=7)
     return start.strftime("%Y-%m-%d") + "||/d"
 
 
+def end_of_this_week(start_of_week=0, date=None):
+    end = get_start_of_next_week(date, start_of_week) - timedelta(start_of_week)
+    return end.strftime("%Y-%m-%d") + "||/d"
```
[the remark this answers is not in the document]
devketanpro: when we send the full datetime, all filters misbehave, resulting in items with different dates.

server/planning/search/queries/combined.py
petrjasek: we have also default timezone config, I would probably start with that, not sure if we have a need to use different timezones in those exports

server/planning/search/queries/combined.py
```diff
@@ -22,8 +23,12 @@ def construct_combined_view_data_query(
 
     query = elastic.ElasticQuery()
 
-    if len(search_filter["params"]):
-        search_dates(search_filter["params"], query)
+    filter_params = common.get_params_from_search_filter(search_filter)
+    if len(filter_params):
+        filter_params["time_zone"] = params.get("time_zone") or app.config.get("DEFAULT_TIMEZONE")
+        filter_params["start_of_week"] = params.get("start_of_week") or app.config.get("START_OF_WEEK", 0)
```
petrjasek, line 29: this might be tricky if we start using the params for it and there will be `0`, rather do it like:
```py
params.get("start_of_week", app.config.get("START_OF_WEEK", 0))
```

server/planning/tests/utils_tests.py
```diff
@@ -62,3 +62,12 @@ def test_end_of_next_week(self):
         end_date_str = elastic.end_of_next_week(date=start_date, start_of_week=1)
         expected_weekday = self.get_start_date_and_weekday(end_date_str)
         self.assertEqual(expected_weekday, 0)  # Monday
+
+    def test_events_within_current_week(self):
+        # Test events that start and end within the current week
+        start_date = datetime(2024, 5, 15)  # Assuming today is May 15, 2024 (Wed)
```
petrjasek, line 68: pls add a few more cases, especially for edge cases like monday/sunday

server/planning/tests/utils_tests.py
```diff
@@ -71,3 +71,21 @@ def test_events_within_current_week(self):
...
+        self.assertEqual(end, "2024-05-20||/d")
+
+    def test_events_within_current_week_sunday(self):
+        # Test case for Sunday
+        start_date = datetime(2024, 5, 19)  # May 19, 2024 is a Sunday
+        start = elastic.start_of_this_week(date=start_date, start_of_week=1)
+        end = elastic.start_of_next_week(date=start_date, start_of_week=1)
+
+        self.assertEqual(start, "2024-05-20||/d")
+        self.assertEqual(end, "2024-05-27||/d")
```
petrjasek, line 91: I think we should get rid of that `//d`

server/planning/tests/utils_tests.py
```diff
@@ -69,23 +69,23 @@ def test_events_within_current_week(self):
...
 
     def test_events_within_current_week_sunday(self):
         # Test case for Sunday
         start_date = datetime(2024, 5, 19)  # May 19, 2024 is a Sunday
         start = elastic.start_of_this_week(date=start_date, start_of_week=1)
         end = elastic.start_of_next_week(date=start_date, start_of_week=1)
 
-        self.assertEqual(start, "2024-05-20||/d")
-        self.assertEqual(end, "2024-05-27||/d")
+        self.assertEqual(start, "2024-05-20")
```
petrjasek, line 90: imo this should be 13.5.

server/planning/tests/utils_tests.py
```diff
@@ -69,23 +69,23 @@ def test_events_within_current_week(self):
...
+        self.assertEqual(end, "2024-05-27")
```
petrjasek, line 91: and this one should be 20.5.

devketanpro: thanks for pointing

petrjasek (approved):

PR #1988, opened 2024-05-17
Fix: display of accurate Items using manual filters search [SDESK-7264]

devketanpro:
When using the next_filter from the sidenav bar (Advanced Search), the list is filtered and displays the accurate items. However, when applying filters manually, the data is not filtered correctly. The root cause was identified as the absence of the `advancedSearch.dates.range` parameter in the manual filter process. so when this param is missing, the datetime is not updated from here : https://github.com/superdesk/superdesk-planning/blob/develop/client/utils/index.ts#L690-L728

### DIFF
--- a/client/api/ui/list.ts
+++ b/client/api/ui/list.ts
@@ -34,7 +34,7 @@ function reloadList(params: ICombinedEventOrPlanningSearchParams = {}) {
 
     if (currentView === MAIN.FILTERS.PLANNING) {
         dispatch(actions.eventsPlanning.ui.clearList());
-        dispatch(actions.planning.ui.clearList());
+        dispatch(actions.events.ui.clearList());
         promise = dispatch<any>(actions.planning.ui.fetchToList({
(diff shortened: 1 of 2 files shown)


client/api/ui/list.ts
```diff
@@ -34,15 +34,15 @@ function reloadList(params: ICombinedEventOrPlanningSearchParams = {}) {
 
     if (currentView === MAIN.FILTERS.PLANNING) {
         dispatch(actions.eventsPlanning.ui.clearList());
-        dispatch(actions.events.ui.clearList());
+        dispatch(actions.planning.ui.clearList());
```
MarkLark86, line 37: This is invalid. If the user is currently in the Planning view, then we clear other lists (in this code block, that would be combined and planning lists). So the changes in this file should be reverted

devketanpro: alright

MarkLark86 (approved):

PR #1989, opened 2024-05-21
use name instead of slugline for templates data based on new config [SDBELGA-804]

devketanpro:
https://github.com/superdesk/superdesk-belga/pull/476

### DIFF
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -381,6 +381,10 @@ class AssignmentsService(superdesk.Service):
         coverage_type = updates.get("planning", original.get("planning", {})).get("g2_content_type", "")
         slugline = updates.get("planning", original.get("planning", {})).get("slugline", "with no slugline")
 
+        subject_field = app.config.get("ASSIGNMENT_MAIL_SUBJECT_FIELD")
+        if subject_field:
+            slugline = updates.get(subject_field, original.get(subject_field, slugline))
+
         client_url = app.config["CLIENT_URL"]
 
         assignment = deepcopy(original)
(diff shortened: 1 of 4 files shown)


server/planning/assignments/assignments.py
```diff
@@ -381,6 +381,10 @@ def send_assignment_notification(self, updates, original=None, force=False):
         coverage_type = updates.get("planning", original.get("planning", {})).get("g2_content_type", "")
         slugline = updates.get("planning", original.get("planning", {})).get("slugline", "with no slugline")
 
+        subject_field = app.config.get("ASSIGNMENT_MAIL_SUBJECT_FIELD")
+        if subject_field:
+            slugline = updates.get(subject_field, original.get(subject_field, slugline))
```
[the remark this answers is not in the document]
devketanpro: yep, I'll add it, Thanks

server/planning/planning_notifications.py
```diff
@@ -304,7 +304,7 @@ def _send_user_email(user_id, contact_id, source, meta_message, data):
 
     admins = app.config["ADMINS"]
 
-    data["subject"] = "Superdesk assignment" + ": {}".format(data.get("slugline") if data.get("slugline") else "")
+    data["subject"] = render_template("assignment_mail_subject" + ".txt", **data)
```
petrjasek, line 307: not sure why this string concatenation, you can just do `"assignment_mail_subject.txt"`

server/planning/templates/assignment_mail_subject.txt
```diff
@@ -0,0 +1 @@
+Superdesk assignment planning: {{slugline}}
```
petrjasek, line 1: I think it should be just `Superdesk assignment: {{slugline}}` no need to change it

devketanpro: my bad

server/planning/templates/assignment_mail_subject.txt
```diff
@@ -0,0 +1 @@
+Superdesk assignment : {{slugline}}
```
petrjasek, line 1: without the space before ` :`

PR #1990, opened 2024-05-22
Fix : Wrong date in assignment email [SDBELGA-801]

devketanpro:


server/planning/tests/utils_tests.py
```diff
@@ -2,36 +2,37 @@
...
 from planning.search.queries import elastic
+import pytz
 
 
 class TestGetEventFormattedDates(TestCase):
     def test_multi_day_event(self):
-        start = datetime(2024, 5, 28, 5, 00, 00)
-        end = datetime(2024, 5, 29, 6, 00, 00)
-        event = {"dates": {"start": start, "end": end}}
+        start = datetime(2024, 5, 28, 5, 00, 00, tzinfo=pytz.utc)
```
petrjasek, line 10: not sure if it would change the tests, but afaik we won't have tzinfo set there, so would do the tests without it

devketanpro: we includes timezone info in the start and end datetime of an event:- 
image

petrjasek: yep it looks like it, but that's just how it's formatted it's always adding `+0000`. in db it's stored without timezone afaik

devketanpro: [here](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/assignments/assignments.py#L460) also I checked the date time we got in this format `2024-04-27 18:30:00+00:00`, that's why I added `tzinfo`

devketanpro: when we didn't include timezone information in the datetime objects for the test cases, the tests failed inconsistently on our CI system.

server/planning/tests/utils_tests.py
```diff
@@ -8,30 +8,30 @@ class TestGetEventFormattedDates(TestCase):
...
     def test_dates_same_and_different_time(self):
-        start = datetime(2024, 5, 28, 6, 00, 00)
-        end = datetime(2024, 5, 28, 5, 00, 00)
-        event = {"dates": {"start": start, "end": end}}
+        start = datetime(2024, 5, 28, 5, 00, 00)
+        end = datetime(2024, 5, 28, 6, 00, 00)
+        event = {"dates": {"start": start, "end": end, "tz": "Asia/Calcutta"}}
         result = get_event_formatted_dates(event)
-        self.assertEqual(result, "06:00 - 05:00, 28/05/2024")
+        self.assertEqual(result, "05:00 - 06:00, 28/05/2024")
```
petrjasek, line 34: imo this should be actually `10:30-11:30` if the timezone offset is `+5:30`

petrjasek (approved):

PR #1991, opened 2024-05-23
[STT-6] Feature(backend): Support linking multiple Events into a Planning item

MarkLark86:


server/planning/events/events_schema.py
```diff
@@ -325,6 +322,37 @@
...
+                "link_type": {
+                    "type": "string",
+                    "required": True,
+                    "default": "primary",
+                    "allowed": ["primary", "secondary"],
+                },
+            },
+        },
+        "mapping": {
+            "type": "nested",
```
petrjasek, line 348: I'm wondering why using nested here, wouldn't it work without it? I don't see any query that would need it

petrjasek (approved):

server/planning/events/events_schema.py
MarkLark86: There's this [query](https://github.com/superdesk/superdesk-planning/blob/789a3f7de2300e82a3d52ff9f3abd05d55d24190/server/planning/search/queries/planning.py#L148) that searches Planning items based on IDs of Events with a Primary link. This is needed because, for example, the front-end queries for Planning items that would be affected by an action on an Event.
Without nested mapping we would not be able to search based on a combination of Event ID and link type

petrjasek: missed that one, ok makes sense

PR #1996, opened 2024-06-10
[STT-7] fix(ui): Update UI due to changes with Planning RelatedEvents schema

MarkLark86:


client/interfaces.ts
```diff
@@ -694,14 +694,22 @@ export interface IPlanningCoverageItem {
     scheduled_updates: Array<ICoverageScheduledUpdate>;
 }
 
+export type IPlanningRelatedEventLinkType = 'primary' | 'secondary';
```
tomaskikutis, line 697: could you add a comment describing what primary/secondary means?

client/actions/planning/notifications.ts
```diff
@@ -301,13 +302,20 @@ const udpateAssignment = (planningId) => (
...
+                promiseResponses.push(
+                    dispatch(assignments.api.fetchAssignmentById(cov.assigned_to.assignment_id, true))
+                );
+                promiseResponses.push(
+                    dispatch(assignments.api.fetchAssignmentHistory({_id: cov.assigned_to.assignment_id}))
+                );
             }
         });
+
+        return Promise.all(promiseResponses);
```
tomaskikutis, line 318: `promises` would be a better name. Promise "response" I would understand as a resolved value.

tomaskikutis (approved):

PR #2001, opened 2024-06-19
Fix suggestions

thecalcc:
SDBELGA-815
Lowercase search query match has been adjusted, and also if the input text matches a suggestion (case sensitive), that suggestion disappears and only other ones that don't match the input do appear

https://github.com/superdesk/superdesk-planning/assets/45574768/ad80089b-c7bf-4271-85d3-096fe03dfa6a


tomaskikutis:
Could you update the ticket so the issue is clearly defined? I think it's more correct to always show all matching results, irrespective of case. So if you type `Test` but `Test` and `test` should appear. I don't think any well configured data source should have items that are same, except capitalization.

tomaskikutis (approved):

PR #2002, opened 2024-06-20
Improve desk select

thecalcc:
SDBELGA-830
SDBELGA-817

### DIFF
--- a/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
@@ -86,15 +86,6 @@ interface IState {
     selectedDeskId?: string;
 }
 
-const propertiesToFilterUsersBy = [
-    'username',
-    'display_name',
-    'first_name',
-    'last_name',
-    'email',
-    'sign_off',
-];
-
 export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps, IState> {
     constructor(props) {
         super(props);
@@ -199,25 +190,7 @@ export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps, I
                         >
                             <SelectUser
                                 key={this.state.selectedDeskId}
-                                getItems={(searchString) => {
-                                    const getFilteredUsers = (users: Array<IUser>) => {
-                                        return users.filter((user) =>
-                                            propertiesToFilterUsersBy.some((filterProp) => {
-                                                if (user[filterProp]?.toLowerCase().includes(searchString.toLowerCase())) {
-                                                    return true;
-                                                }
-
-                                                return false;
-                                            }),
-                                        )
-                                    };
-
-                                    if (searchString === '' || searchString == null) {
-                                        return Promise.resolve(this.state.userList);
-                                    }
-
-                                    return Promise.resolve(getFilteredUsers(this.state.userList));
-                                }}
+                                deskId={this.state.selectedDeskId}
                                 onSelect={(user) => {
                                     this.onUserChange(null, user);
                                 }}
--- a/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
@@ -204,7 +204,6 @@ export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps, I
                         <List.Row>
                             <Row>
                                 <Select
-                                    key={this.state.selectedDeskId}
                                     label={gettext('Language:')}
                                     value={language}
                                     onChange={(item) => this.onLanguageChange(item)}
--- a/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
@@ -107,6 +107,7 @@ export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps, I
         const updates: Partial<ICoverageDetails> = {
             desk: newDesk,
             filteredUsers: getUsersForDesk(newDesk, this.props.users),
+            user: null,
         };
 
         if ((this.props.coverage.language ?? '').length < 1) {
@@ -193,6 +194,7 @@ export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps, I
                                 onSelect={(user) => {
                                     this.onUserChange(null, user);
                                 }}
+                                selectedUserId={coverage.user?._id}
                                 autoFocus={false}
                                 horizontalSpacing={true}
                                 clearable={true}
--- a/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
@@ -81,9 +81,28 @@ function getLanguagesForCoverage(
     };
 }
 
-export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps> {
+interface IState {
+    userList: Array<IUser>;
+    selectedDeskId?: string;
+}
+
+const propertiesToFilterUsersBy = [
+    'username',
+    'display_name',
+    'first_name',
+    'last_name',
+    'email',
+    'sign_off',
+];
+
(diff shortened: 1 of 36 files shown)


thecalcc:
Note: Also improved types for some components, and improved/refactored a function

client/utils/index.ts
```diff
@@ -544,8 +544,7 @@ export const getItemsById = (ids, items) => (
 export const getUsersForDesk = (desk, globalUserList = []) => {
     if (!desk) return globalUserList;
 
-    return globalUserList.filter((user) =>
-        map(desk.members, 'user').indexOf(user._id) !== -1);
+    return globalUserList.filter(({_id}) => (desk.members ?? []).find((member) => member.user === _id));
```
tomaskikutis: If there are 20 items in `globalUserList` and 10 `desk.members`, you'll do `20*10=200` iterations (worst case).

If you make a set from `desk.members`, you'll have 20 iterations + 1 to create a set.

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
```diff
@@ -81,28 +81,52 @@ function getLanguagesForCoverage(
     };
 }
 
-export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps> {
+interface IState {
+    userList: Array<IUser>;
+    selectedDeskId?: string;
+}
+
+const propertiesToFilterUsersBy = [
```
tomaskikutis: It's not nice that these fields are being duplicated here.

Looking at this again, planning shouldn't be implementing its own custom search. And having user select supporting to limit to desk users would be a useful feature accross superdesk and extensions. Could you instead adjust UserSelect to add `deskId: string` prop? And then use existing `getUserSearchMongoQuery`? It would be like:

```
const query = {
    $and: [
        {_id: {$in: ['id1', 'id2', 'id3']}},
        getUserSearchMongoQuery('abc'),
    ],
};
```

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
```diff
@@ -81,28 +81,43 @@ function getLanguagesForCoverage(
...
+        const newDesk = deskId == null || deskId == '' ? null : this.props.desks.find((desk) => desk._id === deskId);
 
         const updates: Partial<ICoverageDetails> = {
             desk: newDesk,
             filteredUsers: getUsersForDesk(newDesk, this.props.users),
         };
 
         if ((this.props.coverage.language ?? '').length < 1) {
-            updates.language = newDesk.desk_language;
+            updates.language = newDesk?.desk_language;
```
tomaskikutis: is `updates.language = undefined` intentional? If not, add a condition one line above to make sure it's not null before adding it to updates.

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
```diff
@@ -174,6 +189,8 @@ export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps> {
                             style={{padding: '2rem 0'}}
                         >
                             <SelectUser
+                                key={this.state.selectedDeskId}
```
tomaskikutis: Use `selectedUserId` prop here instead of `key`. Key could be a solution if no proper way was available.

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
```diff
@@ -187,11 +204,12 @@ export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps> {
                         <List.Row>
                             <Row>
                                 <Select
+                                    key={this.state.selectedDeskId}
```
tomaskikutis: Why are you using a key here? value changes and component doesn't respect the prop? It should be fixed inside the select component then.

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
thecalcc: It's intentional

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
thecalcc: No, the idea is that I want the SelectUser to re-render if the selectedDeskId changes. So that the user can't get into an invalid state where he selected a desk and a user from that desk, then he switches to another desk on which the initial user isn't available but still input data is "correct".

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
thecalcc: My bad. If language is set once, we check for that and only if it's not we set a new language value. So the value does update (with the same value) and that's by design. `Select` is working properly.

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
tomaskikutis: could you add a fallback to `null`? Then it would be more obvious it's intentional

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
tomaskikutis: Yep, I get it. Add `selectedDeskId` to state and set to `null` inside `onDeskChange` and you'll achieve it.

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
thecalcc: If it doesn't exist at all what's the reason?

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
thecalcc: No need, it's working as intended now. Besides this if I set it to `null` it counts as an invalid key. I'm using it as a key with `SelectUser`

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
tomaskikutis: The reason is to make it clear to the person reading the code that your intention is to remove a value vs that you forgot to handle the scenario where it's not defined

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
tomaskikutis: I'm not suggesting you to pass `selectedUserId` as a key, but an actual prop of `SelectUser` component.

thecalcc: Got it, well I'm passing it but if it changes the `SelectUser` doesn't update based on it. Should I make the `SelectUser` keep it in its state so it updates. So far nothing else has worked so I can try this

thecalcc: I've pushed a fix to client-core which fixes this issue. Now I don't need to pass a key and `SelectUser` handles updates internally

thecalcc:
Before merging this, merge superdesk-core and update this PR with latest superdesk-core

tomaskikutis (commented):
See 2 unresolved comments

client/components/fields/editor/EventRelatedPlannings/EmbeddedCoverageForm.tsx
```diff
@@ -193,6 +194,7 @@ export class EmbeddedCoverageFormComponent extends React.PureComponent<IProps, I
                                 onSelect={(user) => {
                                     this.onUserChange(null, user);
                                 }}
+                                selectedUserId={coverage.user?._id}
```
tomaskikutis: Fallback to `null` would be helpful - again just to make the intention clear. I thinking there are probably no valid reasons to do optional chaining without providing a fallback except function calls.

tomaskikutis (approved):

PR #2004, opened 2024-06-26
Stt 15

thecalcc:
STT-15


MarkLark86:
@thecalcc There are failing tests on both client & E2E

PR #2005, opened 2024-06-26
merge template and planning subject when creating item

petrjasek:
SDBELGA-821

### DIFF
--- a/server/planning/assignments/assignments_content.py
+++ b/server/planning/assignments/assignments_content.py
@@ -35,14 +35,13 @@ from planning.archive import create_item_from_template
 from planning.signals import assignment_content_create
 
 
-FIELDS_TO_COPY = ("anpa_category", "urgency", "place")
+FIELDS_TO_COPY = ("urgency",)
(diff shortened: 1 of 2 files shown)


server/planning/assignments/assignments_content.py
```diff
@@ -321,3 +323,26 @@ class AssignmentsContentResource(Resource):
...
     privileges = {"POST": "archive"}
+
+
+def merge_subject(item, planning):
+    vocabularies = get_resource_service("vocabularies").get_from_mongo(
+        req=None, lookup={"selection_type": "single selection"}, projection={"_id": 1}
+    )
+    single_value_vocabularies = set([v["_id"] for v in vocabularies])
+    subject = item.setdefault("subject", [])
+    if planning.get("subject"):
```
MarkLark86: Might be good to place this at the top of this function, so it doesn't needlessly lookup vocabs, and return if the Planning item doesn't have a subject

MarkLark86 (approved):

MarkLark86 (approved):

PR #2007, opened 2024-06-28
Respect date filter in event list for multi day events

thecalcc:
SDESK-7310

### DIFF
--- a/client/apps/Planning/PlanningList.tsx
+++ b/client/apps/Planning/PlanningList.tsx
@@ -186,14 +186,17 @@ export class PlanningListComponent extends React.PureComponent<IProps> {
             <React.Fragment>
                 <PlanningListSubNav />
                 <ListPanel
-                    groups={groups.filter((group) => {
-                        const groupDateToMoment = moment(group.date);
+                    groups={(() => {
                         const dateFilter = currentSearch.advancedSearch?.dates?.start;
 
-                        return dateFilter != null
-                            ? groupDateToMoment.isSameOrAfter(dateFilter)
-                            : false;
-                    })}
+                        if (dateFilter != null) {
+                            return groups.filter((group) =>
+                                moment(group.date).isSameOrAfter(dateFilter),
+                            );
+                        }
+
+                        return groups;
+                    })()}
                     onItemClick={openPreview}
                     onDoubleClick={edit}
                     agendas={agendas}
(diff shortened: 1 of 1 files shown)


client/apps/Planning/PlanningList.tsx
```diff
@@ -185,7 +186,14 @@ export class PlanningListComponent extends React.PureComponent<IProps> {
             <React.Fragment>
                 <PlanningListSubNav />
                 <ListPanel
-                    groups={groups}
+                    groups={groups.filter((group) => {
+                        const groupDateToMoment = moment(group.date);
+                        const dateFilter = currentSearch.advancedSearch?.dates?.start;
```
tomaskikutis, line 190: does this view only get rendered in search?

thecalcc: Gets rendered in Combined, Planning only and Event only views

tomaskikutis: Only in search mode? You're filtering out items that don't have some search filters. What if user isn't in search mode?

tomaskikutis (commented):
https://github.com/superdesk/superdesk-planning/pull/2007#discussion_r1660728345

client/apps/Planning/PlanningList.tsx
thecalcc: If user isn't in search mode the filter would be pointless. Good point, I'll add a check beforehand so we don't iterate needlessly, we can pass the items right away

thecalcc:
@tomaskikutis added a comment & pushed changes

client/apps/Planning/PlanningList.tsx
tomaskikutis: image

Needless iteration aside, wouldn't there have been a bug in previous code ^ where if you are not in search mode, there would be no groups visible? Since if `dateFilter` is `null` you were returning `false` from the filter function?

tomaskikutis (approved):

client/apps/Planning/PlanningList.tsx
thecalcc: Nope, date filter has a default of today, so in theory, it's never `null`

tomaskikutis: I see. Still, no filter should mean no filtering should be happening. Which means all items should be returned. Even if all existing code paths return the same result before/after, your latest approach is still better.

There's a good [rule of least surprise](http://principles-wiki.net/principles:principle_of_least_surprise) that I haven't had a chance to mention to you yet that applies here.

PR #2010, opened 2024-07-02
[SDBELGA-816] Search in Planning should recognise CVs case insensitive and not only the name field but also the translations 

devketanpro:


server/planning/events/events_schema.py
```diff
@@ -195,7 +195,31 @@
...
+                "name": string_with_analyzer,
+                "scheme": string_with_analyzer,
+                "translations": {
+                    "type": "object",
+                    "properties": {
+                        "name": {
+                            "type": "object",
+                            "properties": {
+                                "fr": string_with_analyzer,
+                                "nl": string_with_analyzer,
```
petrjasek: this would make it work only for belga, would be better probably to define dynamic mapping here to allow any language to be present

server/planning/events/events_schema.py
```diff
@@ -207,15 +207,8 @@
...
-                    "properties": {
-                        "name": {
-                            "type": "object",
-                            "properties": {
-                                "fr": string_with_analyzer,
-                                "nl": string_with_analyzer,
-                            },
-                        }
-                    },
+                    "dynamic": True,
```
petrjasek: here it should be false probably, we want only name to be searchable

server/planning/events/events_schema.py
```diff
@@ -195,7 +195,24 @@
...
+    "subject": {
+        "type": "list",
+        "mapping": {
+            "type": "nested",
+            "include_in_parent": True,
+            "dynamic": False,
+            "properties": {
+                "qcode": string_with_analyzer,
+                "name": string_with_analyzer,
+                "scheme": string_with_analyzer,
```
petrjasek: scheme shouldn't be analyzed

server/planning/events/events_schema.py
```diff
@@ -195,7 +195,24 @@
...
     # Content metadata
-    "subject": metadata_schema["subject"],
+    "subject": {
+        "type": "list",
+        "mapping": {
+            "type": "nested",
+            "include_in_parent": True,
+            "dynamic": False,
+            "properties": {
+                "qcode": string_with_analyzer,
```
petrjasek: scheme shouldn't be analyzed

server/planning/events/events_schema.py
```diff
@@ -195,7 +195,24 @@
...
+                "name": string_with_analyzer,
```
petrjasek: this would not allow us to create aggregations, so we might need to have it indexed 2x

petrjasek (commented):
pls add test

server/planning/events/events_schema.py
```diff
@@ -195,7 +195,30 @@
         "mapping": {"properties": {"qcode": not_analyzed, "name": not_analyzed}},
     },
     # Content metadata
-    "subject": metadata_schema["subject"],
+    "subject": {
```
MarkLark86: We should also have the same schema for Planning's [subject field](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning.py#L1834)

devketanpro: alright, I will update it, 
Thanks

server/planning/events/events_schema.py
```diff
@@ -195,7 +195,30 @@
...
+        "type": "list",
+        "mapping": {
+            "type": "nested",
```
petrjasek: not sure why we need nested type here?

server/planning/events/events_schema.py
```diff
@@ -195,7 +195,30 @@
...
+            "include_in_parent": True,
+            "dynamic": False,
+            "properties": {
+                "qcode": not_analyzed,
+                "name": {
+                    "type": "string",
```
petrjasek: for BC you can keep this `not_analyzed`

server/planning/events/events_schema.py
```diff
@@ -195,7 +195,30 @@
...
+                    "fields": {
+                        "raw": not_analyzed,
+                        "analyzed": string_with_analyzer,
```
petrjasek: here we only need 1, otherwise this would be indexed 3x

server/planning/planning/planning.py
```diff
@@ -1831,7 +1831,30 @@ def _iter_recurring_plannings_to_update(self, updates, original, update_method):
     "description_text": metadata_schema["description_text"],
     "internal_note": {"type": "string", "nullable": True},
     "anpa_category": metadata_schema["anpa_category"],
-    "subject": metadata_schema["subject"],
+    "subject": {
```
petrjasek, line 1834: would be good to import it from events schema and just copy it here (or vice versa)

devketanpro: done

server/planning/planning/planning.py
```diff
@@ -1831,7 +1831,28 @@ def _iter_recurring_plannings_to_update(self, updates, original, update_method):
...
+        "type": "list",
+        "mapping": {
+            "type": "object",
+            "dynamic": False,
+            "properties": {
+                "qcode": not_analyzed,
+                "name": {
+                    "type": "string",
```
petrjasek: here the type should be `not_analyzed`

devketanpro: like this ?
```.py
"name": {
    "type": "string",
    "index": "not_analyzed",
    "fields": {
        "analyzed": string_with_analyzer,
    },
},
```
if we used not_analyzed then case insensitive is not working

petrjasek: in theory `"type": "keyword"` should be enoug

petrjasek (approved):

PR #2014, opened 2024-07-03
Respect date filter for today

thecalcc:
SDESK-7310

Date filter can never be "null". It either has today's value (by default) or it has a user set value

### DIFF
--- a/client/apps/Planning/PlanningList.tsx
+++ b/client/apps/Planning/PlanningList.tsx
@@ -187,15 +187,12 @@ export class PlanningListComponent extends React.PureComponent<IProps> {
                 <PlanningListSubNav />
                 <ListPanel
                     groups={(() => {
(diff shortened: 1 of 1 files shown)


client/apps/Planning/PlanningList.tsx
```diff
@@ -187,15 +187,12 @@ export class PlanningListComponent extends React.PureComponent<IProps> {
                 <PlanningListSubNav />
                 <ListPanel
                     groups={(() => {
-                        const dateFilter = currentSearch.advancedSearch?.dates?.start;
+                        // Date filter by default has today's value
+                        const dateFilter = currentSearch.advancedSearch?.dates?.start ?? moment().date();
```
tomaskikutis, line 190: > Date filter can never be "null". 

Why then optional chaining is used and default is being applied?

tomaskikutis (approved):

tomaskikutis (approved):

PR #2015, opened 2024-07-03
[STT-15] backend: Support AddAsEvent when Planning already has an Event linked

MarkLark86:


MarkLark86:
@thecalcc Tested this locally. 'Add As Event' now works by creating secondary links if an Event is already linked.
image
You can see the second linked Event is a secondary link, as it does not have the 'locked' red vertical border

thecalcc:
> @thecalcc Tested this locally. 'Add As Event' now works by creating secondary links if an Event is already linked. !You can see the second linked Event is a secondary link, as it does not have the 'locked' red vertical border

Good stuff :)

devketanpro (approved):

PR #2016, opened 2024-07-05
Assignments day filter

thecalcc:
SDBELGA-819

TODO: 
- [x] Check counter of list items
- [x] Check calendar 'Clear' button alignment
- [x] Check filter persistence when switching between Desk/My assignments view with Gideon
- [x] Go over all usages in superdesk-cc [client-core pr](https://github.com/superdesk/superdesk-client-core/pull/4564)
- [x] Implement day filter persistence


thecalcc:
@tomaskikutis ignore the commented out code when doing a review now, as this is something to be discussed with Gideon

client/components/UI/List/Header.tsx
```diff
@@ -1,13 +1,19 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
+interface IProps {
+    title?: string;
+    id?: string;
+    marginTop?: boolean;
+    marginBottom?: boolean;
+    children?: JSX.Element | Array<JSX.Element>;
```
tomaskikutis: wouldn't `children?: React.ReactNode` work?

client/components/Assignments/FiltersBar.tsx
```diff
@@ -3,26 +3,46 @@ import PropTypes from 'prop-types';
...
 
-import {SubNav} from 'superdesk-ui-framework/react';
+import {DatePicker, SubNav} from 'superdesk-ui-framework/react';
 import {StretchBar, Spacer} from '../UI/SubNav';
 import {Checkbox} from '../UI/Form';
 import {OrderFieldInput} from '../OrderBar';
 import {DesksSubnavDropdown} from './DesksSubNavDropDown';
+import {appConfig} from 'superdesk-core/scripts/appConfig';
 
+interface IProps {
```
tomaskikutis, line 10: if you're adding `IProps`, drop prop-types.

client/components/Assignments/FiltersBar.tsx
```diff
@@ -3,26 +3,46 @@ import PropTypes from 'prop-types';
...
+    showAllDeskOption?: boolean;
+    showDeskAssignmentView?: boolean;
+}
+
+const DAYS_AHEAD = 6;
 
 export const FiltersBar = ({
-    filterBy,
-    orderByField,
+    filterBy = 'Desk',
```
tomaskikutis, line 27: double-check that react default props use the same logic to apply a default as default function arguments. That is if value is undefined - a default gets applied. If `null` - then not.

client/components/Assignments/AssignmentGroupList.tsx
```diff
@@ -248,9 +254,9 @@ class AssignmentGroupListComponent extends React.Component {
                             <div className="sd-list-header__number sd-flex-grow">
                                 <span className="a11y-only">{gettext(
                                     'Number of Assignments: ',
-                                    {count: totalCount}
+                                    {count: (filteredAssignments?.length ?? 0)}
```
tomaskikutis, line 297: would be good to update this to use placeholder based translation. It should be possible to pass a react component like `gettext('Number of Assignments: {{n}}', {n: () => <div>5</div>})`

client/components/Assignments/AssignmentGroupList.tsx
```diff
@@ -283,8 +289,8 @@ class AssignmentGroupListComponent extends React.Component {
                         <ListItemLoader />
                     )}
                     {isLoading !== true && (
-                        get(assignments, 'length', 0) > 0 ? (
-                            assignments.map((assignment, index) => this.rowRenderer(index))
+                        (filteredAssignments?.length ?? 0) > 0 ? (
+                            filteredAssignments.map((assignment, index) => this.rowRenderer(index))
```
tomaskikutis: if first argument isn't being used, we should rename it to `_`

client/apps/Assignments/AssignmentList.tsx
```diff
@@ -79,6 +79,7 @@ AssignmentListContainer.defaultProps = {
 };
 
 const mapStateToProps = (state) => ({
+    dayField: selectors.getDayField(state),
```
tomaskikutis: is this used?

client/components/Assignments/AssignmentGroupList.tsx
thecalcc: We can create a ticker for this if you want to update it

client/apps/Assignments/AssignmentList.tsx
thecalcc: nope, removed

client/components/Assignments/FiltersBar.tsx
thecalcc: I don't think this is React specific. It's a function like any other one in javascript. In any case I tested how JS functions work with default arguments, so the same:

image

thecalcc: I've also tested it with this current component. When I pass `undefined` to `orderByField` argument, it falls back. When I pass `null` to it, there's no fallback.

thecalcc:
demo:

https://github.com/superdesk/superdesk-planning/assets/45574768/75b601ca-4b9b-443b-83f2-6b7ad6a4ccd1

client/components/Assignments/FiltersBar.tsx
tomaskikutis: > I don't think this is React specific.

Of course logic on how `defaultProps` work is react specific. `defaultProps` don't exist without react. But all good if it works the same as default values in native JS functions.

client/components/Assignments/AssignmentGroupList.tsx
tomaskikutis: it's 2 min maybe you can update without a ticket? :)

tomaskikutis (approved):

PR #2018, opened 2024-07-08
Handle assignment notification based on user preferences [SDBELGA-818]

devketanpro:


server/planning/assignments/assignments.py
```diff
@@ -344,6 +344,14 @@ def send_assignment_notification(self, updates, original=None, force=False):
         ):
             return
 
+        user = get_user()
```
petrjasek: this is current user again, who imo is not the one who should be receiving the notification

devketanpro: my bad,I forgot to push this file

server/planning/assignments/assignments.py
```diff
@@ -344,15 +344,14 @@ def send_assignment_notification(self, updates, original=None, force=False):
         ):
             return
 
-        user = get_user()
+        assigned_to = updates.get("assigned_to", {})
 
         # No assignment notification sent if user is not enabled that notification
         if not superdesk.get_resource_service("preferences").assignment_notification_is_enabled(
-            user_id=user.get("_id")
+            user_id=assigned_to.get("user")
```
petrjasek: I think it should first check if there is user assigned, it could be only desk afaik

server/planning/assignments/assignments_unlink_test.py
```diff
@@ -5,9 +5,32 @@
 
 
 class AssignmentUnlinkTestCase(TestCase):
+    def setUp(self):
+        super().setUp()
+        with self.app.app_context():
+            users = [
+                {
+                    "_id": ObjectId("5d385f31fe985ec67a0ca583"),
```
petrjasek: would be good to store that as a constant, it's kind of hard to check later all those ids are matching

petrjasek (approved):

PR #2020, opened 2024-07-08
reingest event on subject name change

petrjasek:
SDCP-794 SDCP-795

### DIFF
--- a/server/planning/tests/events_service_test.py
+++ b/server/planning/tests/events_service_test.py
@@ -17,6 +17,11 @@ def test_is_new_version():
 
     assert not service.is_new_version(new_event, old_event)
 
+    new_event["subject"] = [{"qcode": "foo"}, {"qcode": "bar"}]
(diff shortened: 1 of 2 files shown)


server/planning/tests/events_service_test.py
```diff
@@ -22,6 +22,16 @@ def test_is_new_version():
...
+    new_event["subject"] = [{"qcode": "foo", "name": "Foo"}]
+    old_event["subject"] = [{"qcode": "foo", "name": "foo"}]
+
+    assert service.is_new_version(new_event, old_event)
+
+    new_event["subject"] = [{}]
+    old_event["subject"] = [{"qcode": "foo", "name": "foo"}]
+
+    assert service.is_new_version(new_event, old_event)
+
```
devketanpro, line 39: Ideally, we should add a test scenario for subjects in different order.

petrjasek: done

devketanpro (approved):

PR #2024, opened 2024-07-12
fix date filtering

petrjasek:
- always return ongoing events for selected date
- check properly all_day / no_end_time flags

SDESK-7334


MarkLark86 (approved):
Would be good to have tests for these filter conditions

PR #2028, opened 2024-07-16
Assignment email notification registration

thecalcc:
SDBELGA-818

### DIFF
--- a/client/planning-extension/src/extension.ts
+++ b/client/planning-extension/src/extension.ts
@@ -128,6 +128,12 @@ const extension: IExtension = {
                         },
                     },
                 },
+                notifications: {
+                    'assignment:notification': {
+                        type: "email",
+                        label: superdesk.localization.gettext('Send Assignment notifications')
+                    },
+                },
                 globalMenuHorizontal: displayTopbarWidget ? [AssignmentsList] : [],
             },
         };
--- a/client/planning-extension/src/extension.ts
+++ b/client/planning-extension/src/extension.ts
@@ -128,12 +128,7 @@ const extension: IExtension = {
                         },
                     },
                 },
-                notifications: {
-                    'assignment:notification': {
-                        type: "email",
-                        label: superdesk.localization.gettext('Send Assignment notifications')
-                    },
-                },
+                notifications: {'assignment:notification': {type: "email"}},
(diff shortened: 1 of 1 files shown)


thecalcc:
Merge after client-core is updated, and `package-lock.json` is updated in planning

client/planning-extension/src/extension.ts
```diff
@@ -128,6 +128,12 @@ const extension: IExtension = {
                         },
                     },
                 },
+                notifications: {
+                    'assignment:notification': {
+                        type: "email",
+                        label: superdesk.localization.gettext('Send Assignment notifications')
```
tomaskikutis: We should phrase it better. Notification name shouldn't have verbs in it. Also I'm thinking it would be good to have a description where we'd specify under what circumstanced does the notification gets sent.

thecalcc: It's not a name, rather a label. We display that label in settings view

tomaskikutis: It should be a name though. I use the word name and label interchangeably. Extension point may expose the available notifications and their names, but the extension should not be concerned about where will that name be used. Thus we should try to use a universal name like "assignment" and then we could use it in various different cases e.g. 
`gettext('Send {{type}} notifications', {type: 'assignment'})`,
`gettext('Pause {{type}} notifications', {type: 'assignment'})`
etc.

thecalcc: got it, that makes sense considering name and label is different. But for your example, it doesn't make sense in general, because actual label data is coming from the backend - registered for each notification type, so I'd even say it's future proofing to have `name`/`type` since we have all necessary data from the BE.

MarkLark86: `gettext('Send {{type}} notifications', {type: 'assignment'})` would not get a translation for the word `assignment`

tomaskikutis: I am future proofing a bit you could argue, but this is the point with the extensions API. To keep it small and well designed and change it rarely. The label is not a huge thing though, let's go with your version.

You're correct @MarkLark86. Here's the translated placeholder ;)
```
gettext('Send {{type}} notifications', {type: gettext('assignment')})
```

tomaskikutis (approved):
fix CI

PR #2029, opened 2024-07-17
Planning: Adjust and improve the visual indication of the workflow status of Coverages

dzonidoo:
SDESK-7311

### DIFF
--- a/client/utils/planning.ts
+++ b/client/utils/planning.ts
@@ -1224,12 +1224,16 @@ function getCoverageIcon(
     return coverageIcons[type]?.[iconType] ?? iconForUnknownType;
 }
 
-function getCoverageIconColor(item: IPlanningCoverageItem, field = 'state'): string {
+function getCoverageIconColor(item: IPlanningCoverageItem): string | undefined {
     if (item.workflow_status === 'cancelled') {
         return 'var(--sd-colour-state--canceled)';
     }
 
-    switch (getItemWorkflowState(item.assigned_to, field)) {
+    if (item.assigned_to == null) {
+        return undefined;
+    }
+
+    switch (getItemWorkflowState(item.assigned_to)) {
     case ASSIGNMENTS.WORKFLOW_STATE.ASSIGNED:
         return 'var(--sd-colour-state--in-workflow)';
     case ASSIGNMENTS.WORKFLOW_STATE.IN_PROGRESS:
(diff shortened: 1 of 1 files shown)


client/utils/planning.ts
```diff
@@ -1223,15 +1224,24 @@ function getCoverageIcon(
...
 }
 
-function getCoverageIconColor(coverage: IPlanningCoverageItem): string {
-    if (coverage.workflow_status === COVERAGES.WORKFLOW_STATE.ACTIVE) {
-        return 'var(--sd-colour-success)';
-    } else if (get(coverage, 'assigned_to.state') === ASSIGNMENTS.WORKFLOW_STATE.COMPLETED) {
-        return 'var(--sd-colour-success)';
-    } else if (isCoverageDraft(coverage) || get(coverage, 'workflow_status') === COVERAGES.WORKFLOW_STATE.ACTIVE) {
-        return 'var(--sd-colour-highlight)';
+function getCoverageIconColor(item: IPlanningCoverageItem, field = 'state'): string {
```
tomaskikutis: Do not add `field` parameter if it's not used. It makes it appear as multiple options are available and in the end it makes it harder to understand all possible usages.

client/utils/planning.ts
```diff
@@ -1223,15 +1224,24 @@ function getCoverageIcon(
...
+    switch (getItemWorkflowState(item.assigned_to, field)) {
+    case ASSIGNMENTS.WORKFLOW_STATE.ASSIGNED:
+        return 'var(--sd-colour-state--in-workflow)';
+    case ASSIGNMENTS.WORKFLOW_STATE.IN_PROGRESS:
+        return 'var(--sd-colour-state--in-progress)';
+    case ASSIGNMENTS.WORKFLOW_STATE.COMPLETED:
+        return 'var(--sd-colour-state--completed)';
+    }
+
+    if (item.assigned_to.user === null || item.assigned_to.desk === null) {
```
tomaskikutis: Avoid using negatives in naming and flow. For example, it wouldn't be a good name `const notUnassigned= true` because it is difficult to understand. In a similar manner, your `if` condition is whether "unassigned" label should be applied. Instead use the condition to determine if `assigned` should be applied and use "else" for unassigned.

tomaskikutis (commented):
fix CI

tomaskikutis (approved):

PR #2030, opened 2024-07-17
limit displayed days

petrjasek:
based on the interval picker

SDESK-7335

### DIFF
--- a/client/selectors/events.ts
+++ b/client/selectors/events.ts
@@ -2,7 +2,7 @@ import {createSelector} from 'reselect';
 import {get, sortBy} from 'lodash';
 
 import {appConfig} from 'appConfig';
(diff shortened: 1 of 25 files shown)


client/selectors/events.ts
```diff
@@ -47,7 +52,7 @@ export const orderedEvents = createSelector(
             }];
         }
 
-        const dateRange = getSearchDateRange(search, appConfig.start_of_week);
+        const dateRange = getSearchDateRange(search, appConfig.start_of_week, viewInterval);
```
MarkLark86, line 55: Will this be required for [selectors/planning.ts](https://github.com/superdesk/superdesk-planning/blob/develop/client/selectors/planning.ts#L85) as well, so the Planning view is capped to the same date range logic?

petrjasek: yep makes sense to be consistent

MarkLark86 (approved):

PR #2031, opened 2024-07-17
implement expanding related events for planning items

tomaskikutis:
STT-24


client/actions/planning/notifications.ts
```diff
@@ -74,13 +74,9 @@ const onPlanningUpdated = (_e: {}, data: IWebsocketMessageData['PLANNING_UPDATED
         promises.push(dispatch(planning.ui.scheduleRefetch())
             .then((results) => {
                 if (selectors.general.currentWorkspace(getState()) === WORKSPACE.ASSIGNMENTS) {
-                    const selectedItems = selectors.multiSelect.selectedPlannings(getState());
                     const currentPreviewId = selectors.main.previewId(getState());
 
-                    const loadedFromRefetch = selectedItems.indexOf(data.item) !== -1 &&
-                    !get(results, '[0]._items').find((plan) => plan._id === data.item);
-
-                    if (!loadedFromRefetch && currentPreviewId === data.item) {
```
MarkLark86, line 83: I think the purpose of this is to not fetch the item from the API when it was already fetched from `planning.ui.scheduleRefetch()`. Was this not working, or were there scenarios that aren't covered by this?

client/utils/planning.ts
```diff
@@ -1658,6 +1610,40 @@ export function getRelatedEventIdsForPlanning(
     return getRelatedEventLinksForPlanning(plan, linkType).map((event) => event._id);
 }
 
+export function pickRelatedEventsForPlanning(
+    planning: IPlanningItem,
+    _events: Array<IEventItem> | null,
```
MarkLark86: Using a `_` prefix generally means this argument is ignored

client/components/Planning/PlanningItem.tsx
```diff
@@ -271,14 +272,64 @@ class PlanningItemComponent extends React.Component<IProps, IState> {
...
+                                    onClick={(event) => {
+                                        event.stopPropagation();
+
+                                        relatedEventsUI.setVisibility(!relatedEventsUI.visible);
+                                    }}
+                                >
+                                    <Spacer h gap="4" alignItems="center" noWrap>
+                                        <span
+                                            style={{
+                                                paddingBlockStart: 1, // fixing icon alignment
```
MarkLark86, line 306: Should this fix rather go into the ui-framework? @fritzSF

client/components/fields/agendas.tsx
```diff
@@ -1,31 +1,35 @@
...
-            )}
-        >
-            <AgendaNameList agendas={get(fieldsProps, 'agendas.agendas')} />
-        </span>
-    </Fragment>
-);
+interface IProps {
+    fieldsProps: {
+        agendas: {
+            agendas: any;
```
MarkLark86: I believe this might be [IAgenda](https://github.com/superdesk/superdesk-planning/blob/develop/client/interfaces.ts#L372)

client/selectors/events.ts
```diff
@@ -1,6 +1,8 @@
 import {createSelector} from 'reselect';
 import {get, sortBy} from 'lodash';
 
+// PR-TODO: look again at diff of this file
```
MarkLark86: Was this self review completed?

MarkLark86 (approved):
Just a couple small things, overall it looks fine

client/actions/planning/notifications.ts
tomaskikutis: It was never working I think. I brought it up previously in a [slack discussion](https://sourcefabric.slack.com/archives/D01333L81SL/p1720603102083749).

I can't remember 100% if I was running it in the correct context. I did try out again today by commenting out this condition locally:

```if (selectors.general.currentWorkspace(getState()) === WORKSPACE.ASSIGNMENTS)```

and debugged in while in planning items only view.  As I wrote on slack `selectedItems` is an array of objects and `data.item` is a string to it would never evaluate to `true`. The message is coming from a websocket so it should have the same data in any view.

I was furher thinking now maybe I could fix it, but the logic doesn't make sense. There's no multiselect mode in assignments so I don't know how `selectedItems` could ever be not empty.
Also `get(results, '[0]._items')` - none of the items inside the results array have "items" key so I'm not sure how I would test it.

If you know the answers, let me know and I could fix it.

client/components/fields/agendas.tsx
tomaskikutis: You're right. I checked the child component and the properties being accessed match.

client/utils/planning.ts
tomaskikutis: I've changed to `suffix` now. Or you have better ideas?

client/selectors/events.ts
tomaskikutis: I forgot to finalize this one. I was planning to simply add a tag comment here, but I found the selector was being used by component that is itself unused so I was able to drop the selector too.

tomaskikutis:
I'll be off until Monday @MarkLark86 - would be great if you could finalize and merge this PR. Gideon was saying we should prioritize this one.

That websocket event I could address in a separate PR if you advise me. I'm pretty sure it wasn't working for some time so merging it as it shouldn't have a runtime impact.

client/actions/planning/notifications.ts
MarkLark86: We can look at this another time then, this was only meant as a performance helper (to help reduce requests to the server).

PR #2043, opened 2024-07-30
[SDESK-7320] fix 3dot menu disappearing

fritzSF:


client/styles/index.scss
```diff
@@ -128,3 +128,11 @@
         word-break: break-word;
     }
 }
+
```
tomaskikutis, line 131: How is `--space-between-items` relevant? Shouldn't the main `sd-list-item-group` be targeted?

All these classes are defined in ui-framework shouldn't improvements go there as well?

fritzSF: The --space-between-items modifier is specific to the item-group we use in planning (list items have gaps between). I targeted it the same way we already do some other changes. We actually don't need the group part at all, I will remove it. And yes, I will update the framework with the changes as well.

tomaskikutis: > The --space-between-items modifier is specific to the item-group we use in planning (...) I targeted it the same way we already do some other changes

We should build our stuff in a reusable manner and be explicit if something is meant to be reusable or not. If you are making general improvements to `sd-list-item-group` - then it shouldn't matter whether it used in planning or how list group is spaced. It should work everywhere where that component is used.

If you want to change something, but do not want to make it generic - you should not target reusable classes and use a specific one, line `.planning-list-item {overflow: hidden}`

> And yes, I will update the framework with the changes as well.

Same ones? Do we close this PR then?

fritzSF: > We should build our stuff in a reusable manner and be explicit if something is meant to be reusable or not. If you are making general improvements to sd-list-item-group - then it shouldn't matter whether it used in planning or how list group is spaced. It should work everywhere where that component is used.

@tomaskikutis this is not a newly added class. It is currently used in planning but could be used anywhere. The first parent, although not needed, was making it more specific to the current usage. Now that I removed it applies to all nested list items.

fritzSF: > Same ones?

Yes, but my thought was this will be quicker.

tomaskikutis: Can you open a PR on ui-framework side then and close this one?

PR #2045, opened 2024-07-30
Add field classnames, improve code readability, remove dead code

thecalcc:
SDESK-7354

### DIFF
--- a/client/components/Events/EventItem.tsx
+++ b/client/components/Events/EventItem.tsx
@@ -219,8 +219,12 @@ class EventItemComponent extends React.Component<IProps, IState> {
                 >
                     <Row>
                         <span className="sd-overflow-ellipsis sd-list-item--element-grow">
-                            {renderFields(get(listFields, 'event.primary_fields',
-                                EVENTS.LIST.PRIMARY_FIELDS), item, {}, language)}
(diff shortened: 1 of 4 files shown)


client/components/Events/EventItem.tsx
```diff
@@ -219,8 +219,12 @@ class EventItemComponent extends React.Component<IProps, IState> {
                 >
                     <Row>
                         <span className="sd-overflow-ellipsis sd-list-item--element-grow">
-                            {renderFields(get(listFields, 'event.primary_fields',
-                                EVENTS.LIST.PRIMARY_FIELDS), item, {}, language)}
+                            {renderFields(
+                                get(listFields, 'event.primary_fields', EVENTS.LIST.PRIMARY_FIELDS),
```
MarkLark86: Ideally we should be replacing lodash functions with typescript syntax, where applicable.

MarkLark86 (approved):

client/components/Events/EventItem.tsx
thecalcc: Agreed

MarkLark86: This lad is quick!

PR #2046, opened 2024-07-31
fix preferences check

petrjasek:
SDBELGA-818


devketanpro (commented):
I think, now we don't need to register this user preference

https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/assignments/__init__.py#L141-L150

devketanpro (approved):

PR #2047, opened 2024-07-31
fix 'loading more items' indicator and remove cursor pointer (#2040)

dzonidoo:
SDESK-7347


tomaskikutis (commented):
what happens when all items are loaded?

tomaskikutis (approved):

PR #2051, opened 2024-08-05
As a Planning editor I´d like to have the custom CV values sortable [SDBELGA-854]

devketanpro:


client/components/ContentProfiles/FieldTab/FieldEditor.tsx
```diff
@@ -94,6 +94,7 @@ export class FieldEditor extends React.Component<IProps, IState> {
             'schema.expandable': {enabled: fieldType === 'multi_line'},
             'schema.format_options': {enabled: fieldType === 'editor_3'},
             'schema.vocabularies': {enabled: this.props.item.name === 'custom_vocabularies'},
+            'schema.sortable': {enabled: this.props.item.name === 'custom_vocabularies'},
```
MarkLark86: Having this option configurable is not necessary. This requirement was removed from the ticket (having the values sortable does not remove any functionality, nor does it break anything).

Therefor, all we need for this change is to make the custom vocabulary input sortable

MarkLark86 (approved):

PR #2052, opened 2024-08-05
Update data

thecalcc:
SDBELGA-818

### DIFF
--- a/client/planning-extension/src/extension.ts
+++ b/client/planning-extension/src/extension.ts
@@ -128,7 +128,9 @@ const extension: IExtension = {
                         },
                     },
                 },
-                notifications: {'email:notification:assignments': {type: 'email'}},
+                notifications: {
+                    'email:notification:assignments': {name: superdesk.localization.gettext('Assignment')}
+                },
                 globalMenuHorizontal: displayTopbarWidget ? [AssignmentsList] : [],
             },
         };


thecalcc:
Merge [this client-coPR ](https://github.com/superdesk/superdesk-client-core/pull/4589) to develop before merging this one

client/planning-extension/src/extension.ts
```diff
@@ -128,7 +128,9 @@ const extension: IExtension = {
                         },
                     },
                 },
-                notifications: {'email:notification:assignments': {type: 'email'}},
+                notifications: {
+                    'email:notification:assignments': {name: superdesk.localization.gettext('Assignment')}
```
tomaskikutis, line 132: Do desktop notifications work for this?

thecalcc: I don't think there are desktop ones, since there wasn't any handler present before or after

thecalcc: Or in general if there should be desktop ones

tomaskikutis: Check with Petr if it can be made to send desktop notifications too. Because in settings users can toggle both desktop and emai. And if they set desktop to "on" and don't get notified - that's a bug.

thecalcc: It does send email & desktop.

tomaskikutis: It'd be good to update the name then that indicates otherwise.

tomaskikutis (approved):

PR #2055, opened 2024-08-08
Do not send desktop notification if user has toggled off global deskt…

thecalcc:
SDBELGA-818

Sending assignment desktop notifications should not be possible if the user has them toggled off in settings

### DIFF
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -356,7 +356,7 @@ class AssignmentsService(superdesk.Service):
             return
 
         # No assignment desktop notification sent, if a user has disabled their global desktop notifications
-        if assigned_to_user["user_preferences"]["desktop:notification"]["enabled"] is False:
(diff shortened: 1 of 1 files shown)


server/planning/assignments/assignments.py
```diff
@@ -355,6 +355,10 @@ def send_assignment_notification(self, updates, original=None, force=False):
         if assigned_to_user and get_user_notification_preferences(assigned_to_user, "assignments")["email"] is False:
             return
 
+        # No assignment desktop notification sent, if a user has disabled their global desktop notifications
+        if not assigned_to_user.get("user_preferences", {}).get("desktop:notification", {}).get("enabled", True):
```
devketanpro, line 359: If the user has an assignment mail toggle is on and desktop notification is off, so in this case nothing is going to happen, means user has not received any mail.

devketanpro:
here is the PR for this issue:- [https://github.com/superdesk/superdesk-planning/pull/2056](https://github.com/superdesk/superdesk-planning/pull/2056) 

@thecalcc  you can close this one

PR #2065, opened 2024-08-22
Get translated location

thecalcc:
SDBELGA-852

## Front-end checklist


client/components/GeoLookupInput/index.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {AddGeoLookupInput, GeoLookupInputComponent} from './AddGeoLookupInput';
+import {AddGeoLookupInput} from './AddGeoLookupInput';
```
thecalcc, line 2: Type no longer exists, thus removed it

petrjasek (approved):

PR #2070, opened 2024-08-26
Fix closing items from the bottom bar in planning module

thecalcc:
SDESK-7370

## Front-end checklist

### DIFF
--- a/client/actions/main.ts
+++ b/client/actions/main.ts
@@ -1,7 +1,10 @@
 import {get, isEmpty, isEqual, isNil, omit} from 'lodash';
 import moment from 'moment';
 
-import {appConfig} from 'appConfig';
+import {appConfig as config} from 'appConfig';
+
+const appConfig = config as IPlanningConfig;
+
 import {IUser} from 'superdesk-api';
 import {planningApi, superdeskApi} from '../superdeskApi';
 import {
@@ -18,6 +21,8 @@ import {
     ITEM_TYPE,
     IEventTemplate,
     IEventItem,
+    FILTER_TYPE,
+    IPlanningConfig,
 } from '../interfaces';
 
 import {
@@ -68,6 +73,7 @@ import eventsPlanningUi from './eventsPlanning/ui';
 import * as selectors from '../selectors';
 import {validateItem} from '../validators';
 import {searchParamsToOld} from '../utils/search';
+import {searchRawAndStore} from '../api/search';
 
 function openForEdit(item: IEventOrPlanningItem, updateUrl: boolean = true, modal: boolean = false) {
     return (dispatch, getState) => {
@@ -278,22 +284,22 @@ const save = (original, updates, withConfirmation = true) => (
         let confirmation = withConfirmation;
 
         switch (itemType) {
-        case ITEM_TYPE.EVENT:
-            promise = dispatch(eventsUi.save(original, updates, confirmation));
-            confirmation = false;
-            break;
-        case ITEM_TYPE.PLANNING:
-            confirmation = false;
-            promise = dispatch(planningUi.save(original, updates));
-            break;
-        default:
-            promise = Promise.reject(
-                gettext(
-                    'Failed to save, could not find the item {{itemType}}!',
-                    {itemType: itemType}
-                )
-            );
-            break;
+            case ITEM_TYPE.EVENT:
+                promise = dispatch(eventsUi.save(original, updates, confirmation));
+                confirmation = false;
+                break;
+            case ITEM_TYPE.PLANNING:
+                confirmation = false;
+                promise = dispatch(planningUi.save(original, updates));
+                break;
+            default:
+                promise = Promise.reject(
+                    gettext(
+                        'Failed to save, could not find the item {{itemType}}!',
+                        {itemType: itemType}
+                    )
+                );
+                break;
         }
 
         return promise
@@ -329,16 +335,16 @@ const save = (original, updates, withConfirmation = true) => (
                 }
 
                 switch (itemType) {
-                case ITEM_TYPE.EVENT:
-                    dispatch(
-                        eventsApi.receiveEvents([savedItem])
-                    );
-                    break;
-                case ITEM_TYPE.PLANNING:
-                    dispatch(
-                        planningApis.receivePlannings([savedItem])
-                    );
-                    break;
+                    case ITEM_TYPE.EVENT:
+                        dispatch(
+                            eventsApi.receiveEvents([savedItem])
+                        );
+                        break;
+                    case ITEM_TYPE.PLANNING:
+                        dispatch(
+                            planningApis.receivePlannings([savedItem])
+                        );
+                        break;
                 }
 
                 return Promise.resolve(savedItem);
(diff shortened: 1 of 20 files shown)


thecalcc:
Was trying to access `undefined._id` so optional chaining fixes this, allowing the item to visually disappear from the bottom nav. But after a page reload it's still there. Have to dig further and see why we're not receiving the correct items in that function

I also pushed some small time quality of life fixes

MarkLark86:
@thecalcc A better title for this merge request, to describe what is changing and why, will help when generating release changelogs, tracking down changes in commit history etc.

client/components/ItemActionConfirmation/forms/updateRecurringEventsForm.tsx
```diff
@@ -78,7 +78,7 @@ function getRecurringPlanningToUpdate(
     const originalCoverages: IPlanningEmbeddedCoverageMap = (original.planning_ids || [])
         .map((planningId) => plannings[planningId])
         .reduce((planningItems, planningItem) => {
-            planningItems[planningItem._id] = (planningItem.coverages ?? []).reduce(
+            planningItems[planningItem?._id] = (planningItem?.coverages ?? []).reduce(
```
MarkLark86: This indicates that the Planning item is not stored in the Redux store. It might be good to find out why the Event's Planning items were not loaded before opening this form.

thecalcc:
Loading is fixed now. Decided to load the associated plannings once the modal opens so we can have up to date data, and from there it can be used. 

Besides this the actual removal still fails due to the autosave request returning an error. Do you think this might be some data corruption at this point?

client/actions/main.ts
```diff
@@ -718,6 +718,11 @@ const openIgnoreCancelSaveModal = ({
             selectors.planning.storedPlannings(getState());
         const item = get(storedItems, itemId) || {};
 
+        if (itemType === ITEM_TYPE.EVENT) {
+            // Load associated plannings so they can be used later
+            dispatch(eventsApi.loadAssociatedPlannings(item));
```
MarkLark86: The modal would show before this request has finished, and then the modal will be re-drawn once this request finished.

This re-render can be avoided if we wait here for the Planning items to load first, then continue on with displaying the moal

thecalcc:
Not sure what the best approach is here, rather include storing functions as a part of the arguments and then use them inside `searchRawAndStore`, leave it like this, or whether it's worth doing `searchRawAndStore` in general since we can also not abstract storing and leave whoever uses `searchRaw` to handle the storing.

client/actions/main.ts
```diff
@@ -68,6 +73,7 @@ import eventsPlanningUi from './eventsPlanning/ui';
 import * as selectors from '../selectors';
 import {validateItem} from '../validators';
 import {searchParamsToOld} from '../utils/search';
+import {searchRawAndStore} from '../api/search';
```
MarkLark86: Rather than import the function directly, please use the `planningApi` entry point (which is already [imported](https://github.com/superdesk/superdesk-planning/blob/develop/client/actions/main.ts#L6))

Then you can simply use:
`planningApi.search.searchRawAndStore(...`)

client/api/search.ts
```diff
@@ -76,6 +76,22 @@ export function searchRaw<T>(args: ISearchAPIParams): Promise<IRestApiResponse<T
     );
 }
 
+export const searchRawAndStore = <T extends Array<IEventOrPlanningItem>>(args: ISearchAPIParams) => (
+    (dispatch: any): Promise<IRestApiResponse<T>> => superdeskApi.dataApi.queryRawJson<IRestApiResponse<T>>(
```
MarkLark86: Please don't use redux actions in `client/api` module. We want these to be pure python functions in the way they're executed.

Instead you can obtain the `dispatch` function from the API, i.e.
`const {getState, dispatch} = planningApi.redux.store;`

And modify the import
`import {superdeskApi} from '../superdeskApi';`
to be
`import {superdeskApi, planningApi} from '../superdeskApi';`

client/api/search.ts
```diff
@@ -76,6 +76,22 @@ export function searchRaw<T>(args: ISearchAPIParams): Promise<IRestApiResponse<T
...
+        'events_planning_search',
+        excludeNullParams(args)
+    ).then((res) => {
+        if (args.include_associated_planning) {
+            const relatedPlans = res._items.filter((item) => item.type === 'planning');
+
+            dispatch(planningApi.receivePlannings(relatedPlans));
```
MarkLark86, line 92: You could probably just import the `receivePlannings` function directly.
Don't forget to also perform
`dispatch(receiveEvents(events))`, so related events can also be loaded and stored

MarkLark86:
> Besides this the actual removal still fails due to the autosave request returning an error.

@thecalcc Has the autosave issue been fixed? If not, what's the error being returned, and server logs.

thecalcc:
> > Besides this the actual removal still fails due to the autosave request returning an error.
> 
> @thecalcc Has the autosave issue been fixed? If not, what's the error being returned, and server logs.

That's still there. I'm working against bel-uat so here's a curl from that request:

`curl 'https://bel-uat.test.superdesk.org/api/event_autosave/54ab2d58-a2a9-4a9a-91f1-2f10ae392aa7' \
  -H 'authority: bel-uat.test.superdesk.org' \
  -H 'accept: application/json, text/plain, */*' \
  -H 'accept-language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'authorization: Basic YWZhMDI2M2YtNmMyYS00MWY2LTlhZTAtYWE3MzJlNTFlYWE3Og==' \
  -H 'origin: http://localhost:9000' \
  -H 'referer: http://localhost:9000/' \
  -H 'sec-ch-ua: "Not_A Brand";v="8", "Chromium";v="120"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: cross-site' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36' \
  --compressed`
  
  Response: `{
    "_status": "ERR",
    "_error": {
        "code": 404,
        "message": "The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again."
    }
}`

So it looks like that item is in never-land

MarkLark86:
@thecalcc Please fix [tests](https://github.com/superdesk/superdesk-planning/actions/runs/10629603469/job/29466790675?pr=2070#step:6:310).

MarkLark86:
With regards to the autosave, as you pointed out, the workqueue seems to not be getting the correct item when the item is a member of a recurring series of Events. If the workqueue, and associated actions/functions, has the correct item (the one that actually has the lock), then this should fix that bug.

Locks are stored in [redux](https://github.com/superdesk/superdesk-planning/blob/develop/client/reducers/locks.ts), where the appropriate lock logic is used to store the [lock information](https://github.com/superdesk/superdesk-planning/blob/develop/client/interfaces.ts#L354) against the correct item.

client/selectors/locks.ts
```diff
@@ -116,7 +116,7 @@ export const workqueueItems = createSelector<
                 ...newItems.planning,
             ],
             (i) => i.lock_time
-        )
+        ).filter((x) => x.lock_time != null)
```
MarkLark86: I'm still unsure about this fix. It seems that it fixes a symptom and not the cause of the bug. How/when does a lock not have a `lock_time` attribute.

thecalcc: I'm not sure whether we're just getting the wrong data, here's the request that's fetching the locks:

image

```
curl 'https://bel-uat.test.superdesk.org/api/planning_locks' \
  -H 'authority: bel-uat.test.superdesk.org' \
  -H 'accept: */*' \
  -H 'accept-language: en-GB,en-US;q=0.9,en;q=0.8' \
  -H 'authorization: Basic MDBlNmQ2MGQtMGY3MS00YzQ4LWEyYTQtMTM2Zjg1MzRlYzM4Og==' \
  -H 'content-type: application/json' \
  -H 'origin: http://localhost:9000' \
  -H 'referer: http://localhost:9000/' \
  -H 'sec-ch-ua: "Not_A Brand";v="8", "Chromium";v="120"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "macOS"' \
  -H 'sec-fetch-dest: empty' \
  -H 'sec-fetch-mode: cors' \
  -H 'sec-fetch-site: cross-site' \
  -H 'user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36' \
  --compressed
```

I'm doing the filtering later down the line, because there's no actual lock on the item itself. Either data coming from locks request is incorrect or we're missing some afterwards

MarkLark86:
@thecalcc I found the issue, the planning_locks endpoints was using the `recurrence_id` when it should have been using the `item_id` for lock information. Assigning ticket back to you to finalise

MarkLark86 (approved):

PR #2079, opened 2024-09-10
Tooltip not displaying on coverage icon if an event/planning item was created using 'duplicate' option [CPCN-899]

devketanpro:
add coverage assignee details and coverage status on duplicate planning

## Front-end checklist


devketanpro:
Hey @MarkLark86, do we need to add a config for this?

MarkLark86:
> Hey @MarkLark86, do we need to add a config for this?

Yes, and probably default to the current behaviour. Also add it to the config section in the README.md file.

server/planning/planning/planning_duplicate.py
```diff
@@ -99,12 +99,11 @@ def _duplicate_planning(self, original):
...
 
         for cov in new_plan.get("coverages") or []:
-            cov.pop("assigned_to", None)
             cov.get("planning", {}).pop("workflow_status_reason", None)
             cov.pop("scheduled_updates", None)
             cov.get("planning", {})["scheduled"] = new_plan.get("planning_date")
             cov["coverage_id"] = TEMP_ID_PREFIX + "duplicate"
             cov["workflow_status"] = WORKFLOW_STATE.DRAFT
-            cov["news_coverage_status"] = {"qcode": "ncostat:int"}
+            cov["news_coverage_status"] = {"qcode": "ncostat:int", "label": "Planned", "name": "coverage intended"}
```
MarkLark86: Ideally this should come from the CV

MarkLark86 (approved):
There are some tests failing. You might be able to fix those behave tests by adding `@planning_cvs`, which will add the CV required

PR #2087, opened 2024-09-12
Planning workqueue cherrypick

thecalcc:
SDESK-7370
## Front-end checklist


tomaskikutis:
where are you cherry-picking it from?

thecalcc:
> where are you cherry-picking it from?

`develop`. Had to change the base to be `release/2.8`, that's why there were so many commits

tomaskikutis (approved):
I asked because I saw you had set it to be merged TO develop.

PR #2088, opened 2024-09-12
fix related plannings not visible in the editor

petrjasek:
it was only working with primary events, not secondary

STT-72

## Front-end checklist

### DIFF
--- a/server/planning/search/queries/planning.py
+++ b/server/planning/search/queries/planning.py
@@ -150,13 +150,10 @@ def search_by_events(params: Dict[str, Any], query: elastic.ElasticQuery):
 
     if len(event_ids):
         query.must.append(
-            elastic.bool_and(
-                [
-                    elastic.terms(field="related_events._id", values=event_ids),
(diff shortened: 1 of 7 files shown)


server/planning/search/queries/planning.py
```diff
@@ -153,7 +153,7 @@ def search_by_events(params: Dict[str, Any], query: elastic.ElasticQuery):
             elastic.bool_and(
                 [
                     elastic.terms(field="related_events._id", values=event_ids),
-                    elastic.term(field="related_events.link_type", value="primary"),
+                    # elastic.term(field="related_events.link_type", value="primary"),
```
MarkLark86: Commented code! If this is to be removed, do we need to use `elastic.bool_and` still?

MarkLark86 (approved):
@petrjasek My worry with changes like this is the side effects of actions on an Event. For example, when cancelling an Event it should only cancel Planning items that have a primary link to the actioned Event.

These changes seem fine, but we need to make sure secondary links are not affected by actions on an Event.

PR #2089, opened 2024-09-16
Update with new superdesk content profile API

thecalcc:
SDCP-838

## Front-end checklist


client/controllers/AddToPlanningController.tsx
```diff
@@ -20,6 +21,24 @@ const DEFAULT_PLANNING_SCHEMA = {
 };
 
 export class AddToPlanningController {
+    $scope: any;
+    notify: typeof superdeskApi['ui']['notify'];
```
tomaskikutis: It's not correct to specify this type. You should not make assumptions on what superdesk API is using under the hood as implementation. It should be possible to easily change it without breaking anything.

tomaskikutis (commented):
Go over the checklist and mark the ones that apply. It might not be possible to implement some, in that case leave unchecked.

tomaskikutis:
Regarding checklist, if it doesn't apply to the PR (i.e. no lodash usages), mark as done. I'd like to quickly see whether there are any guidelines that do apply, but were not implemented.

thecalcc:
> Regarding checklist, if it doesn't apply to the PR (i.e. no lodash usages), mark as done. I'd like to quickly see whether there are any guidelines that do apply, but were not implemented.

image

I went over the checklist, and I checked all the boxes that are done by me, thus apply to the PR

image

tomaskikutis:
> In this pull request, properties of redux state are not being passed as props to components; instead, we connect it to the component that needs them. Except components where using a react key is required - do not connect those due to performance reasons.

Why didn't you check this one for example?

client/controllers/AddToPlanningController.tsx
```diff
@@ -197,20 +217,14 @@ export class AddToPlanningController {
...
-                            getErrorMessage(error, this.gettext('Failed to load content profile.'))
-                        );
-                        this.$scope.resolve(error);
-                        return Promise.reject(error);
-                    })
-            ), (error) => {
+            .then((newsItem: IArticle) => {
+                const contentProfile = superdeskApi.entities.contentProfile.get(newsItem.profile);
+
+                return Promise.resolve({
```
tomaskikutis: no need for `Promise.resolve` here. You should return the object directly. Since it's inside a `.then` it's a promise already

tomaskikutis (approved):

PR #2092, opened 2024-09-17
fix(ui): AddToPlanning action fails due to change in SuperdeskAPI

thecalcc:
SDCP-838


tomaskikutis (approved):

MarkLark86:
@thecalcc Please use meaningful PR titles, as it describes the change when we make a release and auto-generate [changelog](https://github.com/superdesk/superdesk-planning/releases/tag/v2.6.3) from commit titles.

PR #2102, opened 2024-10-02
Fix tree select item rendering

thecalcc:
SDCP-844

## Front-end checklist

### DIFF
--- a/client/components/fields/editor/CustomVocabularies.tsx
+++ b/client/components/fields/editor/CustomVocabularies.tsx
@@ -1,12 +1,13 @@
 import * as React from 'react';
 import {connect} from 'react-redux';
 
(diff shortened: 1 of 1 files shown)


thecalcc:
@tomaskikutis for some reason the values are not saving correctly, even though they're getting set properly in the UI. Can you take a look as well and see if you find anything? What I don't get is how is `const itemFieldName = cv.schema_field ?? 'subject';` safe in terms of saving the values to the proper field. Most of the time I saw that `cv.schema_field` was empty

tomaskikutis (approved):

tomaskikutis:
doesn't activate save button

thecalcc:
> doesn't activate save button

fixed in the latest commit

tomaskikutis (approved):

PR #2105, opened 2024-10-03
Nested event displays redundant planning icon

dzonidoo:
STT-87

## Front-end checklist

### DIFF
--- a/client/interfaces.ts
+++ b/client/interfaces.ts
@@ -926,7 +926,7 @@ export interface IBaseListItemProps<T> {
 export interface IEventListItemProps extends IBaseListItemProps<IEventItem> {
     calendars: Array<ICalendar>;
     filterLanguage?: string;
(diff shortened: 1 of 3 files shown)


client/interfaces.ts
```diff
@@ -924,10 +924,13 @@ export interface IBaseListItemProps<T> {
 }
 
 export interface IEventListItemProps extends IBaseListItemProps<IEventItem> {
-    relatedPlanningText?: string;
     calendars: Array<ICalendar>;
     filterLanguage?: string;
-    toggleRelatedPlanning?(event: React.MouseEvent): void;
+    planningItemsLength?: number;
```
tomaskikutis: can it be made mandatory?

Also improve naming to be `relatedPlanningsCount`

tomaskikutis (approved):

PR #2107, opened 2024-10-03
automatically show related plannings added to event

petrjasek:
atm it's a workaround, but eventually we will need to get rid of `planning_ids` on event which is computed on load but without updating event itself it gets cached.

STT-86

## Front-end checklist

### DIFF
--- a/client/api/events.ts
+++ b/client/api/events.ts
@@ -71,9 +71,9 @@ export function searchEventsGetAll(params: ISearchParams): Promise<Array<IEventI
     });
 }
 
-export function getEventById(eventId: IEventItem['_id'], params: IGetRequestParams): Promise<IEventItem> {
+export function getEventById(eventId: IEventItem['_id'], params?: IGetRequestParams): Promise<IEventItem> {
     return superdeskApi.dataApi
-        .findOne<IEventItem>('events', eventId + (params?.cache === false ? `?time=${Math.floor(Date.now() / 1000)}` : ''))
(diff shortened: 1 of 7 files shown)


client/api/events.ts
```diff
@@ -70,9 +71,9 @@ export function searchEventsGetAll(params: ISearchParams): Promise<Array<IEventI
     });
 }
 
-export function getEventById(eventId: IEventItem['_id']): Promise<IEventItem> {
+export function getEventById(eventId: IEventItem['_id'], params: IGetRequestParams): Promise<IEventItem> {
     return superdeskApi.dataApi
-        .findOne<IEventItem>('events', eventId)
+        .findOne<IEventItem>('events', eventId + (params?.cache === false ? `?time=${Math.floor(Date.now() / 1000)}` : ''))
```
MarkLark86: Aren't we using the `Cache-Control: no-cache` header in our responses? Meaning every hop along the network right through to the browser should abide by this response header value.

Is this not the case?

petrjasek: we do, but that doesn't [disable caching actually](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#no-cache), it makes a conditional request, and then it gets 304 from server due to last updated time / etag not changed

MarkLark86: Ah, so it's Eve that's doing that. Okay

MarkLark86 (approved):

PR #2110, opened 2024-10-10
Adding related items to plannings and events

tomaskikutis:
STT-67
STT-8

## Front-end checklist


tomaskikutis:
@MarkLark86 If an existing planning item gets drag and dropped as related to an event, I'm adding it using the secondary link. Should it be possible to add coverages to planning items linked using the secondary link type?

client/actions/events/api.ts
```diff
@@ -547,6 +549,58 @@ const uploadFiles = (event) => (
     }
 );
 
+function getLinkedPlanningItems(eventId: string): Promise<IRestApiResponse<IPlanningItem>> {
```
MarkLark86: I'm thinking this function might be useful in other places, so it would be good to have this available under `IPlanningAPI.events.getLinkedPlanningItems` (to be implemented in [client/api/events.ts](https://github.com/superdesk/superdesk-planning/blob/develop/client/api/events.ts#L217))

MarkLark86: Would it be good to just return `Array<IPlanningItem>` instead of `IRestApiResponse<IPlanningItem>`? The function name states `get me the list of Planning items`, or do you see the need for attributes under the `IRestApiResponse` as well?

client/actions/events/api.ts
```diff
@@ -547,6 +549,58 @@ const uploadFiles = (event) => (
...
+     * missing items will be linked, extra items unlinked
+     */
+    planningIds: Array<IPlanningItem['_id']>,
+):Promise<void> {
+    return getLinkedPlanningItems(eventId).then((res) => {
+        // note: planning items to which `eventId` is linked
+        const currentlyLinked: Array<IPlanningItem> = res._items;
+
+        const currentLinkedIds = new Set(currentlyLinked.map((item) => item._id));
+        const toLink: Array<IPlanningItem['_id']> = planningIds.filter((id) => currentLinkedIds.has(id) !== true);
```
MarkLark86: [Set.has](https://tc39.es/ecma262/multipage/keyed-collections.html#sec-set.prototype.has) and [Array.includes](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.includes) already return a boolean value, what is the use case for explicitly checking for `!== true`?

client/components/Events/EventMetadata/index.tsx
```diff
@@ -92,19 +96,51 @@ class EventMetadataComponent extends React.PureComponent<IProps> {
...
+            const showRemoveButton = onRemoveEvent != null;
+
+            if (showEditButton !== true && showRemoveButton !== true) {
+                return null;
+            } else {
+                return (
+                    <Spacer h gap="4">
+                        {
+                            showEditButton && (
+                                <button
```
MarkLark86: Could it be possible to use a button from the UI-Framework instead?

client/components/fields/editor/AssociatedEvent.tsx
```diff
@@ -15,28 +16,121 @@ interface IProps extends IEditorFieldProps {
...
+    }
+
+    private addRelatedEvent(id: IEventItem['_id']) {
+        this.props.onChange(
+            this.props.field,
+            [
+                ...this.getCurrentValue(),
+                {
+                    _id: id,
+                    link_type: 'secondary',
```
MarkLark86: The `link_type` here could be different values, depending on the instance config.

We have a `PLANNING_EVENT_LINK_METHOD` config available in settings.py. This can be exposed to front-end by adding something like the following in [planning/__init__.py](https://github.com/superdesk/superdesk-planning/blob/feature/multiple-events-in-planning/server/planning/__init__.py#L241)

```python
app.client_config["planning"]["event_link_method"] = get_planning_event_link_method()
```

And adding `event_link_method` to [IPlanningConfig](https://github.com/superdesk/superdesk-planning/blob/develop/client/interfaces.ts#L273) and in [client/config.ts](https://github.com/superdesk/superdesk-planning/blob/develop/client/config.ts).

Then based on that config value:

**one_primary**
* Only 1 Event can be linked, therefor if an Event is already linked this action should not be available, and an error thrown here
* Otherwise the `link_type` would be `primary`

**many_secondary**
* `link_type` is always `secondary`

**one_primary_many_secondary**
* `link_type` is `primary` if there are no Event's currently linked
* `link_type` is `secondary` otherwise

client/utils/events.ts
```diff
@@ -461,14 +464,73 @@ function canMarkEventAsComplete(
     }
 }
 
+/**
+ * Will return true if there is at least one event that can be added
+ */
+export function canAddSomeEventsAsRelatedToPlanningEditor(eventIds: Array<IEventItem['_id']>): boolean {
+    const editor = planningApi.editor(EDITOR_TYPE.INLINE);
```
MarkLark86: This is where we would use that `event_link_method` config to add additional clause if this action should be available (i.e. config is `one_primary` and currentRelatedEvents is not empty, then return false)

client/utils/events.ts
```diff
@@ -461,14 +464,73 @@ function canMarkEventAsComplete(
...
+    const isEventAlreadyAddedAsRelated: (eventId: string) => boolean =
+        (eventId: string) => currentRelatedEvents.find((item) => item._id === eventId) != null;
+
+    const nextRelatedEvents: Array<IPlanningRelatedEventLink> = [
+        ...currentRelatedEvents,
+        ...eventIds
+            .filter((eventId) => isEventAlreadyAddedAsRelated(eventId) !== true)
+            .map((eventId): IPlanningRelatedEventLink => ({
+                _id: eventId,
+                link_type: 'secondary',
```
MarkLark86: And same here, `link_type` would depend on the config and currently linked Events (if any)

MarkLark86 (commented):
Looks good, mostly just suggestion about implementing logic around the link config in the front-end

MarkLark86:
> @MarkLark86 If an existing planning item gets drag and dropped as related to an event, I'm adding it using the secondary link. Should it be possible to add coverages to planning items linked using the secondary link type?

I might need a little more context here, as I'm not sure I understand the scenario entirely. When adding a Coverage to a Planning item, there is no logic involved around any linked Events that alllows or denies adding these Coverages. Am I missing something here, could you maybe explain more please

client/actions/events/api.ts
tomaskikutis: I find it quicker to understand code when negation comes after the condition, not before.

tomaskikutis:
FYI I've force-pushed to revert a bad merge. I had by mistake merged develop to this branch, instead of `feature/multiple-events-in-planning`

tomaskikutis:
> I might need a little more context here, as I'm not sure I understand the scenario entirely. When adding a Coverage to a Planning item, there is no logic involved around any linked Events that alllows or denies adding these Coverages. Am I missing something here, could you maybe explain more please

That's what I wanted to know in the end - whether coverages have some relation to the planning item and its link type. [Because in the UI](https://github.com/user-attachments/assets/9a4c8016-162a-49ed-a350-3ef7f06fe0ca) they look attached to planning items. Since you say they're not connected - all good then - I don't have to do anything special to address it.

client/actions/events/api.ts
```diff
@@ -547,6 +549,89 @@ const uploadFiles = (event) => (
...
+
+        const toUnlink: Array<IPlanningItem> = currentlyLinked
+            .filter((item) => {
+                const createdAt = new Date(item._created);
+                const now = new Date();
+
+                const ageSeconds = (now.getTime() - createdAt.getTime()) / 1000;
+                const tooRecent = ageSeconds > 30;
+
+                if (tooRecent) {
```
tomaskikutis, line 575: @MarkLark86 any idea how to avoid this hack?

MarkLark86: Only thing that comes to mind is if you also pass in the original associatedPlanning IDs. Check non-temp IDs (aka not "fake ID") that are in original that are not in updates are to be unlinked.

tomaskikutis:
@MarkLark86 besides the hack, anything else you'd like me to address here?

client/actions/events/api.ts
MarkLark86: Fair enough

client/actions/events/api.ts
```diff
@@ -574,9 +660,22 @@ const save = (original, updates) => (
...
+            const createOrUpdatePromise: Promise<Array<IEventItem>> = originalEvent?._id != null ?
                 planningApi.events.update(originalItem, eventUpdates) :
                 planningApi.events.create(eventUpdates);
+
+            return createOrUpdatePromise.then(([updatedEvent]: Array<IEventItem>) => {
+                if (updates.associated_plannings == null) {
+                    return Promise.resolve([updatedEvent]);
+                }
+
+                // return Promise.resolve([updatedEvent]);
```
MarkLark86: Remove commented code

client/globals.d.ts
```diff
@@ -147,3 +147,44 @@ declare var ResizeObserverSize: {
 interface ResizeObserverCallback {
     (entries: ResizeObserverEntry[], observer: ResizeObserver): void;
 }
+
+// KEEP IN SYNC WITH client/planning-extension/src/globals.d.ts
+declare module 'superdesk-api' {
+    interface ISuperdeskGlobalConfig {
```
MarkLark86, line 158: Nice one :+1: Removes the need for the hack `const appConfig = config.appConfig as IPlanningConfig;`.
Though, where is it getting the references to types from client/interfaces.ts (such as IPlanningItem)?

server/planning/__init__.py
```diff
@@ -242,6 +242,9 @@ def init_app(app):
         "default_create_planning_series_with_event_series"
     ] = get_config_default_create_planning_series_with_event_series(app)
 
+    # TODO: use get_planning_event_link_method
+    app.client_config["planning_event_link_method"] = "one_primary_many_secondary"
```
MarkLark86: TODO action needs to be fixed

MarkLark86 (approved):
@tomaskikutis Looks good to me.

server/planning/__init__.py
tomaskikutis: @petrjasek can you fix this one for me?

client/globals.d.ts
tomaskikutis: Good catch. I forgot to import it so it wasn't getting references and evaluating to `any`

tomaskikutis:
@petrjasek could you take a look why `CI-Server / server-behave` started crashing?

MarkLark86:
> @petrjasek could you take a look why `CI-Server / server-behave` started crashing?

@tomaskikutis Ketan fixed this in a [PR](https://github.com/superdesk/superdesk-planning/pull/2138) for develop branch.

tomaskikutis:
@petrjasek could you merge develop to `feature/multiple-events-in-planning`? There are back-end conflicts there

petrjasek:
@MarkLark86 can you have a look? seems a bit complicated for me

PR #2113, opened 2024-10-14
Fix: Planning items created via an event are not always scheduled, but are being kept in draft mode [SDBELGA-879]

devketanpro:
when we remove the required field from the editor, in the schema its still marked as required, so when we validate the related_item, it gives us validation error as it is required

## Front-end checklist

### DIFF
--- a/client/components/ContentProfiles/ContentProfileModal.tsx
+++ b/client/components/ContentProfiles/ContentProfileModal.tsx
@@ -392,6 +392,11 @@ class ContentProfileModalComponent extends React.Component<IProps, IState> {
             fields.forEach((item, index) => {
                 profile.editor[item.name] = {...item.field};
                 profile.editor[item.name].index = index;
+
+                // If the field is disabled in the editor, mark it as not required to avoid backend validation issues.
+                if (item.field?.enabled == false) {
+                    profile.schema[item.name].required = false;
+                }
             });
 
             return key === 'profile' ?
--- a/client/components/ContentProfiles/ContentProfileModal.tsx
+++ b/client/components/ContentProfiles/ContentProfileModal.tsx
@@ -392,11 +392,6 @@ class ContentProfileModalComponent extends React.Component<IProps, IState> {
             fields.forEach((item, index) => {
                 profile.editor[item.name] = {...item.field};
                 profile.editor[item.name].index = index;
-
-                // If the field is disabled in the editor, mark it as not required to avoid backend validation issues.
-                if (item.field?.enabled == false) {
-                    profile.schema[item.name].required = false;
-                }
             });
 
             return key === 'profile' ?
--- a/server/features/planning_validate.feature
+++ b/server/features/planning_validate.feature
@@ -30,6 +30,11 @@ Feature: Planning Validate
             }
         }, {
             "_id": "planning", "name": "planning",
+            "editor":{
+                "place": {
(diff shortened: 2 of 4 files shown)


client/components/ContentProfiles/ContentProfileModal.tsx
```diff
@@ -392,6 +392,11 @@ class ContentProfileModalComponent extends React.Component<IProps, IState> {
             fields.forEach((item, index) => {
                 profile.editor[item.name] = {...item.field};
                 profile.editor[item.name].index = index;
+
+                // If the field is disabled in the editor, mark it as not required to avoid backend validation issues.
+                if (item.field?.enabled == false) {
```
MarkLark86: This would be better off being fixed on the back-end where the validation occurs, as that is where the issue is. If a field is disabled, it shouldn't matter what other attributes the field has

MarkLark86:
Is there a ticket associated with this change, or did you notice this one yourself?
A ticket would be good so that QA can verify the fix as well

devketanpro:
Gideon and I discovered this issue while testing ticket [SDBELGA-879](https://sofab.atlassian.net/browse/SDBELGA-879). For your reference, you can check the thread here: [Slack Thread](https://sourcefabric.slack.com/archives/C1XBKE70C/p1728400933594089).

[SDBELGA-879]: https://sofab.atlassian.net/browse/SDBELGA-879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

client/components/ContentProfiles/ContentProfileModal.tsx
devketanpro: the issue is that if a field is disabled in the editor but still marked as `required: true` in the content profile schema, it triggers a validation error when validating related planning. You can see the validation logic here: [planning_post.py#L107-L114](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning_post.py#L107-L114).

This causes errors for fields that are not enabled in the editor, which seems incorrect. As I checked, we update the content_profiles from the frontend, which is why I implemented the changes to here handle this.

MarkLark86: The current workaround is to:
1. Open Event and/or Planning profile(s)
2. For every field that is disabled, but marked required:
2.1. add the field to the profile
2.2. disable the required attribute
2.3. remove the field from the profile
3. Save the profile

With the solution in this PR, an administrator would then have to:
1. Open the Event and/or Planning profile(s)
2. For every field that is disabled, but marked required:
2.1. add the field to the profile
2.2. remove the field from the profile
3. Save the profile

This action from an admin is still required as the DB would still hold the `required: True` attribute.

Considering the bug is caused on the back-end where validation occurs, it would be best to fix the bug there, _**without requiring further action from an admin**_.

Fixing the bug in the [planning_validator](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/validate/planning_validate.py#L130-L139) service would be the way to go. Exclude any field that is disabled.

devketanpro: got your point, Thanks Mark

MarkLark86 (approved):

PR #2114, opened 2024-10-17
Nested single day events always shown only with time

dzonidoo:
STT-93

## Front-end checklist

### DIFF
--- a/client/components/Events/EventDateTime.tsx
+++ b/client/components/Events/EventDateTime.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import {superdeskApi} from '../../superdeskApi';
-import {IEventItem} from '../../interfaces';
+import {IEventItem, IPlanningListItemProps} from '../../interfaces';
 
 import {eventUtils, timeUtils} from '../../utils';
 
@@ -13,6 +13,7 @@ interface IProps {
   item: IEventItem;
   ignoreAllDay?: boolean;
   displayLocalTimezone?: boolean;
+  planningItem?: IPlanningListItemProps;
 }
 
 export class EventDateTime extends React.PureComponent<IProps> {
@@ -23,6 +24,7 @@ export class EventDateTime extends React.PureComponent<IProps> {
         const end = eventUtils.getEndDate(item);
         const isAllDay = eventUtils.isEventAllDay(start, end);
         const multiDay = !eventUtils.isEventSameDay(start, end);
+        const showEventStartDate = eventUtils.showEventStartDate(start, this.props.planningItem?.date);
         const isRemoteTimeZone = timeUtils.isEventInDifferentTimeZone(item);
         const withYear = multiDay && start.year() !== end.year();
         const localStart = timeUtils.getLocalDate(start, item.dates.tz);
@@ -79,7 +81,7 @@ export class EventDateTime extends React.PureComponent<IProps> {
                     </span>
                 )}
                 <DateTime
-                    withDate={multiDay}
+                    withDate={(showEventStartDate || multiDay)}
                     withYear={withYear}
                     date={start}
                     {...commonProps}
--- a/client/components/Events/EventDateTime.tsx
+++ b/client/components/Events/EventDateTime.tsx
@@ -13,7 +13,7 @@ interface IProps {
   item: IEventItem;
   ignoreAllDay?: boolean;
   displayLocalTimezone?: boolean;
-  planningItem?: IPlanningListItemProps;
+  planningProps?: IPlanningListItemProps;
 }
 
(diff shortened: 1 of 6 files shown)


client/components/Events/EventDateTime.tsx
```diff
@@ -79,7 +81,7 @@ export class EventDateTime extends React.PureComponent<IProps> {
                     </span>
                 )}
                 <DateTime
-                    withDate={multiDay}
+                    withDate={(showEventStartDate || multiDay)}
```
thecalcc, line 84: use `??` instead

client/components/Events/EventDateTime.tsx
```diff
@@ -23,6 +24,7 @@ export class EventDateTime extends React.PureComponent<IProps> {
         const end = eventUtils.getEndDate(item);
         const isAllDay = eventUtils.isEventAllDay(start, end);
         const multiDay = !eventUtils.isEventSameDay(start, end);
+        const showEventStartDate = eventUtils.showEventStartDate(start, this.props.planningItem?.date);
```
thecalcc: So if `planningItem.date` is `null`, this will continue working?

client/components/Events/EventDateTimeColumn.tsx
```diff
@@ -23,7 +24,7 @@ export class EventDateTimeColumn extends React.PureComponent<IProps> {
             return (
                 <Column border={false} className="flex-justify--start sd-padding-t--1">
                     <Row classes="sd-margin--0">
-                        <EventDateTime item={this.props.item} />
+                        <EventDateTime item={this.props.item} planningItem={this.props.planningItem} />
```
thecalcc: Why is this component called `EventDateTime` and we're passing an event item and a planningItem? If it's meant to show both planning & event item times you should rename it accordingly

client/utils/events.ts
```diff
@@ -80,6 +80,10 @@ function isEventSameDay(startingDate: IDateTime, endingDate: IDateTime): boolean
     return moment(startingDate).format('DD/MM/YYYY') === moment(endingDate).format('DD/MM/YYYY');
 }
 
+function showEventStartDate(eventDate: IDateTime, planningDate: IDateTime): boolean {
```
thecalcc: Since up in the code you're passing a possibly `undefined` planningDate it only makes sense to make this optional here, and handle the case where `planningDate` is undefined explicitly

thecalcc:
@tomaskikutis you can take a look as well if you want, but I feel confident in the review and that this is a simple feature

client/components/Events/EventDateTimeColumn.tsx
dzonidoo: The component only renders the date and time of the event, but conditionally, when the event's date differs from the planning item's date.

client/components/Events/EventDateTime.tsx
dzonidoo: withDate={(showEventStartDate || multiDay)}
its not the same like
withDate={showEventStartDate ?? multiDay}

client/components/Events/EventDateTime.tsx
dzonidoo: yep

thecalcc (approved):

client/components/Events/EventDateTime.tsx
tomaskikutis: Move `multiDay` check to `showEventStartDate` function

tomaskikutis:
I'll take a look

client/utils/events.ts
```diff
@@ -80,6 +80,13 @@ function isEventSameDay(startingDate: IDateTime, endingDate: IDateTime): boolean
     return moment(startingDate).format('DD/MM/YYYY') === moment(endingDate).format('DD/MM/YYYY');
 }
 
+function showEventStartDate(eventDate: IDateTime, planningDate?: IDateTime): boolean {
+    if (planningDate == null) {
+        return true;
+    }
+    return moment(eventDate).format('DD/MM/YYYY') != moment(planningDate).format('DD/MM/YYYY');
```
tomaskikutis: it'd be cleaner to use `isSame` method from moment. We don't care about date format here

client/components/Planning/PlanningItemWithEvents.tsx
```diff
@@ -69,6 +69,7 @@ export class PlanningItemWithEvents extends React.Component<IProps, IState> {
                                                     {...this.props.getEventProps(event)}
                                                     multiSelectDisabled
                                                     key={event._id}
+                                                    planningItem={planningProps}
```
tomaskikutis: `planningProps` would be a more consistent naming for a prop

tomaskikutis (approved):

PR #2116, opened 2024-10-22
Missing 'assignment available' circle on SD left panel

dzonidoo:
SDCP-826

## Front-end checklist

### DIFF
--- a/client/planning-extension/src/assignments-overview/hiddenAssignmentsList.tsx
+++ b/client/planning-extension/src/assignments-overview/hiddenAssignmentsList.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
-import {ISuperdeskQuery, IUser} from 'superdesk-api';
-import {IAssignmentItem} from '../../../interfaces';
+import {IUser} from 'superdesk-api';
+import {getAssignmentsQuery} from '.';
+import {ASSIGNMENT_STATE, IAssignmentItem} from '../../../interfaces';
 import {superdesk} from '../superdesk';
 const {getLiveQueryHOC} = superdesk.components;
 const LiveAssignmentsHOC = getLiveQueryHOC<IAssignmentItem>();
@@ -8,22 +9,6 @@ const LiveAssignmentsHOC = getLiveQueryHOC<IAssignmentItem>();
 interface IState {
     loading: false;
     currentUser: IUser;
-}
-
-function getAssignmentsQuery(userId: IUser['_id']): ISuperdeskQuery {
-    const query: ISuperdeskQuery = {
-        filter: {
-            $and: [
-                {'assigned_to.user': {$eq: userId}},
-                {'assigned_to.state': {$in: ['assigned']}},
-            ],
-        },
-        sort: [{'planning.scheduled': 'asc'}],
-        page: 1,
-        max_results: 100,
-    };
-
-    return query;
 }
 
 export class HiddenAssignmentsList extends React.PureComponent<{}, {loading: true} | IState> {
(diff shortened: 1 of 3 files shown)


thecalcc:
@tomaskikutis from what Nikola told me, you don't agree with extending the `AssignmentsList` component, by adding one more prop e.g. `hidden`. Why? I think it's much better to have a single source of truth in this case with a very minor config option, rather than duplicating code, it looks a bit needless to me if this is the option we go for. If we won't bother to make a different one, like dispatching some event or something else, than it looks reasonable to me, to simply extend it.

client/planning-extension/src/extension.ts
```diff
@@ -251,7 +252,7 @@ const extension: IExtension = {
                         },
                     },
                 ],
-                globalMenuHorizontal: displayTopbarWidget ? [AssignmentsList] : [],
+                globalMenuHorizontal: displayTopbarWidget ? [AssignmentsList] : [HiddenAssignmentsList],
```
tomaskikutis: render `HiddenAssignmentsList` unconditionally

client/planning-extension/src/assignments-overview/hiddenAssignmentsList.tsx
```diff
@@ -0,0 +1,141 @@
...
+                                {menuId: 'MENU_ITEM_PLANNING_ASSIGNMENTS', badgeValue: itemsCount.toString()},
+                            );
+
+                            const resources: ILiveResourcesProps['resources'] = [
+                                {resource: 'desks', ids: Object.keys(grouped)},
+                                {resource: 'vocabularies', ids: ['g2_content_type']},
+                            ];
+
+                            return (
+                                <WithLiveResources resources={resources}>
```
tomaskikutis: why are you rendering `WithLiveResources` and all other child components if they will never be seen? Simply return `null` instead of hiding it with CSS.

client/planning-extension/src/assignments-overview/hiddenAssignmentsList.tsx
```diff
@@ -0,0 +1,141 @@
...
+            <div style={{display: 'none'}}>
+                <LiveAssignmentsHOC resource="assignments" query={getAssignmentsQuery(currentUser._id)}>
+                    {
+                        (data) => {
+                            const assignments = data._items;
+                            const itemsCount = data._meta.total;
+                            const grouped = groupBy(assignments, (item) => item.assigned_to.desk);
+
+
+                            superdesk.dispatchEvent(
```
tomaskikutis: drop this from `client/planning-extension/src/assignments-overview/index.tsx` since this component would be rendered unconditionally

client/planning-extension/src/assignments-overview/hiddenAssignmentsList.tsx
dzonidoo: i cant return null bcz of interface... LiveAssignmentsHOC component expect to return another component
ill return empty div then...

tomaskikutis: yep, div is fine too

dzonidoo: but still i need wrapeer with display none :D

dzonidoo: but still i need wrapper with display none :D

dzonidoo: but still i need wrapeer with display none :D

tomaskikutis: image

adjust the interface so children return not JSX.Element, but React.ReactNode. it will allow null then

tomaskikutis (approved):
Almost all good now. Do a rename `HiddenAssignmentsList` > `AssignmentsCountTracker` or something similar. Because there's no list in reality - not even the hidden one ;D Update the naming and merge this.

Another thing regarding the checklist in the template - if it doesn't apply mark as done. For example if there are no lodash usages, mark as "lodash usages are taken care of". As a person reviewing I want to see what from the checklist you decided not to do and double check if it is a good approach to violate the rule. If the rule doesn't apply, mark as done so I don't need to think about it.

PR #2121, opened 2024-10-31
Show reassign on auto add to workflow, types cleanup

thecalcc:
SDCP-863

## Front-end checklist

### DIFF
--- a/client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
+++ b/client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
@@ -25,7 +25,7 @@ interface IProps {
     addNewsItemToPlanning?: IArticle;
     onChange(field: string, value: any): void;
     onFocus?(): void;
-    onRemoveAssignment?(): void;
+    onRemoveAssignment?(): Promise<void>;
     setCoverageDefaultDesk(coverage: IPlanningCoverageItem | ICoverageScheduledUpdate): void;
     showEditCoverageAssignmentModal(props: {
         field: string;
@@ -76,14 +76,14 @@ export class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
         } = this.props;
 
         const userAssigned = getCreator(value, 'assigned_to.user', users);
-        const deskAssigned = getItemInArrayById(desks, value.assigned_to.desk);
-        const coverageProvider = value.assigned_to.coverage_provider;
-        const assignmentState = value.assigned_to.state;
+        const deskAssigned = getItemInArrayById(desks, value.assigned_to?.desk);
+        const coverageProvider = value.assigned_to?.coverage_provider;
+        const assignmentState = value.assigned_to?.state;
         const cancelled = value.workflow_status === ASSIGNMENTS.WORKFLOW_STATE.CANCELLED;
         const canEditAssignment = planningUtils.isCoverageDraft(value) ||
             (!!addNewsItemToPlanning && !value.coverage_id && !get(value, 'scheduled_update_id'));
-        const shouldShowRemove = (onRemoveAssignment != null
-            && planningConfig.planning_auto_assign_to_workflow === false);
+
+        const autoAddToWf = true;
 
         if (!deskAssigned && (!userAssigned || !coverageProvider)) {
             return (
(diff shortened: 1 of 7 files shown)


MarkLark86:
@thecalcc There seems to be some E2E failing:
image

thecalcc:
E2E to be fixed, handle unsaved coverage assignment removal

client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
```diff
@@ -168,7 +175,7 @@ export class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
                         </ListRow>
                     )}
                 </Column>
-                {canEditAssignment && !readOnly && (
+                {(canEditAssignment || planningConfig.planning_auto_assign_to_workflow) && !readOnly && (
```
MarkLark86: We have conditions set above where remove assignment action is not available (such as when in the "Add To Planning" modal in Monitoring).

Rather than using the auto assign config, why not remove the condition for `isCoverageDraft`, then it will work under the conditions outlined in the ticket and include other current business rules.

client/components/UI/Button.tsx
```diff
@@ -1,41 +1,60 @@
...
 import {KEYCODES} from './constants';
 import {onEventCapture} from './utils';
 
 
-/**
- * @ngdoc react
- * @name Button
- * @description Generic Button component
- */
+interface IButtonProps {
```
MarkLark86, line 8: Just as a note, we should replace these with the ones from the UI Framework

MarkLark86 (commented):
There is 1 more condition that was not in the A.C. for the ticket.

When the user is in the Assignments page, the actions are not available on the Assignment if it is locked.

We should include these same checks here. If the Assignment is not in draft state (aka added to workflow), then we should check the lock state of the Assignment to see if the action can be performed.

client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
thecalcc: Perfect, if that's how we judge whether the instance has `"auto_add_to_wf"`, and all other rules are followed then I'll do it

client/components/UI/Button.tsx
thecalcc: I can do that for this ticket if you mean replacing the button

thecalcc: For buttons related to this assignment

MarkLark86: If it's easy to replace these remove/reassign buttons to UI Framework ones, then it might be beneficial here (while you're editing this component anyways).
That way, PR by PR we can slowly replace these old ones with the new ones, rather than 1 big task to remove old code/components

client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
```diff
@@ -173,27 +187,28 @@ export class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
...
-                                    tabIndex={0}
-                                    enterKeyIsClick
-                                    disabled={!!addNewsItemToPlanning}
-                                    autoFocus
-                                />
-                            </ListRow>
-                        )}
+                        {onRemoveAssignment != null
+                            && !isAssignmentLocked
+                            && !planningUtils.isCoverageDraft(value)
```
MarkLark86: Why not show the remove action if the coverage is draft?

client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
```diff
@@ -173,27 +187,28 @@ export class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
...
-                                    onClick={onRemoveAssignment}
-                                    tabIndex={0}
-                                    enterKeyIsClick
-                                    disabled={!!addNewsItemToPlanning}
-                                    autoFocus
-                                />
-                            </ListRow>
-                        )}
+                        {onRemoveAssignment != null
+                            && !isAssignmentLocked
```
MarkLark86: We can probably add this condition for the re-assign as well. If the Assignment is locked, the user should not be able to re-assign it to someone else

MarkLark86 (approved):

PR #2122, opened 2024-11-05
Nested single day events always shown only with time

dzonidoo:
STT-93

## Front-end checklist

### DIFF
--- a/client/utils/events.ts
+++ b/client/utils/events.ts
@@ -57,7 +57,7 @@ import {
     sanitizeItemFields,
 } from './index';
 import {toUIFrameworkInterface, getRelatedEventIdsForPlanning} from './planning';
-import {isSameDay} from 'helpers';
+import {isSameDay} from './../helpers';
 
 
 /**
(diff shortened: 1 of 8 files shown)


client/utils/events.ts
```diff
@@ -1361,7 +1361,7 @@ const self = {
     canUpdateEventTime,
     canConvertToRecurringEvent,
     canUpdateEventRepetitions,
-    isEventSameDay,
+    isSameDay,
```
tomaskikutis: move it out of event utils since it's generic

dzonidoo: and put it in some helpers?

tomaskikutis: yep

client/validators/events.ts
```diff
@@ -7,6 +7,7 @@ import {gettext, eventUtils} from '../utils';
 import * as selectors from '../selectors';
 import {formProfile} from './profile';
 import {PRIVILEGES, EVENTS, TO_BE_CONFIRMED_FIELD} from '../constants';
+import {isSameDay} from 'helpers';
```
tomaskikutis: use relative imports (starting with a dot)

tomaskikutis (approved):

PR #2126, opened 2024-11-18
[STTNHUB-376] Add config to not cancel Planning when Event is cancelled

MarkLark86:
### Purpose
Add new option to the Event Content Profile's Related Plannings field, that allows to turn off the feature that cancels/unposts Planning item(s) when unposting an Event.

### What has changed
* New option `Cancel planning items with Event` for the `Related Plannings` Event field
* Allow's editing a Planning item that is associated with a Cancelled Event

Resolves: [STTNHUB-376](https://sofab.atlassian.net/browse/STTNHUB-376)

## Front-end checklist

[STTNHUB-376]: https://sofab.atlassian.net/browse/STTNHUB-376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


devketanpro (approved):
looks good !

PR #2131, opened 2024-11-20
[SDESK-7441] Create basic async resource models and services for `planning.events`

eos87:
### Purpose
Implement the basic async resource and service for Planning Events app

### What has changed
- Added resources models and services for Events, Planning, Assignments and Published
- Fixed pytests so they work with async 
- Partially fixed behave tests (`~82%` of them now green)
- Updated GitHub actions to run with Python `v3.10` and latest actions

> [!NOTE]
> This PR depends on https://github.com/superdesk/superdesk-core/pull/2758. ~Once that PR is merged, I will update the requirements here to point to superdesk-core `async` branch instead.~

Thanks for checking!

[SDESK-7441]: https://sofab.atlassian.net/browse/SDESK-7441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/assignments/module.py
+++ b/server/planning/assignments/module.py
@@ -6,12 +6,12 @@ from superdesk.core.resources import (
 )
 
 from planning.types import AssignmentResourceModel
-from .service import AssingmentsAsyncService
+from .service import AssignmentsAsyncService
 
 assignments_resource_config = ResourceConfig(
     name="assignments",
     data_class=AssignmentResourceModel,
-    service=AssingmentsAsyncService,
+    service=AssignmentsAsyncService,
     etag_ignore_fields=["planning", "published_state", "published_at"],
     mongo=MongoResourceConfig(
         indexes=[
--- a/server/planning/assignments/module.py
+++ b/server/planning/assignments/module.py
@@ -12,11 +12,24 @@ assignments_resource_config = ResourceConfig(
     name="assignments",
     data_class=AssignmentResourceModel,
     service=AssingmentsAsyncService,
+    etag_ignore_fields=["planning", "published_state", "published_at"],
     mongo=MongoResourceConfig(
         indexes=[
-            MongoIndexOptions(name="coverage_item_1", keys=[("coverage_item", 1)]),
-            MongoIndexOptions(name="planning_item_1", keys=[("planning_item", 1)]),
-            MongoIndexOptions(name="published_state_1", keys=[("published_state", 1)]),
+            MongoIndexOptions(
+                name="coverage_item_1",
+                keys=[("coverage_item", 1)],
+                unique=False,
+            ),
+            MongoIndexOptions(
+                name="planning_item_1",
+                keys=[("planning_item", 1)],
+                unique=False,
+            ),
+            MongoIndexOptions(
+                name="published_state_1",
+                keys=[("published_state", 1)],
+                unique=False,
+            ),
         ],
     ),
     elastic=ElasticResourceConfig(),
--- a/server/planning/assignments/module.py
+++ b/server/planning/assignments/module.py
@@ -1,0 +1,23 @@
+from superdesk.core.resources import (
+    ResourceConfig,
+    MongoIndexOptions,
+    MongoResourceConfig,
+    ElasticResourceConfig,
+)
+
+from planning.types import AssignmentResourceModel
+from .service import AssingmentsAsyncService
+
+assignments_resource_config = ResourceConfig(
+    name="assignments",
+    data_class=AssignmentResourceModel,
+    service=AssingmentsAsyncService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="coverage_item_1", keys=[("coverage_item", 1)]),
+            MongoIndexOptions(name="planning_item_1", keys=[("planning_item", 1)]),
+            MongoIndexOptions(name="published_state_1", keys=[("published_state", 1)]),
+        ],
+    ),
+    elastic=ElasticResourceConfig(),
+)
--- a/server/planning/assignments/service.py
+++ b/server/planning/assignments/service.py
@@ -1,5 +1,5 @@
 from planning.core.service import BasePlanningAsyncService
 
 
-class AssingmentsAsyncService(BasePlanningAsyncService):
+class AssignmentsAsyncService(BasePlanningAsyncService):
     pass
--- a/server/planning/assignments/service.py
+++ b/server/planning/assignments/service.py
@@ -2,4 +2,4 @@ from planning.core.service import BasePlanningAsyncService
 
 
 class AssingmentsAsyncService(BasePlanningAsyncService):
-    resource_name = "assignments"
+    pass
--- a/server/planning/assignments/service.py
+++ b/server/planning/assignments/service.py
@@ -1,0 +1,5 @@
+from planning.core.service import BasePlanningAsyncService
+
+
+class AssingmentsAsyncService(BasePlanningAsyncService):
+    resource_name = "assignments"
--- a/server/planning/events/module.py
+++ b/server/planning/events/module.py
@@ -5,8 +5,8 @@ from superdesk.core.resources import (
     ElasticResourceConfig,
 )
 
-from planning.events import EventsAsyncService
 from planning.types import EventResourceModel
+from .service import EventsAsyncService
 
 events_resource_config = ResourceConfig(
     name="events",
--- a/server/planning/events/module.py
+++ b/server/planning/events/module.py
@@ -1,0 +1,26 @@
+from superdesk.core.resources import (
+    ResourceConfig,
+    MongoIndexOptions,
+    MongoResourceConfig,
+    ElasticResourceConfig,
+)
+
+from planning.events import EventsAsyncService
+from planning.types import EventResourceModel
+
+events_resource_config = ResourceConfig(
+    name="events",
+    data_class=EventResourceModel,
+    service=EventsAsyncService,
+    default_sort=[("dates.start", 1)],
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="recurrence_id_1", keys=[("recurrence_id", 1)]),
+            MongoIndexOptions(name="state", keys=[("state", 1)]),
+            MongoIndexOptions(name="dates_start_1", keys=[("dates.start", 1)]),
+            MongoIndexOptions(name="dates_end_1", keys=[("dates.end", 1)]),
+            MongoIndexOptions(name="template", keys=[("template", 1)]),
+        ],
+    ),
+    elastic=ElasticResourceConfig(),
+)
--- a/server/planning/events/module.py
+++ b/server/planning/events/module.py
@@ -15,7 +15,11 @@ events_resource_config = ResourceConfig(
     default_sort=[("dates.start", 1)],
     mongo=MongoResourceConfig(
         indexes=[
-            MongoIndexOptions(name="recurrence_id_1", keys=[("recurrence_id", 1)]),
+            MongoIndexOptions(
+                name="recurrence_id_1",
+                keys=[("recurrence_id", 1)],
+                unique=False,
+            ),
             MongoIndexOptions(name="state", keys=[("state", 1)]),
             MongoIndexOptions(name="dates_start_1", keys=[("dates.start", 1)]),
             MongoIndexOptions(name="dates_end_1", keys=[("dates.end", 1)]),
--- a/server/planning/planning/module.py
+++ b/server/planning/planning/module.py
@@ -1,0 +1,23 @@
+from superdesk.core.resources import (
+    ResourceConfig,
+    MongoIndexOptions,
+    MongoResourceConfig,
+    ElasticResourceConfig,
+)
+
+from planning.types import PlanningResourceModel
+
+from .service import PlanningAsyncService
+
+planning_resource_config = ResourceConfig(
+    name="planning",
+    data_class=PlanningResourceModel,
+    service=PlanningAsyncService,
+    default_sort=[("dates.start", 1)],
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="planning_recurrence_id", keys=[("planning_recurrence_id", 1)]),
+        ],
+    ),
+    elastic=ElasticResourceConfig(),
+)
--- a/server/planning/planning/module.py
+++ b/server/planning/planning/module.py
@@ -13,7 +13,6 @@ planning_resource_config = ResourceConfig(
     name="planning",
     data_class=PlanningResourceModel,
     service=PlanningAsyncService,
-    default_sort=[("dates.start", 1)],
     mongo=MongoResourceConfig(
         indexes=[
             MongoIndexOptions(name="planning_recurrence_id", keys=[("planning_recurrence_id", 1)]),
--- a/server/planning/planning/module.py
+++ b/server/planning/planning/module.py
@@ -15,7 +15,11 @@ planning_resource_config = ResourceConfig(
     service=PlanningAsyncService,
     mongo=MongoResourceConfig(
         indexes=[
-            MongoIndexOptions(name="planning_recurrence_id", keys=[("planning_recurrence_id", 1)]),
+            MongoIndexOptions(
+                name="planning_recurrence_id",
+                keys=[("planning_recurrence_id", 1)],
+                unique=False,
+            ),
         ],
     ),
     elastic=ElasticResourceConfig(),
--- a/server/planning/published/__init__.py
+++ b/server/planning/published/__init__.py
@@ -1,0 +1,22 @@
+from superdesk.core.resources import (
+    ResourceConfig,
+    MongoIndexOptions,
+    MongoResourceConfig,
+    ElasticResourceConfig,
+)
+
+from planning.types import PublishedPlanningModel
+
+from .service import PublishedAsyncService
+
+published_resource_config = ResourceConfig(
+    name="published_planning",
+    data_class=PublishedPlanningModel,
+    service=PublishedAsyncService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="item_id_1_version_1", keys=[("item_id", 1), ("version", 1)]),
+        ],
+    ),
+    elastic=ElasticResourceConfig(),
+)
--- a/server/planning/published/__init__.py
+++ b/server/planning/published/__init__.py
@@ -15,7 +15,11 @@ published_resource_config = ResourceConfig(
     service=PublishedAsyncService,
     mongo=MongoResourceConfig(
         indexes=[
-            MongoIndexOptions(name="item_id_1_version_1", keys=[("item_id", 1), ("version", 1)]),
+            MongoIndexOptions(
+                name="item_id_1_version_1",
+                keys=[("item_id", 1), ("version", 1)],
+                unique=False,
+            ),
         ],
     ),
     elastic=ElasticResourceConfig(),
--- a/server/planning/tests/__init__.py
+++ b/server/planning/tests/__init__.py
@@ -1,5 +1,4 @@
-from superdesk.tests import TestCase as _TestCase, update_config, setup
-from superdesk.factory.app import get_app
+from superdesk.tests import TestCase as _TestCase, update_config
 
 
 class TestCase(_TestCase):
@@ -8,6 +7,4 @@ class TestCase(_TestCase):
     def setUp(self):
         config = {"INSTALLED_APPS": ["planning"]}
         update_config(config)
-        # self.app = get_app(config)
-        # setup.app = self.app
         super().setUp()
--- a/server/planning/tests/__init__.py
+++ b/server/planning/tests/__init__.py
@@ -8,6 +8,6 @@ class TestCase(_TestCase):
     def setUp(self):
         config = {"INSTALLED_APPS": ["planning"]}
         update_config(config)
-        self.app = get_app(config)
-        setup.app = self.app
+        # self.app = get_app(config)
+        # setup.app = self.app
         super().setUp()
--- a/server/planning/types/assignment.py
+++ b/server/planning/types/assignment.py
@@ -8,7 +8,7 @@ from superdesk.core.resources.validators import validate_data_relation_async
 
 from .base import BasePlanningModel
 from .common import PlanningCoverage
-from .enums import AssignmentWorkflowState
+from .enums import AssignmentPublishedState, AssignmentWorkflowState
 
 
 @dataclass
@@ -39,7 +39,7 @@ class AssignmentResourceModel(BasePlanningModel):
     item_type: Annotated[fields.Keyword, Field(alias="type")] = "assignment"
     priority: int | None = None
     coverage_item: fields.Keyword | None = None
-    planning_item: Annotated[str, validate_data_relation_async("planning")] | None = None
+    planning_item: Annotated[str, validate_data_relation_async("planning")]
     scheduled_update_id: fields.Keyword | None = None
 
     lock_user: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
@@ -54,3 +54,6 @@ class AssignmentResourceModel(BasePlanningModel):
     description_text: str | None = None
     accepted: bool = False
     to_delete: bool = Field(default=False, alias="_to_delete")
+
+    published_at: datetime | None = None
+    published_state: AssignmentPublishedState | None = None
--- a/server/planning/types/assignment.py
+++ b/server/planning/types/assignment.py
@@ -7,7 +7,7 @@ from superdesk.core.resources import fields, dataclass
 from superdesk.core.resources.validators import validate_data_relation_async
 
 from .base import BasePlanningModel
-from .common import PlanningCoverage
+from .common import LockFieldsMixin, PlanningCoverage
 from .enums import AssignmentPublishedState, AssignmentWorkflowState
 
 
@@ -32,28 +32,29 @@ class AssignedTo:
     coverage_provider: CoverageProvider | None = None
 
 
-class AssignmentResourceModel(BasePlanningModel):
+class AssignmentResourceModel(BasePlanningModel, LockFieldsMixin):
+    id: Annotated[fields.ObjectId, Field(alias="_id", default_factory=fields.ObjectId)]
+
     firstcreated: datetime = Field(default_factory=utcnow)
     versioncreated: datetime = Field(default_factory=utcnow)
 
-    item_type: Annotated[fields.Keyword, Field(alias="type")] = "assignment"
     priority: int | None = None
     coverage_item: fields.Keyword | None = None
-    planning_item: Annotated[str, validate_data_relation_async("planning")]
+    planning_item: Annotated[fields.Keyword, validate_data_relation_async("planning")]
     scheduled_update_id: fields.Keyword | None = None
-
-    lock_user: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
-    lock_time: datetime | None = None
-    lock_session: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
-    lock_action: fields.Keyword | None = None
 
     assigned_to: AssignedTo | None = None
     planning: PlanningCoverage | None = None
 
     name: str | None = None
-    description_text: str | None = None
+    description_text: fields.HTML | None = None
     accepted: bool = False
     to_delete: bool = Field(default=False, alias="_to_delete")
 
     published_at: datetime | None = None
     published_state: AssignmentPublishedState | None = None
+
+    # TODO-ASYNC: this field was in the original schema but we're not sure if it's really required
+    # also it would clash with the computed property `type` from ResourceModel
+    # leaving it here for now until we know if it is required or we can get rid of it
+    # item_type: Annotated[fields.Keyword, Field(alias="type")] = "assignment"
--- a/server/planning/types/assignment.py
+++ b/server/planning/types/assignment.py
@@ -1,0 +1,56 @@
+from pydantic import Field
+from typing import Annotated
+from datetime import datetime
+
+from superdesk.utc import utcnow
+from superdesk.core.resources import fields, dataclass
+from superdesk.core.resources.validators import validate_data_relation_async
+
+from .base import BasePlanningModel
+from .common import PlanningCoverage
+from .enums import AssignmentWorkflowState
+
+
+@dataclass
+class CoverageProvider:
+    qcode: fields.Keyword | None = None
+    name: fields.Keyword | None = None
+    contact_type: fields.Keyword | None = None
+
+
+@dataclass
+class AssignedTo:
+    desk: fields.Keyword | None = None
+    user: fields.Keyword | None = None
+    contact: fields.Keyword | None = None
+    assignor_desk: fields.Keyword | None = None
+    assignor_user: fields.Keyword | None = None
+    assigned_date_desk: datetime | None = None
+    assigned_date_user: datetime | None = None
+    state: AssignmentWorkflowState | None = None
+    revert_state: AssignmentWorkflowState | None = None
+    coverage_provider: CoverageProvider | None = None
+
+
+class AssignmentResourceModel(BasePlanningModel):
+    firstcreated: datetime = Field(default_factory=utcnow)
+    versioncreated: datetime = Field(default_factory=utcnow)
+
+    item_type: Annotated[fields.Keyword, Field(alias="type")] = "assignment"
+    priority: int | None = None
+    coverage_item: fields.Keyword | None = None
+    planning_item: Annotated[str, validate_data_relation_async("planning")] | None = None
+    scheduled_update_id: fields.Keyword | None = None
+
+    lock_user: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
+    lock_time: datetime | None = None
+    lock_session: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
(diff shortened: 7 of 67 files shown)


MarkLark86 (commented):
@eos87 I've had a quick look over it, looks fine so far. Will re-review once you've completed the schema changes etc

server/planning/assignments/module.py
```diff
@@ -0,0 +1,23 @@
...
+from planning.types import AssignmentResourceModel
+from .service import AssingmentsAsyncService
+
+assignments_resource_config = ResourceConfig(
+    name="assignments",
+    data_class=AssignmentResourceModel,
+    service=AssingmentsAsyncService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="coverage_item_1", keys=[("coverage_item", 1)]),
```
MarkLark86: `MongoIndexOptions.unique` defaults to `True`, which won't work here (especially for `planning_item` and `published_state`). I'm wondering if we should just change the default to False instead (might require changes on the Newshub side if we do).

So we have 2 choices here:
1. Modify the config options in this PR to add `unique=False` where unique is not already set to `True`, and be mindful of this default
2. Change the default to `False` and update Newshub code to reflect this (for the ones that require `unique` to be set to `True`)

server/planning/assignments/module.py
```diff
@@ -0,0 +1,23 @@
...
+    ElasticResourceConfig,
+)
+
+from planning.types import AssignmentResourceModel
+from .service import AssingmentsAsyncService
+
+assignments_resource_config = ResourceConfig(
+    name="assignments",
+    data_class=AssignmentResourceModel,
+    service=AssingmentsAsyncService,
```
MarkLark86: Will also need to set the `etag_ignore_fields` config, original one uses `["planning", "published_state", "published_at"]`

server/planning/assignments/service.py
```diff
@@ -0,0 +1,5 @@
+from planning.core.service import BasePlanningAsyncService
+
+
+class AssingmentsAsyncService(BasePlanningAsyncService):
+    resource_name = "assignments"
```
MarkLark86: Any reason why we're explicitly assigning the `resource_name` here? This is populated when the service is registered and defined on the ResourceConfig

server/planning/events/module.py
```diff
@@ -0,0 +1,26 @@
...
+from .service import EventsAsyncService
+
+events_resource_config = ResourceConfig(
+    name="events",
+    data_class=EventResourceModel,
+    service=EventsAsyncService,
+    default_sort=[("dates.start", 1)],
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="recurrence_id_1", keys=[("recurrence_id", 1)]),
```
MarkLark86: Same with these `MongoIndexOptions` and the `unique` default

server/planning/planning/module.py
```diff
@@ -0,0 +1,22 @@
...
+
+from .service import PlanningAsyncService
+
+planning_resource_config = ResourceConfig(
+    name="planning",
+    data_class=PlanningResourceModel,
+    service=PlanningAsyncService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="planning_recurrence_id", keys=[("planning_recurrence_id", 1)]),
```
MarkLark86: And again with the `unique` default value

server/planning/published/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+from .service import PublishedAsyncService
+
+published_resource_config = ResourceConfig(
+    name="published_planning",
+    data_class=PublishedPlanningModel,
+    service=PublishedAsyncService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="item_id_1_version_1", keys=[("item_id", 1), ("version", 1)]),
```
MarkLark86: `version` might be incremented, but the original does't state `unique`. So to keep with the original, might need `unique` to be `False` here too

server/planning/tests/__init__.py
```diff
@@ -8,6 +8,6 @@ class TestCase(_TestCase):
     def setUp(self):
         config = {"INSTALLED_APPS": ["planning"]}
         update_config(config)
-        self.app = get_app(config)
-        setup.app = self.app
+        # self.app = get_app(config)
```
MarkLark86: Commented code not required

server/planning/types/assignment.py
```diff
@@ -0,0 +1,56 @@
...
+    assignor_desk: fields.Keyword | None = None
+    assignor_user: fields.Keyword | None = None
+    assigned_date_desk: datetime | None = None
+    assigned_date_user: datetime | None = None
+    state: AssignmentWorkflowState | None = None
+    revert_state: AssignmentWorkflowState | None = None
+    coverage_provider: CoverageProvider | None = None
+
+
+class AssignmentResourceModel(BasePlanningModel):
```
MarkLark86: This one actually uses ObjectId as the ID type, so would need to either:
1. Inherit from `superdesk.core.resources.ResourceModelWithObjectId`
2. re-implement `id` field with `id: Annotated[ObjectId, Field(alias="_id", default_factory=ObjectId)]` (with `ObjectId` being `superdesk.core.resources.fields.ObjectId`

server/planning/types/assignment.py
```diff
@@ -0,0 +1,56 @@
...
+    firstcreated: datetime = Field(default_factory=utcnow)
+    versioncreated: datetime = Field(default_factory=utcnow)
+
+    item_type: Annotated[fields.Keyword, Field(alias="type")] = "assignment"
```
MarkLark86: I know this is on the original resource (actually I don't understand why, this resource only stores assignments), but doesn't this clash with the computed field `type` from `ResourceModel`?

server/planning/types/assignment.py
```diff
@@ -0,0 +1,56 @@
...
+    priority: int | None = None
+    coverage_item: fields.Keyword | None = None
+    planning_item: Annotated[str, validate_data_relation_async("planning")] | None = None
```
MarkLark86: I think this should be a `fields.Keyword`, as the original is `not_analyzed`

server/planning/types/assignment.py
```diff
@@ -0,0 +1,56 @@
...
+    scheduled_update_id: fields.Keyword | None = None
+
+    lock_user: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
```
MarkLark86: Wondering if could use a mixin class to provide lock information, as it's used in couple different resources.

i.e.
```python

class LockFieldsMixin:
    lock_user: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
    lock_time: datetime | None = None
    lock_session: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
    lock_action: fields.Keyword | None = None

class AssignmentResourceModel(BasePlanningModel, LockFieldsMixin):
    ...
```

server/planning/types/assignment.py
```diff
@@ -0,0 +1,56 @@
...
+from superdesk.utc import utcnow
+from superdesk.core.resources import fields, dataclass
+from superdesk.core.resources.validators import validate_data_relation_async
+
+from .base import BasePlanningModel
+from .common import PlanningCoverage
+from .enums import AssignmentWorkflowState
+
+
+@dataclass
```
MarkLark86, line 14: I have made changes to how we should use dataclasses, to provide similar interface to normal BaseModel and allow unknown fields as well.

i.e.
```python
from superdesk.core.resources import dataclass, Dataclass

@dataclass
class CoverageProvider(Dataclass):
    qcode: fields.Keyword | None = None
    name: fields.Keyword | None = None
    contact_type: fields.Keyword | None = None
```

By inheriting from `superdesk.core.resources.Dataclass` adds certain extra methods, such as `to_dict` and `from_dict`.

server/planning/types/enums.py
```diff
@@ -0,0 +1,56 @@
...
+    KILLED = "killed"
+    CANCELLED = "cancelled"
+    RESCHEDULED = "rescheduled"
+    POSTPONED = "postponed"
+    SPIKED = "spiked"
+
+
+@unique
+class AssignmentWorkflowState(str, Enum):
+    DRAFT = "draft"
```
MarkLark86, line 19: This seems incorrect for the Assignment's workflow state. Should it not be one of the following values:
* draft
* assigned
* in_progress
* completed
* submitted
* cancelled

server/planning/types/common.py
```diff
@@ -0,0 +1,98 @@
...
+
+@dataclass
+class RelatedEvent:
+    id: Annotated[str, validate_data_relation_async("events")] = Field(alias="_id")
+    recurrence_id: str | None = None
+    link_type: str | None = None
+
+
+@dataclass
+class PlanningCoverage:
```
MarkLark86, line 238: I'm not sure if the fields here are correct, checking the mapping for [coverage_schema["planning"]](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/planning/planning_schema.py#L55)

server/planning/types/assignment.py
```diff
@@ -0,0 +1,56 @@
...
+    lock_user: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
+    lock_time: datetime | None = None
+    lock_session: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
+    lock_action: fields.Keyword | None = None
+
+    assigned_to: AssignedTo | None = None
+    planning: PlanningCoverage | None = None
+
+    name: str | None = None
+    description_text: str | None = None
```
MarkLark86: This should rather be `fields.HTML` as it adds the `"analyzer": "html_field_analyzer"` to the mapping (same as original schema)

server/planning/types/common.py
MarkLark86: The Assignment schema also [adds _time_to_be_confirmed field](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/assignments/assignments.py#L1535) to this as well. We could probably just have it here for both uses as well

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+from superdesk.core.resources import fields, dataclass
+from superdesk.core.resources.validators import validate_data_relation_async
+
+from .base import BasePlanningModel
+from .event_dates import EventDates, OccurStatus
+from .enums import ContentState, PostStates, UpdateMethods, WorkflowState
+from .common import CoverageStatus, KeywordQCodeName, PlanningSchedule, RelationshipItem, Subject
+
+
+class NameAnalyzed(str, fields.CustomStringField):
```
MarkLark86: Duplicate class from `planning.types.common.NameAnalyzed`

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+        alias="_planning_schedule", default_factory=list
+    )
+
+    occur_status: OccurStatus | None = None
+    news_coverage_status: CoverageStatus | None = None
+    registration: str | None = None
+    access_status: KeywordQCodeName | None = None
+
+    # Content metadata
+    subject: list[Subject | None] = Field(default_factory=list)
```
MarkLark86: Based on the [original schema](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L199), this should be a [nested list](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L199) (with `include_in_parent` = `True`)

MarkLark86: I see that the Planning schema uses a slight variation. Might be good to combine the two, so we use the same mapping for both (and later content resources in superdesk-core). So I would recommend to keep your current mapping of the Subject specifics, but add the nested mapping as suggested above

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+        "type": "keyword",
+        "fields": {
+            "analyzed": {"type": "text", "analyzer": "html_field_analyzer"},
+        },
+    }
+
+
+class SlugLine(str, fields.CustomStringField):
+    elastic_mapping = {
+        "type": "string",
```
MarkLark86: I think this top later type would be `text` for an elasticsearch types (in eve-elastic we're converting `string` to elastic `text` and/or `keyword` etc, that's why the original schemas use `string`, but we should use elastic standards)

MarkLark86: Same with the other 2 `"type": "string"` in this mapping fields

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # Item type used by superdesk publishing
+    item_type: Annotated[fields.Keyword, Field(alias="type")] = "event"
+
+    # Named Calendars
+    calendars: list[KeywordQCodeName] | None = None
+
+    # The previous state the item was in before for example being spiked,
+    # when un-spiked it will revert to this state
+    revert_state: ContentState | None = None
```
MarkLark86: I think the original schema was incorrect here, it should be `WorkflowState` (as this is populated from `event.state`)

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # Used when duplicating/rescheduling of Events
+    duplicate_from: Annotated[str, validate_data_relation_async("events")] | None = None
```
MarkLark86: `duplicate_from`, `duplicate_to`, `reschedule_from` and `reschedule_to` should all be `fields.Keyword` as we don't want elasticsearch to analyze it's value

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+    duplicate_to: list[Annotated[str, validate_data_relation_async("events")]] = Field(default_factory=list)
+
+    reschedule_from: Annotated[str, validate_data_relation_async("events")] | None = None
+    reschedule_to: Annotated[str, validate_data_relation_async("events")] | None = None
+    reschedule_from_schedule: datetime | None = Field(default=None, alias="_reschedule_from_schedule")
+    place: list[Place] = Field(default_factory=list)
+    ednote: Annotated[str, fields.elastic_mapping({"analyzer": "html_field_analyzer"})] | None = None
```
MarkLark86: This can be simply `fields.HTML`

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+    state_reason: str | None = None
+
+    # Datetime when a particular action (postpone, reschedule, cancel) took place
+    actioned_date: datetime | None = None
+    completed: bool = False
+    time_to_be_confirmed: bool = Field(default=False, alias="_time_to_be_confirmed")
+
+    # This is used if an Event is created from a Planning Item
+    # So that we can link the Planning item to this Event upon creation
+    planning_item: Annotated[str | None, validate_data_relation_async("planning")] = Field(
```
MarkLark86: This should be `fields.Keyword` as we don't want elastic to analyze it's value

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+        default=None, alias="_planning_item"
+    )
+
+    # This is used when event creation was based on `events_template`
+    template: Annotated[str | None, validate_data_relation_async("events_template")] = None
```
MarkLark86: I believe this should be `fields.ObjectId`, as `events_template` stores the ID as an ObjectID

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # This is used when enhancing fetch items to add ids of associated Planning items
+    planning_ids: list[Annotated[str, validate_data_relation_async("planning")]] = Field(default_factory=list)
```
MarkLark86: This should be `fields.ObjectId`

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # HACK: ``coverages`` and ``related_events``
+    # adds these fields to the Events elastic type. So when we're in the Events & Planning filter,
+    # we can send a query to both Event & Planning index without modifying the query.
+    # Otherwise elastic will raise an exception stating the field doesn't exist on the index
+    coverages: CoveragesIndex | None = None
+    related_events: RelatedEvents | None = None
```
MarkLark86, line 260: I don't see `related_events` in the original schema, or am I missing something here?

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+    # HACK: end. We'll try to move this hacks somewhere else
+
+    extra: Annotated[dict[str, Any], fields.elastic_mapping({"type": "object", "dynamic": True})] = Field(
```
MarkLark86: You could use `fields.dynamic_mapping` instead of `fields.elastic_mapping` here

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+        default_factory=dict
+    )
+    translations: Annotated[list[Translation], fields.nested_list()]
+
+    # This is used from the EmbeddedCoverage form in the Event editor
+    # This list is NOT stored with the Event
+    embedded_planning: Annotated[list[EmbeddedPlanning], fields.not_indexed] = Field(default_factory=list)
```
MarkLark86: `fields.not_indexed` is a function, so would need to be called

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # This is used to create new planning items from the event editor
+    # TODO-ASYNC: consider adding proper types instead of a dynamic dict
```
MarkLark86: No need for proper types here, in fact this might be removed in the future (based on the work Tomas is working on). So this comment can be removed

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+    associated_plannings: Annotated[
+        list[dict[str, Any]], fields.elastic_mapping({"type": "object", "dynamic": True})
```
MarkLark86: You an use `fields.dynamic_mapping` instead of `fields.elastic_mapping` here

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+    ] = Field(default_factory=list)
+
+    related_items: list[ContentAPIItem] = Field(default_factory=list)
+    failed_planned_ids: list[str] = Field(default_factory=list)
+
+    # TODO-ASYNC: check why do we have `type` and `_type`
+    _type: str | None = None
```
MarkLark86: `_type` is not needed, as we already have the computed field `type` in core (which provides `_type` already).
Not sure why we have both, but for historic reasons will need it (some code checks for it)

server/planning/types/event.py
```diff
@@ -0,0 +1,285 @@
...
+    # This list is NOT stored with the Event
+    embedded_planning: Annotated[list[EmbeddedPlanning], fields.not_indexed] = Field(default_factory=list)
+
+    # This is used to create new planning items from the event editor
+    # TODO-ASYNC: consider adding proper types instead of a dynamic dict
+    associated_plannings: Annotated[
+        list[dict[str, Any]], fields.elastic_mapping({"type": "object", "dynamic": True})
+    ] = Field(default_factory=list)
+
+    related_items: list[ContentAPIItem] = Field(default_factory=list)
```
MarkLark86: We only need a subset of the fields from the ContentAPIItem (these could also be external items). Therefor it would be good to only define what is in the original schema here

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+    # Ingest Details
+    ingest_provider: Annotated[fields.ObjectId, validate_data_relation_async("ingest_providers")] | None = None
+    source: fields.Keyword | None = None
+    original_source: fields.Keyword | None = None
+    ingest_provider_sequence: fields.Keyword | None = None
+    ingest_firstcreated: datetime = Field(default_factory=utcnow)
+    ingest_versioncreated: datetime = Field(default_factory=utcnow)
+
+    # Agenda Item details
+    agendas: list[Annotated[str, validate_data_relation_async("agenda")]] = Field(default_factory=list)
```
MarkLark86: I believe this would be a list of ObjectIds

server/planning/types/common.py
```diff
@@ -0,0 +1,98 @@
...
+    feature_class: fields.Keyword | None = None
+    location: fields.Geopoint | None = None
+    rel: fields.Keyword | None = None
+
+
+@dataclass
+class RelatedEvent:
+    id: Annotated[str, validate_data_relation_async("events")] = Field(alias="_id")
+    recurrence_id: str | None = None
+    link_type: str | None = None
```
MarkLark86: `link_type` should be one of `["primary", "secondary"]`

server/planning/types/common.py
```diff
@@ -0,0 +1,98 @@
...
+    world_region_code: fields.Keyword | None = None
+    feature_class: fields.Keyword | None = None
+    location: fields.Geopoint | None = None
+    rel: fields.Keyword | None = None
+
+
+@dataclass
+class RelatedEvent:
+    id: Annotated[str, validate_data_relation_async("events")] = Field(alias="_id")
+    recurrence_id: str | None = None
```
MarkLark86: All 3 of these should be `fields.Keyword` as we don't want elastic to analyze their value

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+    related_events: list[RelatedEvent] = Field(default_factory=list)
+    recurrence_id: fields.Keyword | None = None
+    planning_recurrence_id: fields.Keyword | None = None
+
+    # Planning Details
+    # NewsML-G2 Event properties See IPTC-G2-Implementation_Guide 16
+    # Planning Item Metadata - See IPTC-G2-Implementation_Guide 16.1
+    item_class: str = Field(default="plinat:newscoverage")
+    ednote: str | None = None
```
MarkLark86: I think this should be `fields.HTML`, same with `description_text` and `internal_note`, as these can contain rich text (aka HTML)

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+    description_text: str | None = None
+    internal_note: str | None = None
+    anpa_category: list[CVItem] = Field(default_factory=list)
+    subject: list[Subject] = Field(default_factory=list)
```
MarkLark86: This should be nested, as mentioned in a previous comment

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+    genre: list[CVItem] = Field(default_factory=list)
+    company_codes: list[CVItem] = Field(default_factory=list)
+
+    # Content Metadata - See IPTC-G2-Implementation_Guide 16.2
+    language: fields.Keyword | None = None
+    languages: list[fields.Keyword] = Field(default_factory=list)
+    translations: Annotated[list[Translation], fields.nested_list()] = Field(default_factory=list)
+    abstract: str | None = None
```
MarkLark86: Please check `abstract`, `headline`, `keywords` and `slugline`, I believe the first 3 should use `fields.HTML` while the 3rd one should use slugline mapping

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+    # These next two are for spiking/unspiking and purging of planning/agenda items
+    state: WorkflowState = WorkflowState.DRAFT
+    expiry: datetime | None = None
+    expired: bool = False
+    featured: bool = False
+    lock_user: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
+    lock_time: datetime | None = None
+    lock_session: Annotated[fields.ObjectId, validate_data_relation_async("users")] | None = None
+    lock_action: fields.Keyword | None = None
+    coverages: list[PlanningCoverage] = Field(default_factory=list)
```
MarkLark86: This should be a nested mapping

MarkLark86: Also please look at `planning.planning.planning_schema.coverage_schema`, as it contains a lot of fields not defined in `PlanningCoverage` dataclass

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # field to sync coverage scheduled information
+    # to be used for sorting/filtering on scheduled
+    planning_schedule: Annotated[list[dict[str, Any]], fields.nested_list()] = Field(
```
MarkLark86: Might be good to have a dataclass for this one, including fields for `coverage_id` and `scheduled`

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+        default_factory=list, alias="_planning_schedule"
+    )
+
+    # field to sync scheduled_updates scheduled information
+    # to be used for sorting/filtering on scheduled
+    updates_schedule: Annotated[list[dict[str, Any]], fields.nested_list()] = Field(
```
MarkLark86: Same with this one, it would be good to have a dataclass for the fields `scheduled_update_id` and `scheduled`

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+        default_factory=list, alias="updates_schedule"
+    )
+    planning_date: datetime
+    flags: Flags = Field(default_factory=Flags)
+    pubstatus: PostStates | None = None
+    revert_state: WorkflowState | None = None
+
+    # Item type used by superdesk publishing
+    item_type: Annotated[fields.Keyword, Field(alias="type")] = "planning"
```
MarkLark86: This one won't be needed here, as the `planning_resource_config.name` has a value of `planning` which the `type` computed field on `ResourceModel` will use.

This is different for `Assignments` and Events` and the `ResourceConfig.name` and `_type` have different values

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+    place: list[Place] = Field(default_factory=list)
+    name: str | None = None
+    files: list[Annotated[str, validate_data_relation_async("planning_files")]] = Field(default_factory=list)
```
MarkLark86: This should rather be a list of ObjectId.
Also when annotating a list type you place `list[<str>]` inside annotated (this is because you're annotated the type `list[str]` not a list of annotated type `str`). i.e.

```python
files: Annotated[list[fields.ObjectId], validate_data_relation_async("planning_files")]
```

server/planning/types/planning.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Reason (if any) for the current state (cancelled, postponed, rescheduled)
+    state_reason: str | None = None
+    time_to_be_confirmed: bool = Field(default=False, alias="_time_to_be_confirmed")
+    extra: Annotated[dict[str, Any], fields.elastic_mapping({"type": "object", "dynamic": True})] = Field(
```
MarkLark86: You can use `fields.dynamic_mapping` here instead of `fields.elastic_mapping`

server/planning/assignments/module.py
eos87: I just realised that `published_state` is missing from the fields here and in the old schema. I will add it. 

As for the options, I'd say we leave it as it is for now and just change it where it is required. If we encounter more issues like this we can change it in `superdesk-core` later and adjust `newsroom-core`. Or do you see this happening very often?

server/planning/assignments/service.py
eos87: It was throwing an error before I register the module and I didn't realise that was the problem instead. I'll remove it.

server/planning/types/assignment.py
eos87: yeah, I think you're right it would clash. I would comment it for now and leave a note and then we see if we have to adjust it or remove it.

server/planning/types/assignment.py
eos87: Good idea. Done.

server/planning/types/assignment.py
eos87: having the decorator and inheriting from class feels redundant. Can we move or add the things from the decorator into the class also?

server/planning/types/enums.py
eos87: oops, not sure how I've missed that.

server/planning/types/event.py
eos87: right, I forgot to remove it when I moved it to common.

server/planning/types/event.py
eos87: Done, I added it to common as a type and reused it in `event.py` and `planning.py`

server/planning/types/event.py
eos87: Indeed, at the time I had to look out for the difference actually, but this one escaped my eyes. Thanks!

server/planning/types/event.py
eos87: It's there in the async branch, as part of "hack" for the searches. I will remove this in my next PR and replace it with the other solution we were talking about the other day. 
https://github.com/superdesk/superdesk-planning/blob/82ccb8ce64d7c189c41353338e5afac1929df2b0/server/planning/events/events_schema.py#L329-L358

server/planning/types/planning.py
eos87: done

server/planning/types/planning.py
eos87: removed

server/planning/types/common.py
eos87: Fixed. I've mixed things up

eos87:
Hey @MarkLark86, I've made the changes and left a couple of comments in the "unresolved" conversations. I added this small helper method https://github.com/superdesk/superdesk-planning/blob/b85a4e230898fda3eb717c305e1459749eeef6c2/server/planning/types/common.py#L202-L207 do you think this could be useful in other places? At the moment I'm using it like this https://github.com/superdesk/superdesk-planning/blob/b85a4e230898fda3eb717c305e1459749eeef6c2/server/planning/types/planning.py#L103-L106 to avoid duplicating the same thing just in the form of a elastic mapping. I guess there would be other places where we would need something similar.

server/planning/types/event.py
MarkLark86: Oh right, I was looking at the `develop` branch, my bad

server/planning/assignments/module.py
MarkLark86: There wouldn't be much more, they're all left in superdesk-core now I believe, so we just have to be mindful when configuring the indexes. So yes, leave as is for now and explicitly put `unique=False` where required

server/planning/types/assignment.py
MarkLark86: I agree, though not sure how to do that at this point in time to make the class a dataclass so it works with Pydantic. Inheriting from BaseModel is not advised by pydantic due to performance issues there, and dataclass won't work (to me, dataclasses seem more natural for a programmer, using dot notation to access fields).

Only other way I can think of is the [ORM Mode](https://docs.pydantic.dev/1.10/usage/models/#orm-mode-aka-arbitrary-class-instances), but not sure if that creates a performance issue as mentioned above, and then each root model would need to have the `orm_mode` turned on

MarkLark86:
> Hey @MarkLark86, I've made the changes and left a couple of comments in the "unresolved" conversations. I added this small helper method
> 
> https://github.com/superdesk/superdesk-planning/blob/b85a4e230898fda3eb717c305e1459749eeef6c2/server/planning/types/common.py#L202-L207
> 
> do you think this could be useful in other places? At the moment I'm using it like this
> https://github.com/superdesk/superdesk-planning/blob/b85a4e230898fda3eb717c305e1459749eeef6c2/server/planning/types/planning.py#L103-L106
> 
> to avoid duplicating the same thing just in the form of a elastic mapping. I guess there would be other places where we would need something similar.

Not sure there would be many other places where this would be used, as it's mostly covered by the new type system. This use case is an edge case, all others I can think of are for the storage and search of the data itself, so should be fine to leave where it is

server/planning/types/common.py
```diff
@@ -0,0 +1,267 @@
...
+
+    assigned_to: CoverageAssignedTo | None = None
+    news_coverage_status: NewsCoverageStatus = Field(default_factory=dict)
+    planning: ScheduledUpdatePlanning = Field(default_factory=dict)
+
+
+@dataclass
+class PlanningCoverage:
+    # Identifiers
+    coverage_id: fields.Keyword
```
MarkLark86: Looks like this will also need to be optional, as we're generating one in the planning service it one is not defined

BrianMwangi21:
Hey @eos87 just a few more changes to some fields from mandatory to optional: 

For `EventResourceModel` the following fields:
- `name`
- `event_contact_info`
- `translations`

For `PlanningResourceModel` the following fields: 
- `coverages.coverage_id`
- `coverages.assigned_to`

server/planning/types/common.py
eos87: done

eos87:
@BrianMwangi21 fixed `event_contact_info` and `translations` providing default factory constructors for them. As for `name` I think it should be required. I suppose we should not create an event without a name at least.

Regarding `PlanningResourceModel`, `coverage_id` has been fixed.  `assigned_to` had already a default factory constructor.

BrianMwangi21:
Thanks. I'll see what to do about the names.

server/planning/types/assignment.py
eos87: I will play with it maybe I can come up with some idea.

server/planning/assignments/service.py
```diff
@@ -0,0 +1,5 @@
+from planning.core.service import BasePlanningAsyncService
+
+
+class AssingmentsAsyncService(BasePlanningAsyncService):
```
BrianMwangi21: Small nit here. `Assignments` is misspelled

eos87: Nice catch. Fixed.

server/planning/types/assignment.py
eos87: https://github.com/superdesk/superdesk-core/pull/2764

MarkLark86 (approved):

PR #2132, opened 2024-11-21
Add config to disable automatic assignment assignee reassignment [SDCP-860]

devketanpro:
This PR introduces a new configuration, `ASSIGNMENT_MANUAL_REASSIGNMENT_ONLY`, to allow manual control over assignment reassignments. When enabled, actions like saving or moving an article will no longer automatically update the assignee. Manual reassignment becomes the only way to change the assignee


server/planning/assignments/assignments.py
```diff
@@ -862,7 +863,9 @@ def _get_empty_updates_for_assignment(self, assignment):
 
     def _set_user_for_assignment(self, assignment, assignee, assignor=None):
         updates = self._get_empty_updates_for_assignment(assignment)
-        updates["assigned_to"]["user"] = assignee
+
+        if not get_config_assignment_manual_reassignment_only():
+            updates["assigned_to"]["user"] = assignee
 
         if assignor:
```
MarkLark86: If the config is turned off, then we should probably not change the `assignor_user` either (as no re-assignment was performed)

MarkLark86 (commented):
Would be good to have some tests too, for when this config is turned ON

server/planning/assignments/assignments.py
devketanpro: got it

MarkLark86 (commented):
Looks good, but can you please investigate the failing tests.

devketanpro:
I checked on the develop branch, and the tests are failing there as well. It seems the issue is not related to these changes.I'll continue investigating further.

devketanpro:
Hey @MarkLark86, Here is the PR for the test case fix: https://github.com/superdesk/superdesk-planning/pull/2138

MarkLark86 (approved):

PR #2134, opened 2024-11-22
[SDESK-7443] - Planning: Migrate planning:delete_spiked command to async

BrianMwangi21:
### Purpose

This PR updates the `planning:delete_spiked` command to async by: 

- Making use of the new async methods from events and planning
- Fixing failing tests

Solves [SDESK-7443](https://sofab.atlassian.net/browse/SDESK-7443)

> [!NOTE]
> This PR depends on https://github.com/superdesk/superdesk-planning/pull/2131 so it is quite big but will become concise when the PR is merged.

[SDESK-7443]: https://sofab.atlassian.net/browse/SDESK-7443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/commands/delete_spiked_items.py
+++ b/server/planning/commands/delete_spiked_items.py
@@ -20,8 +20,8 @@ from superdesk.celery_task_utils import get_lock_id
 from superdesk.lock import lock, unlock, remove_locks
 from planning.common import WORKFLOW_STATE
 from planning.events import EventsAsyncService
-from planning.planning.service import PlanningAsyncService
-from planning.assignments.service import AssingmentsAsyncService
+from planning.planning import PlanningAsyncService
+from planning.assignments import AssingmentsAsyncService
 from .async_cli import planning_cli
 
 
--- a/server/planning/commands/delete_spiked_items.py
+++ b/server/planning/commands/delete_spiked_items.py
@@ -13,13 +13,13 @@ from contextvars import ContextVar
 
 from superdesk.core import get_app_config
 from superdesk.resource_fields import ID_FIELD
-from superdesk import get_resource_service
 from superdesk.logging import logger
 from superdesk.utc import utcnow
 from superdesk.celery_task_utils import get_lock_id
 from superdesk.lock import lock, unlock, remove_locks
 from planning.common import WORKFLOW_STATE
 from planning.events import EventsAsyncService
+from planning.events.utils import get_recurring_timeline
 from planning.planning import PlanningAsyncService
 from planning.assignments import AssingmentsAsyncService
 from .async_cli import planning_cli
@@ -93,7 +93,7 @@ async def delete_spiked_events(expiry_datetime):
 
     for event_id, event in events.items():
         if event.get("recurrence_id") and event["recurrence_id"] not in series_to_delete:
-            spiked, events = is_series_expired_and_spiked(event, expiry_datetime)
+            spiked, events = await is_series_expired_and_spiked(event, expiry_datetime)
             if spiked:
                 series_to_delete[event["recurrence_id"]] = events
         else:
@@ -108,24 +108,24 @@ async def delete_spiked_events(expiry_datetime):
     logger.info(f"{log_msg} {len(events_deleted)} Events deleted: {list(events_deleted)}")
 
 
-def is_series_expired_and_spiked(event, expiry_datetime):
-    historic, past, future = get_resource_service("events").get_recurring_timeline(event, spiked=True)
+async def is_series_expired_and_spiked(event, expiry_datetime):
+    historic, past, future = await get_recurring_timeline(event, spiked=True)
 
     # There are future events, so the entire series is not expired.
     if len(future) > 0:
-        return False
+        return False, []
 
     def check_series_expired_and_spiked(series):
         for event in series:
             if event.get("state") != WORKFLOW_STATE.SPIKED or event["dates"]["end"] > expiry_datetime:
-                return False
+                return False, []
 
-        return True
+        return True, []
 
     if check_series_expired_and_spiked(historic) and check_series_expired_and_spiked(past):
         return True, [historic + past]
 
-    return False
+    return False, []
 
 
 async def delete_spiked_planning(expiry_datetime):
--- a/server/planning/commands/delete_spiked_items.py
+++ b/server/planning/commands/delete_spiked_items.py
@@ -103,13 +103,13 @@ async def delete_spiked_events(expiry_datetime):
     # Delete recurring series
     for recurrence_id, events in series_to_delete.items():
         await events_service.delete_many(lookup={"recurrence_id": recurrence_id})
-        events_deleted.add(events)
+        events_deleted.add([event["_id"] for event in events])
 
     logger.info(f"{log_msg} {len(events_deleted)} Events deleted: {list(events_deleted)}")
 
 
 async def is_series_expired_and_spiked(event, expiry_datetime):
-    historic, past, future = await get_recurring_timeline(event, spiked=True)
+    historic, past, future = await get_recurring_timeline(event, spiked=True, postponed=True)
 
     # There are future events, so the entire series is not expired.
     if len(future) > 0:
@@ -118,9 +118,9 @@ async def is_series_expired_and_spiked(event, expiry_datetime):
     def check_series_expired_and_spiked(series):
         for event in series:
             if event.get("state") != WORKFLOW_STATE.SPIKED or event["dates"]["end"] > expiry_datetime:
-                return False, []
+                return False
 
-        return True, []
+        return True
 
     if check_series_expired_and_spiked(historic) and check_series_expired_and_spiked(past):
         return True, [historic + past]
--- a/server/planning/commands/delete_spiked_items.py
+++ b/server/planning/commands/delete_spiked_items.py
@@ -20,6 +20,8 @@ from superdesk.celery_task_utils import get_lock_id
 from superdesk.lock import lock, unlock, remove_locks
 from planning.common import WORKFLOW_STATE
 from planning.events import EventsAsyncService
+from planning.planning.service import PlanningAsyncService
+from planning.assignments.service import AssingmentsAsyncService
 from .async_cli import planning_cli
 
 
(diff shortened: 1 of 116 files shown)


server/planning/events/service.py
```diff
@@ -0,0 +1,50 @@
...
+        if spiked_events_only:
+            query["query"] = {"bool": {"must": [{"term": {"state": WORKFLOW_STATE.SPIKED}}]}}
+
+        total_received = 0
+        total_events = -1
+
+        while True:
+            query["from"] = total_received
+
+            results = self.search(query)
```
MarkLark86: `search` is an async method, so you would need to await it's response.

server/planning/events/service.py
```diff
@@ -0,0 +1,50 @@
...
+
+            # If the total_events has not been set, then this is the first query
+            # In which case we need to store the total hits from the search
+            if total_events < 0:
+                total_events = results.count()
```
MarkLark86: The `count` method on the cursor is also an async method, so you will need to await it's response.
See [the docs for examples](https://superdesk.readthedocs.io/en/async/core/resource_management.html#resource-services)

server/planning/events/service.py
```diff
@@ -0,0 +1,50 @@
...
+
+                # If the search doesn't contain any results, return here
+                if total_events < 1:
+                    break
+
+            # If the last query doesn't contain any results, return here
+            if not len(results.docs):
```
MarkLark86: Rather than counting `len(results.docs` (which is an internal field to the cursor), you can store `await results.count()` before `if total_events < 0:` and re-use that count in this loop

server/planning/events/service.py
```diff
@@ -0,0 +1,50 @@
...
+                break
+
+            total_received += len(results.docs)
+
+            # Yield the results for iteration by the callee
+            yield list(results.docs)
```
MarkLark86: Here you would yield from the cursor instead, for example: ```yield await results.to_list_raw()```

server/planning/events/service.py
```diff
@@ -0,0 +1,50 @@
...
+
+from planning.types import EventResourceModel
+from planning.common import get_max_recurrent_events, WORKFLOW_STATE
+from planning.core.service import BasePlanningAsyncService
+
+
+class EventsAsyncService(BasePlanningAsyncService[EventResourceModel]):
+    resource_name = "events"
+
+    async def get_expired_items(self, expiry_datetime, spiked_events_only=False):
```
MarkLark86: It would be good to add types here. Such as:

```python
from typing import AsyncGenerator, Any
from datetime import datetime

class EventsAsyncService(BasePlanningAsyncService[EventResourceModel]):
    async def get_expired_items(self, expiry_datetime: datetime, spiked_events_only: bool = False) -> AsyncGenerator[dict[str, Any]]:
        ...
```

This way the `mypy` command can check types used

server/planning/events/utils.py
```diff
@@ -0,0 +1,99 @@
...
+    WORKFLOW_STATE,
+    get_max_recurrent_events,
+)
+from planning.events import EventsAsyncService
+from superdesk.resource_fields import ID_FIELD
+from superdesk.utc import utcnow
+
+
+async def get_series(query, sort, max_results):
+    events_service = EventsAsyncService()
```
MarkLark86: There has been an improvement to the core framework, you can now get the service directly from the resource model type (if all you're doing is simple CRUD of the underlying data, any custom service methods still need to directly import the service).

For example:
```python

async def get_series(query, sort, max_results):
    events_service = EventResourceModel.get_service()
```

server/planning/events/utils.py
```diff
@@ -0,0 +1,99 @@
...
+from planning.common import (
+    WORKFLOW_STATE,
+    get_max_recurrent_events,
+)
+from planning.events import EventsAsyncService
+from superdesk.resource_fields import ID_FIELD
+from superdesk.utc import utcnow
+
+
+async def get_series(query, sort, max_results):
```
MarkLark86: Please add types to this function, including the return type (which is also an AsyncGenerator of dict[str, Any])

server/planning/events/utils.py
```diff
@@ -0,0 +1,99 @@
...
+    events_service = EventsAsyncService()
+    page = 1
+
+    while True:
+        # Get the results from mongo
+        req = ParsedRequest()
```
MarkLark86: This was the old way of doing search, instead you can use `events.find` with the param `use_mongo=True`. For example:

```python
results = await events_service.find(query, sort=sort, max_results=max_results, page=page)
```

If you look at the type signature of that method, you will see `sort` uses `superdesk.core.types.SortParam` (which you can use in the signature of this function)

MarkLark86: Also, the response from `.find` is our own cursor, so list(results) won't work, you would need to use `docs = await results.to_list_raw())`

server/planning/events/utils.py
```diff
@@ -0,0 +1,99 @@
...
+            break
+
+        page += 1
+
+        # Yield the results for iteration by the callee
+        for doc in docs:
+            yield doc
+
+
+async def get_recurring_timeline(
```
MarkLark86, line 35: Would be good to have the signature of this function typed, including the return type (tuple of 3 dicts)

server/planning/events/utils.py
```diff
@@ -0,0 +1,99 @@
...
+        "$and": [
+            {"recurrence_id": selected["recurrence_id"]},
+            {"_id": {"$ne": selected[ID_FIELD]}},
+        ]
+    }
+
+    if excluded_states:
+        query["$and"].append({"state": {"$nin": excluded_states}})
+
+    sort = '[("dates.start", 1)]'
```
MarkLark86: Here you would use the `superdesk.core.types.SortParam` type to construct the sort param to use

server/planning/planning/service.py
```diff
@@ -0,0 +1,78 @@
...
+
+from planning.types import PlanningResourceModel
+from planning.common import WORKFLOW_STATE
+from planning.core.service import BasePlanningAsyncService
+
+
+class PlanningAsyncService(BasePlanningAsyncService[PlanningResourceModel]):
+    resource_name = "planning"
+
+    async def get_expired_items(self, expiry_datetime, spiked_planning_only=False):
```
MarkLark86: Same as with Events.get_expired_items method, would be good to have the type defined for the signature

server/planning/planning/service.py
```diff
@@ -0,0 +1,78 @@
...
+        query["sort"] = [{"planning_date": "asc"}]
+        query["size"] = 200
+
+        total_received = 0
+        total_items = -1
+
+        while True:
+            query["from"] = total_received
+
+            results = self.search(query)
```
MarkLark86: Same as with the Events service, use await and the cursor properly here

server/planning/commands/delete_spiked_items.py
```diff
@@ -30,123 +39,125 @@ class DeleteSpikedItems(Command):
...
-            else:
-                events_service.delete_action(lookup={"_id": event_id})
-                events_deleted.add(event_id)
+    for event_id, event in events.items():
+        if event.get("recurrence_id") and event["recurrence_id"] not in series_to_delete:
+            spiked, events = await is_series_expired_and_spiked(event, expiry_datetime)
+            if spiked:
+                series_to_delete[event["recurrence_id"]] = events
+        else:
+            await events_service.delete_action(lookup={"_id": event_id})
```
MarkLark86: There is no `delete_action` in the new resource services, instead there is a `delete_many` method

server/planning/commands/delete_spiked_items.py
```diff
@@ -30,123 +39,125 @@ class DeleteSpikedItems(Command):
...
+            events_deleted.add(event_id)
 
-        # Delete recurring series
-        for recurrence_id, events in series_to_delete.items():
-            events_service.delete_action(lookup={"recurrence_id": recurrence_id})
-            events_deleted.add(events)
+    # Delete recurring series
+    for recurrence_id, events in series_to_delete.items():
+        await events_service.delete_action(lookup={"recurrence_id": recurrence_id})
```
MarkLark86: Same here, use the `delete_many` method

server/planning/commands/delete_spiked_items.py
```diff
@@ -30,123 +39,125 @@ class DeleteSpikedItems(Command):
...
+            assignment_id = (coverage.get("assigned_to") or {}).get("assignment_id")
+            if assignment_id:
+                assignments_to_delete.append(assignment_id)
 
-        logger.info(
-            "{} {} Assignments deleted: {}".format(self.log_msg, len(assignments_deleted), list(assignments_deleted))
-        )
-        logger.info("{} {} Planning items deleted: {}".format(self.log_msg, len(plans_deleted), list(plans_deleted)))
+        # Now, delete the planning item
+        await planning_service.delete_action(lookup={"_id": plan_id})
```
MarkLark86: And again here, use the `delete_many` method instead

server/planning/commands/delete_spiked_items.py
```diff
@@ -30,123 +39,125 @@ class DeleteSpikedItems(Command):
...
+        plans_deleted.add(plan_id)
 
+    # Delete assignments
+    assignment_service = AssingmentsAsyncService()
+    for assign_id in assignments_to_delete:
+        await assignment_service.delete(lookup={"_id": assign_id})
```
MarkLark86: And once more, use the `delete_many` method. Otherwise if you want to delete a single item you can:

```python
assignment = await assignment_service.find_by_id(assign_id)
await assignment_service.delete(assignment)
```

server/planning/events/utils.py
BrianMwangi21: Oh, this is nice

MarkLark86:
@BrianMwangi21 There seems to be a couple unit tests failing. One with the elasticsearch query for the delete command, and another in the test using dictionary access instead of ResourceModel access. i.e.

```python
assignments[plan_id] = planning["coverages"][0]["assigned_to"]["assignment_id"]
TypeError: 'PlanningResourceModel' object is not subscriptable
```

BrianMwangi21:
Yeah, I'm working on those.

MarkLark86 (approved):

PR #2138, opened 2024-11-25
Fix behave tests

devketanpro:
Check Behave tests


MarkLark86 (approved):
Who would have thought 2024 would no longer be in the future, hahaha :stuck_out_tongue:

PR #2142, opened 2024-12-02
[SDESK-7445] - Planning: Migrate planning:purge_expired_locks command to async

BrianMwangi21:
### Purpose

This PR updates the `planning:purge_expired_locks` command to async by: 

- Making use of the new async methods from events, planning and assignments to get list of expired locks and remove them.
- Updated tests

Solves [SDESK-7445](https://sofab.atlassian.net/browse/SDESK-7445)

> [!NOTE]
> This PR depends on https://github.com/superdesk/superdesk-planning/pull/2131 so it is quite big but will become concise when the PR is merged.

[SDESK-7443]: https://sofab.atlassian.net/browse/SDESK-7445

[SDESK-7445]: https://sofab.atlassian.net/browse/SDESK-7445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/commands/purge_expired_locks.py
+++ b/server/planning/commands/purge_expired_locks.py
@@ -121,7 +121,6 @@ async def purge_item_locks(resource: str, expiry_datetime: datetime):
                         LOCK_SESSION: None,
                         LOCK_TIME: None,
                     },
-                    item,
                 )
             except Exception as err:
                 logger.exception(f"Failed to purge item lock ({err})")
@@ -162,6 +161,7 @@ async def get_locked_items(resource: str, expiry_datetime: datetime) -> AsyncGen
         query["from"] = total_received
         results = await resource_service.search(query)
         items = await results.to_list_raw()
+        print("Items from query:", items)
 
         num_results = len(items)
 
--- a/server/planning/commands/purge_expired_locks.py
+++ b/server/planning/commands/purge_expired_locks.py
@@ -113,7 +113,7 @@ async def purge_item_locks(resource: str, expiry_datetime: str):
 
             try:
                 # Remove all lock information from this item
-                resource_service.system_update(
+                await resource_service.system_update(
                     item_id,
                     {
                         LOCK_USER: None,
--- a/server/planning/commands/purge_expired_locks.py
+++ b/server/planning/commands/purge_expired_locks.py
@@ -10,7 +10,7 @@
 
 import click
 import logging
-from datetime import timedelta
+from datetime import timedelta, datetime
 from eve.utils import date_to_str
 from typing import AsyncGenerator, Any
 
@@ -83,7 +83,7 @@ async def purge_expired_locks_handler(resource: str, expire_hours: int = 24):
         logger.info("purge expired locks task is already running")
         return
 
-    expiry_datetime = date_to_str(utcnow() - timedelta(hours=expire_hours))
+    expiry_datetime = utcnow() - timedelta(hours=expire_hours)
     for resource_name in resources:
         try:
             await purge_item_locks(resource_name, expiry_datetime)
@@ -94,7 +94,7 @@ async def purge_expired_locks_handler(resource: str, expire_hours: int = 24):
     logger.info("Completed purging expired item locks")
 
 
-async def purge_item_locks(resource: str, expiry_datetime: str):
+async def purge_item_locks(resource: str, expiry_datetime: datetime):
     logger.info(f"Purging expired locks for {resource}")
     resource_service = SERVICE_MAPPING[resource]()
     try:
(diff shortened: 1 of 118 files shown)


server/planning/commands/purge_expired_locks.py
```diff
@@ -39,109 +65,103 @@ class PurgeExpiredLocks(Command):
...
+
+    unlock(lock_name)
+    logger.info("Completed purging expired item locks")
+
+
+async def purge_item_locks(resource: str, expiry_datetime: str):
+    logger.info(f"Purging expired locks for {resource}")
+    resource_service = SERVICE_MAPPING[resource]
+    try:
+        autosave_service = get_resource_service("event_autosave" if resource == "events" else f"{resource}_autosave")
```
BrianMwangi21, line 93: @MarkLark86 we are making use of autosave services here. They are yet to be converted to async. Do they remain as is and just add a TODO to get back to them when finished ?

MarkLark86: That sounds fine

server/planning/commands/purge_expired_locks.py
```diff
@@ -8,22 +8,48 @@
...
+from .async_cli import planning_cli
+from planning.events import EventsAsyncService
+from planning.planning import PlanningAsyncService
+from planning.assignments import AssignmentsAsyncService
 
 logger = logging.getLogger(__name__)
-
-
-class PurgeExpiredLocks(Command):
+SERVICE_MAPPING = {
```
MarkLark86: It would be good to have types defined here. It might also help with potential cyclic imports to use the resource models instead. For example:

```python
# file: server/planning/utils.py
from planning.types import BasePlanningModel, EventResourceModel, PlanningResourceModel, AssignmentResourceModel
from planning.core import BasePlanningAsyncService

RESOURCE_MAPPING: dict[str, BasePlanningModel] = {
    "events": EventResourceModel,
    "planning": PlanningResourceModel,
    "assignments": AssignmentResourceModel,
}

def get_service(item_type: str) -> BasePlanningAsyncService:
    resource_model = RESOURCE_MAPPING.get(item_type)
    if resource_model is None:
        raise BadRequest()
    return resource_model.get_service()
```

Then in any of our code where we need to handle different types, we can import this function and use it.
For example:
```python
from planning.utils import get_service

async def test_something() -> None:
    cursor = await get_service("events").search({...})
```

This will provide type checking as well as other improvements

server/planning/commands/purge_expired_locks.py
```diff
@@ -8,22 +8,48 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
+import click
 import logging
-from datetime import timedelta
+from datetime import timedelta, datetime
 from eve.utils import date_to_str
```
MarkLark86: One goal of this project is to remove the eve library. We now have this [date_to_str](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/utils.py#L69) in our own core library, therefor it would be good to use that import here instead

server/planning/commands/purge_expired_locks_test.py
```diff
@@ -24,102 +26,138 @@
 
 # TODO: Add Assignments
 class PurgeExpiredLocksTest(TestCase):
+    app_config = {
```
MarkLark86: If we're not adding/modifying the `app_config`, then there's no need to re-define it here as it's already defined in the parent class

server/planning/commands/purge_expired_locks_test.py
```diff
@@ -24,102 +26,138 @@
...
-                ("planning", "expired_plan_1", True),
-                ("assignments", assignment_1_id, True),
-                ("assignments", assignment_2_id, True),
-            ]
-        )
-
-    def test_invalid_resource(self):
-        with self.assertRaises(ValueError):
-            PurgeExpiredLocks().run("blah")
+        self.app_config.update({"MODULES": ["planning.module"]})
```
MarkLark86: This is already defined in the parent class' `app_config`, so this line can be removed

server/planning/commands/purge_expired_locks_test.py
```diff
@@ -24,102 +26,138 @@
...
+
+        self.service_mapping = {
```
MarkLark86: Here we can also use the util I mentioned earlier to get the async resource service

MarkLark86 (commented):
Looks good, just a couple small suggestions

MarkLark86 (approved):

PR #2143, opened 2024-12-02
Fix coverage data saving/rendering from embed plan

thecalcc:
SDCP-863
SDCP-875

## Front-end checklist

### DIFF
--- a/client/api/events.ts
+++ b/client/api/events.ts
@@ -176,7 +176,7 @@ function create(updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
 function update(original: IEventItem, updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
     return superdeskApi.dataApi.patch<IEventItem>('events', original, {
         ...updates,
-        associated_plannings: undefined,
+        associated_plannings: updates.associated_plannings,
         embedded_planning: updates?.associated_plannings?.map((planning) => ({
             planning_id: planning._id.startsWith(TEMP_ID_PREFIX) ? undefined : planning._id,
             update_method: planning.update_method,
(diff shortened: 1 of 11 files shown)


client/api/events.ts
```diff
@@ -176,7 +176,7 @@ function create(updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
 function update(original: IEventItem, updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
     return superdeskApi.dataApi.patch<IEventItem>('events', original, {
         ...updates,
-        associated_plannings: undefined,
+        associated_plannings: updates.associated_plannings,
```
thecalcc: @MarkLark86 this is the fix I found to be working so we persist the changes, and send the proper data back in the UI. 

This is a follow up PR based on these comments - https://sourcefabric.slack.com/archives/C06HDLLA1SP/p1733139297225319

Let me know if that's how things should work, from my side there's no issues I see in the UI, and the changes do work.

thecalcc: All other changes are just improvements, this is the actual change that's fixing the issues mentioned in the linked tickets

MarkLark86: Hey @thecalcc 
`associated_plannings` is set to `undefined` because we don't want to store this on the Event item itself (this happens for both create and update actions).

Instead what I would try is to re-populate this field after the response from the server/update action:

https://github.com/superdesk/superdesk-planning/blob/f69090523b4fc179c82d36dfa0086fb984cbb38c/client/api/events.ts#L199-L202

```ts
.then((response) => { 
    const events = modifySaveResponseForClient(response);
    
    events[0].associated_plannings = updates.associated_plannings
    return planningApi.planning.searchGetAll({ 
```

Can you please try that and see if that fixes the issue

thecalcc: Ok makes sense, just tried that and it does work

MarkLark86 (approved):
Looks fine

PR #2147, opened 2024-12-03
Planning authoring-react fields

thecalcc:
STT-63

## Front-end checklist


thecalcc:
@tomaskikutis what's a better way to handle this? I have the branch based off of yours - `authoring-react-planning`

Edit: might be a good idea for you to push your branch to the remote so I can merge to it as well, while not messing directly with it

tomaskikutis:
I've deleted the branch from my fork and pushed it to this repo. You'll have to fix a conflict now since I've changed the content profile file.

tomaskikutis:
why are you removing package-lock from planning extension?

thecalcc:
> why are you removing package-lock from planning extension?

Unintentionally, I will revert it

client/planning-extension/src/extension.ts
```diff
@@ -255,6 +256,7 @@ const extension: IExtension = {
                 globalMenuHorizontal: [AssignmentsCountTracker, ...(displayTopbarWidget ? [AssignmentsList] : [])],
                 customFieldTypes: [
                     getAttachmentsField(),
+                    getCustomVocabulariesField(),
```
tomaskikutis: how are custom vocabularies different from a [vocabulary based dropdown](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring-react/fields/dropdown/dropdown-vocabulary/editor.tsx)?

thecalcc: I went with using this field first, but then decided on rather reusing the planning one so it's quicker. One reason is that CustomVocabularies actually loops through an array of custom vocabularies. Starting with this I will need to actually remake the Vocabulary based dropdown field we have. Potentially export the component, or something else. Then I have to take care of fetching all the vocabularies and other dependencies already handled in the planning one I reused. With that there's no need to expose other interfaces (as I saw), functions, etc. 

I then checked the `CustomVocabularies.tsx` implementation and it's actually very simple. So I decided to reuse that under the hood, while doing minimal changes to everything else. Let me know if you still want to make it that way. But for now I see no reason to change it. In the future we can still decouple the implementation of the [vocabulary based dropdown](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/apps/authoring-react/fields/dropdown/dropdown-vocabulary/editor.tsx), export it and reuse that under the hood. But a new field type would need to be defined - the main difference being that custom vocabularies is an array of vocabularies (vocabulary IDs to be specific)

tomaskikutis:
Usually if I leave some comments I expect you would only push next after you handled it which seems not to be the case here right?

client/planning-extension/package-lock.json
```diff
@@ -1,1804 +1,1804 @@
...
-      "requires": {
-        "acorn": "^3.0.4"
-      },
-      "dependencies": {
-        "acorn": {
-          "version": "3.3.0",
-          "resolved": "https://registry.npmjs.org/acorn/-/acorn-3.3.0.tgz",
-          "integrity": "sha512-OLUyIIZ7mF5oaAUT1w0TFqQS81q3saT46x8t7ukpPjMNk+nbs4ZHhs7ToV8EWnLYLepjETXd4XaCE4uxkMeqUw==",
-          "dev": true
+    "name": "superdesk-planning-extension",
```
tomaskikutis: no changes should be present in the PR of
`client/planning-extension/package-lock.json`

if you revert by copying over manually - use an editor that doesn't modify formatting

client/components/planning-editor-standalone/profile.ts
```diff
@@ -52,108 +71,173 @@ interface IFieldDefinition {
...
+
+                return field;
+            },
+        }
+    ];
+    const planningProfile = planningApi.contentProfiles.get('planning');
+    const customVocabularyIds =
+        (planningProfile.schema?.['custom_vocabularies'] as IProfileSchemaTypeList)?.vocabularies;
+
+    if ((customVocabularyIds?.length ?? 0) > 0) {
```
tomaskikutis: custom vocabularies definition is quite big, could you move it to a separate file? here you would import the function, call it and push the result to the array of fields

client/components/planning-editor-standalone/storage-adapter.ts
```diff
@@ -15,7 +16,18 @@ export const storageAdapterPlanningItem: IStorageAdapter<IPlanningItem> = {
         const fieldDefinitions = getFieldDefinitions();
         const fieldStorageAdapter = fieldDefinitions[fieldId]?.storageAdapter;
 
-        if (fieldStorageAdapter != null) {
+        if (fieldId.includes(SUBJECT_PREFIX_ID)) {
```
tomaskikutis: use `.startsWith` instead of `.includes`

Regarding `SUBJECT_PREFIX_ID` - would be good if we could avoid this approach of checking contents of an ID. I suggested the idea to assure you that it can be done using existing vocabulary fields from client-core. Since we agree now that it can be done, we can look into how to do it better :) It's a bit messy that you run storage adapter twice for vocabulary fields. Once in field definition and second time in main storage adapter. Change `IFieldDefinition['storageAdapter']` so you can do it all in once place.

client/components/planning-editor-standalone/storage-adapter.ts
```diff
@@ -45,15 +43,16 @@ export const storageAdapterPlanningItem: IStorageAdapter<IPlanningItem> = {
     retrieveStoredValue: (item, fieldId, fieldType) => {
         const {getContentStateFromHtml} = superdeskApi.helpers;
         const value = (item as {[key: string]: any})[fieldId] ?? undefined;
-
         const fieldDefinitions = getFieldDefinitions();
         const fieldStorageAdapter = fieldDefinitions[fieldId]?.storageAdapter;
 
-        if (fieldStorageAdapter != null) {
+        if (fieldId.startsWith(SUBJECT_PREFIX_ID)) {
```
tomaskikutis: use the same approach here as well to avoid running 2 storage adapters

client/components/planning-editor-standalone/profile.ts
```diff
@@ -177,13 +203,23 @@ export function getProfile() {
...
-
     const fieldDefinitions = getFieldDefinitions();
+    const customVocabularyIds =
+        (planningProfile.schema?.['custom_vocabularies'] as IProfileSchemaTypeList)?.vocabularies;
 
     for (const fieldId of planningFieldIds) {
         const required = planningProfile.schema?.[fieldId]?.required ?? false;
 
-        if (fieldDefinitions[fieldId] != null) {
+        if (fieldId === 'custom_vocabularies') {
```
tomaskikutis: It looks to me this conditional can be avoided. Let's chat if you think otherwise

client/components/planning-editor-standalone/field-adapters/custom-vocabularies.ts
```diff
@@ -0,0 +1,69 @@
...
+                        fieldType: 'dropdown',
+                        fieldConfig: fieldConfig,
+                    };
+
+                    return field;
+                },
+                storageAdapter: {
+                    storeValue: (item, operationalValue: IVocabularyItem['qcode']) => {
+                        const vocabulary = allVocabularies.get(id);
+                        const vocabItems = vocabulary.items.filter((x) => operationalValue?.includes(x.qcode)) ?? [];
```
tomaskikutis: only return a boolean from a `.filter` callback. Nullish coalescing would be nicer here than optional chaining here.

client/components/planning-editor-standalone/profile-converter.ts
```diff
@@ -0,0 +1,45 @@
...
+    required: boolean;
+}
+
+interface ICustomVocabularyField extends IBaseField<'custom_vocabulary'> {
+    vocabularyId: string;
+}
+
+type IFieldConverted = IBaseField<'normal'> | ICustomVocabularyField;
+
+export const getProfileFieldsConverted = (): Array<IFieldConverted> => {
```
tomaskikutis: Avoid terminology(converting) if it's not helping significantly - because if you add terminology - you have to define it. Here I think it would be understandable enough to name it `getPlanningProfileFields`. You can add a comment about expansion of vocabulary fields.

client/components/planning-editor-standalone/field-adapters/custom-vocabularies.ts
thecalcc: Actually if you check the type of operationalValue it doesn't even make sense to have a fallback

thecalcc: I've doublechecked if that's actually the case and the operational value is never `null/undefined` and it is. Even when I clear the whole field the value is an empty array

client/components/planning-editor-standalone/field-adapters/custom-vocabularies.ts
```diff
@@ -0,0 +1,69 @@
...
+                        fieldType: 'dropdown',
+                        fieldConfig: fieldConfig,
+                    };
+
+                    return field;
+                },
+                storageAdapter: {
+                    storeValue: (item, operationalValue: IVocabularyItem['qcode']) => {
+                        const vocabulary = allVocabularies.get(id);
+                        const vocabItems = vocabulary.items.filter((x) => operationalValue.includes(x.qcode)) ?? [];
```
tomaskikutis: Default will never get applied here. If `.filter` does not crash it will always return an array.

client/components/planning-editor-standalone/profile.ts
```diff
@@ -4,17 +4,18 @@ import {
...
 } from 'superdesk-api';
-import {planningApi, superdeskApi} from '../../superdeskApi';
-import {getEditorFormGroupsFromProfile} from '../../utils/contentProfiles';
+import {superdeskApi} from '../../superdeskApi';
 import {
     IAttachmentsFieldConfig,
 } from '../../planning-extension/src/authoring-react-fields/planning-attachments/interfaces';
+import {getCustomVocabularyFields} from './field-adapters/custom-vocabularies';
+import {getPlanningProfileFields} from './profile-fields';
+import {IEventOrPlanningItem} from 'interfaces';
```
tomaskikutis: use only planning item interface here. Event will have a separate editor

tomaskikutis:
FYI it was pretty annoying to have the date PR together with vocabularies PR. Since you was refactoring a lot in this PR I would never review diff between commits, but a diff between target branch and this branch - so I saw how final changes look. I then had to look at date field multiple times when there were no changes there.

Next time spin up another branch from the PR branch to separate work on different features.

tomaskikutis (approved):

PR #2149, opened 2024-12-06
[SDESK-7444] - Planning: Migrate planning:flag_expired command to async

BrianMwangi21:
### Purpose

This PR updates the `planning:flag_expired command` command to async by: 

- Making use of the new async methods from events and planning to get and flag expired commands
- Updated tests

Solves [SDESK-7444](https://sofab.atlassian.net/browse/SDESK-7444)

[SDESK-7444]: https://sofab.atlassian.net/browse/SDESK-7444


server/planning/__init__.py
```diff
@@ -321,7 +326,9 @@ def init_scheduled_exports_task(app):
 
 @celery.task(soft_time_limit=600)
 def flag_expired():
-    FlagExpiredItems().run()
+    import asyncio
```
MarkLark86: You should be able to simply prefix the function definition with `async ` as we added async support to Celery tasks, and used in [Newshub](https://github.com/superdesk/newsroom-core/blob/async/newsroom/push/tasks.py#L31).

That way we don't have to worry about running `asyncio.run` or our own event loop. If this does't work here for some reason, then that's a different issue that needs to be looked at for the Planning module

MarkLark86 (commented):
Looks good, just some unit tests failing that needs to be looked into

server/planning/commands/flag_expired_items.py
```diff
@@ -33,170 +41,167 @@ class FlagExpiredItems(Command):
...
+    logger.info(f"{log_msg} {len(events_expired)} Events expired: {list(events_expired)}")
+
+
+async def flag_expired_planning(expiry_datetime: datetime):
+    log_msg = log_msg_context.get()
+    logger.info(f"{log_msg} Starting to flag expired planning items")
+    planning_service = PlanningAsyncService()
+
+    # Obtain the full list of Planning items that we're to process first
+    # As subsequent queries will change the list of returnd items
```
eos87: nit:
```suggestion
    # as subsequent queries will change the list of returned items
```

server/planning/commands/flag_expired_items.py
```diff
@@ -33,170 +41,167 @@ class FlagExpiredItems(Command):
...
+    if len(locked_plans) > 0:
+        logger.info(f"{log_msg} Skipping {len(locked_plans)} locked Planning items: {list(locked_plans)}")
+
+    if len(plans_expired) > 0:
+        push_notification("planning:expired", items=list(plans_expired))
+
+    logger.info(f"{log_msg} {len(plans_expired)} Planning items expired: {list(plans_expired)}")
+
+
+def set_event_plans(events):
```
eos87: Let's add the type for `events` param here

server/planning/commands/flag_expired_items.py
```diff
@@ -33,170 +41,167 @@ class FlagExpiredItems(Command):
...
+    for plan in get_related_planning_for_events(list(events.keys()), "primary"):
+        for related_event_id in get_related_event_ids_for_planning(plan, "primary"):
+            event = events[related_event_id]
+            if "_plans" not in event:
+                event["_plans"] = []
+            event["_plans"].append(plan)
+
+
+def get_event_schedule(event):
```
eos87: also here and the return type

server/planning/commands/flag_expired_items_test.py
```diff
@@ -37,40 +39,44 @@
 
 
 class FlagExpiredItemsTest(TestCase):
+    app_config = {
+        **TestCase.app_config.copy(),
+        # Expire items that are scheduled more than 24 hours from now
+        "PLANNING_EXPIRY_MINUTES": 1440,
```
eos87: nit 😅 : 
```suggestion
        "PLANNING_EXPIRY_MINUTES": 24 * 60,
```

server/planning/commands/flag_expired_items_test.py
```diff
@@ -12,12 +12,14 @@
...
 from superdesk import get_resource_service
 from superdesk.utc import utcnow
 
 from planning.tests import TestCase
 from planning.types import PlanningRelatedEventLink
-from .flag_expired_items import FlagExpiredItems
+from .flag_expired_items import flag_expired_items_handler
 
 now = utcnow()
 yesterday = now - timedelta(hours=48)
```
eos87: It seems more like two days than yesterday 😉 
```suggestion
two_days_ago = now - timedelta(hours=48)
```

server/planning/planning/service.py
```diff
@@ -17,9 +17,17 @@ class PlanningAsyncService(BasePlanningAsyncService[PlanningResourceModel]):
...
+        have no future schedules or primary-linked events, and are not already expired.
+
+        By default, items are filtered to exclude:
+        - Items linked to a primary event or,
+        - Items already expired or,
+        - Items with future scheduling or a planning_date beyond `expiry_datetime`.
 
-        Where planning_date is in the past
+        If `spiked_planning_only` is True, only spiked items are returned, still excluding
+        those with future schedules or planning_dates.
```
eos87, line 30: 👏

server/planning/events/service.py
```diff
@@ -0,0 +1,70 @@
...
+                    "filter": {"range": {"dates.end": {"lte": date_to_str(expiry_datetime)}}},
+                },
+            },
+            "sort": [{"dates.start": "asc"}],
+            "size": get_max_recurrent_events(),
+        }
+
+        if spiked_events_only:
+            del query["query"]["bool"]["must_not"]
+            query["query"]["bool"]["must"] = [{"term": {"state": WORKFLOW_STATE.SPIKED}}]
```
eos87: in here I would suggest we use the new Enum class from `types.enums.WorkflowState` as it is being used by the model too:
```suggestion
            query["query"]["bool"]["must"] = [{"term": {"state": WorkflowState.SPIKED}}]
```

eos87 (commented):
It looks good, I just left a few comments. I can see these functions rely much on dictionaries. In general, we should lean into using the pydantic models as much as possible.

eos87 (approved):
Looks good to me!

eos87:
@BrianMwangi21 there are some broken references to old `events.service.py`. Please check.

```
from planning.events.service import EventsAsyncService
E   ModuleNotFoundError: No module named 'planning.events.service'
```

server/planning/commands/flag_expired_items_test.py
```diff
@@ -170,25 +173,24 @@ async def test_planning(self):
...
+            await self.assertExpired(
                 "planning",
                 {
                     "p1": False,
                     "p2": False,
                     "p3": False,
                     "p4": False,
                     "p5": True,
-                    "p6": False,
+                    "p6": True,
```
MarkLark86, line 186: I'm wondering, why were `p6` and `p8` changed in the tests from `False` to `True`. Is there any logic or query that is not the same as the old code?

server/planning/commands/flag_expired_items_test.py
```diff
@@ -213,25 +214,23 @@ async def test_event_with_single_planning_no_coverages(self):
...
                 ],
             )
-            FlagExpiredItems().run()
-
-            self.assertExpired("events", {"e1": False, "e2": False, "e3": False, "e4": True})
-
-            self.assertExpired("planning", {"p1": False, "p2": False, "p3": False, "p4": True})
+            await flag_expired_items_handler()
+            await self.assertExpired("events", {"e1": False, "e2": False, "e3": False, "e4": True})
+            await self.assertExpired("planning", {"p1": False, "p2": False, "p3": True, "p4": True})
```
MarkLark86, line 230: Also here as well, why is `p3` now `True`

MarkLark86:
Looks good @BrianMwangi21 :+1: 
Just left a couple small comments, mostly about the changes in the test data. Besides that, everything looks good.

MarkLark86:
@BrianMwangi21 Looks like the changes to the test data is required, as the new async Planning service doesn't yet have all the functionality from the original service (which involved processing the Coverages and dates etc), which would be why these tests were failing before your test condition changes.

We can look at this later once the async Planning service is updated

BrianMwangi21:
@MarkLark86 this is noted. I've added a TODO on the test file on the same.

PR #2150, opened 2024-12-08
[SDESK-7442] Migrate Events resource service to async

eos87:
### Purpose
Extend the new Event async resource service with the next methods
- `on_create` 
- `create` 
- `on_created`
- `validate_event` and related private methods
- `on_update` 
- `update`
- `on_updated`
- `on_deleted`

### What has changed
- The methods mentioned above have been migrated from old `sync` service to new async one and the corresponding code has been adjusted as much as possible to be async (along with the depending internal/private methods)
- Added new `events_utils` module in order to make some of the functions reusable
- Adjusted related test to run

### What's pending
- Move all the tests that have references to `get_resource_service("events")`. There was not enough time to do so, although some tests were already adjusted and fixed.
- Some tests are broken because of an issue with `prod_api` on superdesk-core (`AttributeError: __enter__`) produced here 
https://github.com/superdesk/superdesk-core/blob/88d2543a841896731725cb5f2e5cd2b7b34d2a6a/prod_api/conftest.py#L53-L55 
(most likely because `prodapi_app.app_context()` is not awaited)

> [!NOTE]  
> I will go through the docstrings of these changes and update them. If there is not time before sprint end, I will provide another small PR for that. 

Thanks for reviewing!

Resolves: [SDESK-7442]

[SDESK-7442]: https://sofab.atlassian.net/browse/SDESK-7442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+        """
+        Reset fields that are not required by the new (recurring) events
+        """
+        fields_to_reset = ["lock_user", "lock_time", "lock_session", "lock_action"]
+
+        for field in fields_to_reset:
+            setattr(event, field, None)
+
+    def _generate_recurring_events(
+        self, event: EventResourceModel, recurrence_id: int | None = None
```
MarkLark86: I think `recurrence_id` should be a string here

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+        # set a limit to prevent too many events to be created
+        for date in itertools.islice(recurring_dates, 0, max_recurring_events):
+            # prepare data for new recurring event
+            new_id = generate_guid(type=GUID_NEWSML)
+            recurring_event_updates = {"dates": dict(start=date, end=(date + time_delta)), "guid": new_id, "id": new_id}
+
+            if not recurrence_id:
+                recurring_event_updates["recurrence_id"] = new_id
+
+            # reset fields not required by new events
```
MarkLark86, line 740: We also need to have special handling for the first Event with `embedded_planning` defined, as this auto-generates associated Planning item with Coverages to the Event.
```python
if key == "embedded_planning":
    if not embedded_planning_added:
        # If this is the first Event in the series, then keep
        # the ``embedded_planning`` field for processing later
        embedded_planning_added = True
    else:
        # Otherwise remove the ``embedded_planning`` from all other Events
        # in the series
        new_event.pop("embedded_planning")
```

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+
+        for doc in docs:
+            event_id = doc.id
+
+            # If we duplicated this event, update the history
+            if doc.duplicate_from:
+                parent_id = doc.duplicate_from
+                parent_event = await self.find_by_id(parent_id)
+
+                assert parent_event is not None
```
MarkLark86: Ideally we should raise an exception instead of using assert statements (in this context, this would be a bad request error, as the data supplied is invalid - item duplicated from not found)

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+
+            # Don't send notification if one has already been sent
+            # This is to ensure recurring events don't send multiple notifications
+            if event_id in notifications_sent or doc.previous_recurrence_id:
+                continue
+
+            notifications_sent.append(event_id)
+            push_notification(event_type, item=event_id, user=user_id)
+
+    async def on_update(self, updates: dict[str, Any], original: EventResourceModel):
```
MarkLark86, line 237: `on_updated` seems to be missing from this service, it sends websocket notification, among other side effects.

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+
+        # And finally push a notification to connected clients
+        push_notification(
+            "events:updated:recurring",
+            item=str(original[ID_FIELD]),
+            recurrence_id=str(original.recurrence_id),
+            user=str(updates.get("version_creator", "")),
+        )
+
+    def _get_recurring_timeline(self, selected: dict[str, Any], spiked: bool = False):
```
MarkLark86: Brian has added this as a [util function](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/events/utils.py#L35), not sure if we need this on the service anymore

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+            # And update the start/end dates to be in line with the new recurring rules
+            updates["dates"]["start"] = updated_event.dates.start
+            updates["dates"]["end"] = updated_event.dates.end
+            updates["_planning_schedule"] = [x.to_dict() for x in self._create_planning_schedule(updated_event)]
+            remove_lock_information(item=updates)
+
+        # Create the new events and generate their history
+        await self.create(generated_events)
+        app = get_current_app().as_any()
+        app.on_inserted_events(generated_events)
```
MarkLark86: It would be good to set this up as a [signal](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/signals.py) instead, using the [newer signals](https://superdesk.readthedocs.io/en/async/core/signals.html)

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+    @staticmethod
+    def _link_to_planning(event: EventResourceModel):
+        """
+        Links an Event to an existing Planning Item
+
+        The Planning item remains locked, it is up to the client to release this lock
+        after this operation is complete
+        """
+        # TODO-ASYNC: replace when planning service is async
+        planning_service = get_resource_service("planning")
```
MarkLark86: The Planning service should already be available. The only functionality I see missing is the `validate_on_update` method. Might be good to start using the async service, but comment out `planning_service.validate_on_update(updates, planning_item, get_user())` with a TODO-ASYNC comment

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+        if event.recurrence_id:
+            related_planning["recurrence_id"] = event.recurrence_id
+            if not planning_item.get("recurrence_id") and link_type == "primary":
+                updates["recurrence_id"] = event.recurrence_id
+
+        planning_service.validate_on_update(updates, planning_item, get_user())
+        planning_service.system_update(plan_id, updates, planning_item)
+
+        app = get_current_app().as_any()
+        app.on_updated_planning(updates, planning_item)
```
MarkLark86: And again here for using the newer signal system

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+
+        # Determine if we're to convert this single event to a recurring of events
+        if (
+            original.lock_action == "convert_recurring"
+            and updates.get("dates", {}).get("recurring_rule", None) is not None
+        ):
+            generated_events = await self._convert_to_recurring_event(updates, original)
+
+            # if the original event was "posted" then post all the generated events
+            # if original.get("pubstatus") in [ POST_STATE.CANCELLED, POST_STATE.USABLE]:
```
MarkLark86: I'm guessing this commented line no longer needs to be here

server/planning/events/events_service.py
```diff
@@ -0,0 +1,761 @@
...
+                self.mark_event_complete(updates, ev, mark_complete_validated)
+                # It is validated if the previous funciton did not raise an error
+                mark_complete_validated = True
+
+            # Remove ``embedded_planning`` before updating this event, as this should only be handled
+            # by the event provided to this update request
+            new_updates.pop("embedded_planning", None)
+
+            app = get_current_app().as_any()
+            app.on_updated_events(new_updates, {"_id": event_id})
```
MarkLark86: And once more here for improving signals

MarkLark86 (commented):
Looks good, just a couple small suggestions

server/planning/events/events_service.py
eos87: Done.

BrianMwangi21 (commented):
LGTM!

server/planning/events/events_service.py
eos87: done

BrianMwangi21 (approved):

eos87:
I will fix the test in a subsequent PR.

PR #2151, opened 2024-12-09
Add Coverage Provider to Embedded Planning and Synchronize Coverage Assignee Updates to Assignments [SDCP-875]

devketanpro:
SDCP-875

### DIFF
--- a/server/planning/planning/planning.py
+++ b/server/planning/planning/planning.py
@@ -1012,7 +1012,7 @@ class PlanningService(superdesk.Service):
                 assignment["name"] = planning["name"] if not translated_value and translated_name else translated_name
 
             # If the coverage assignee has been changed and workflow status is active
-            if original.get("workflow_status") == WORKFLOW_STATE.ACTIVE and self.is_coverage_assignment_modified(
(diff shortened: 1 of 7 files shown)


server/planning/planning/planning.py
```diff
@@ -1011,6 +1011,13 @@ def _create_update_assignment(
             if planning_original.get("name") != planning_updates.get("name"):
                 assignment["name"] = planning["name"] if not translated_value and translated_name else translated_name
 
+            # If the coverage assignee has been changed and workflow status is active
+            if original.get("workflow_status") == WORKFLOW_STATE.ACTIVE and self.is_coverage_assignment_modified(
```
MarkLark86: I'm wondering if this should be excluding states instead of explicitly checking for 1.
`original.get("workflow_status") != WorkflowStates.DRAFT` might suffice for the state checking here

MarkLark86 (commented):
@devketanpro Looks good

server/planning/planning/planning.py
devketanpro: let me check

MarkLark86 (approved):

PR #2153, opened 2024-12-10
[SDESK-7442] Fix and improve tests

eos87:
### Purpose
To fix the tests that are related to `EventsAsyncService`. 

### What has changed
- Added a helper method (`clone_with`) in `EventResourceModel` to facilitate the cloning and deep merging of a model instance and a given dict of updates (we can consider moving this base model later on if needed)
- Adjusted `EventsAsyncService` `create` and `update` methods to match the same functionality, as the legacy service uses `patch` or `post` method to bypass the hooks. Instead, in the async service, I've abstracted the logic, which is bypassed in the legacy service, to explicitly use it so it is easier to follow and to understand.
- Updated `events_tests.py` to use new `EventsAsyncService` 
- Improved test case so we don't have to repeat `async with self.app.app_context():` all over the tests

### What's pending
- There are still references to old events service but we will move them as we move forward
- prod_api tests are still broken but I will take a look at those in [SDESK-7456]
[SDESK-7442]

Thanks for checking!

[SDESK-7456]: https://sofab.atlassian.net/browse/SDESK-7456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDESK-7442]: https://sofab.atlassian.net/browse/SDESK-7442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/tests/__init__.py
```diff
@@ -17,3 +17,15 @@ def setup_test_user(self):
         user = {"_id": ObjectId()}
         self.app.data.insert("users", [user])
         g.user = user
+
+    async def setUp(self):
```
MarkLark86: If a test cases setup is async, then you use the `async def asyncSetUp(self):` method instead (and make sure to call `await super().asyncSetUp()` as well.

As for the app_context, I thought this was already being done on the [AsyncFlaskTestCase](https://github.com/superdesk/superdesk-core/blob/async/superdesk/tests/__init__.py#L643) from core?

server/planning/types/event.py
```diff
@@ -272,3 +275,21 @@ class EventResourceModel(BasePlanningModel, LockFieldsMixin):
 
     related_items: list[RelatedItem] = Field(default_factory=list)
     failed_planned_ids: list[str] = Field(default_factory=list)
+
+    def clone_with(self, updates: dict[str, Any]) -> Self:
```
MarkLark86: I think this should definitely go into core

server/planning/tests/__init__.py
eos87: you're right. For some reason, it was failing before, because of the missing app context but, I guess we were inheriting from the wrong base class.

server/planning/types/event.py
eos87: I'll move it

eos87: done

server/planning/tests/__init__.py
eos87: Done.

MarkLark86 (approved):

PR #2154, opened 2024-12-11
allow creation of events without start/end time

petrjasek:
STT-40

## Front-end checklist


client/components/UI/Form/TimeInput/TimeInputPopup.tsx
```diff
@@ -10,22 +9,41 @@ import {Button} from '../../';
 
 import './style.scss';
 
+interface IProps {
+    value: any;
+    onChange(value: string): void;
```
thecalcc: Probably `value` type can be improved here with `string | null` since on line 101 you're passing `null`

client/components/UI/Form/TimeInput/TimeInputPopup.tsx
```diff
@@ -64,7 +82,7 @@ export class TimeInputPopup extends React.Component {
         const {onChange, close} = this.props;
 
         if (addMinutes) {
-            let newTime = moment.clone(this.state.currentTime);
+            let newTime = moment(this.state.currentTime);
```
thecalcc: can be `const`

client/components/fields/editor/EventSchedule.tsx
```diff
@@ -51,53 +49,58 @@ export class EditorFieldEventSchedule extends React.PureComponent<IProps> {
...
+            const changes = {
+                _startTime: null,
+                _endTime: null,
+                'dates.start': startDate ? localDateToUtc(startDate, this.props.item?.dates?.tz) : null,
+                'dates.end': endDate ? localDateToUtc(endDate, this.props.item?.dates?.tz) : null,
+                'dates.all_day': true,
+                'dates.no_end_time': false,
+            };
+
+            this.props.onChange(changes, null);
```
thecalcc: A second argument `null` is passed to onChange always here, so if it's easy could you remove that?

client/components/fields/editor/EventSchedule.tsx
```diff
@@ -107,41 +110,57 @@ export class EditorFieldEventSchedule extends React.PureComponent<IProps> {
...
         if (!value) {
-            this.props.onChange('_endTime', null);
+            const hasStartTime = this.props.item._startTime != null;
+            const changes = {
+                _endTime: null,
+                'dates.end': this.props.item.dates?.end ?
+                    localDateToUtc(this.props.item.dates.end, this.props.item.dates.tz)
+                    : null,
+                'dates.all_day': !hasStartTime,
+                'dates.no_end_time': hasStartTime,
```
thecalcc, line 149: So with this logic if there's a start time there has to also be end time and vise versa? No option for one without the other?

client/components/fields/editor/EventSchedule.tsx
```diff
@@ -297,3 +270,24 @@ export class EditorFieldEventSchedule extends React.PureComponent<IProps> {
         );
     }
 }
+
+/**
+ * Combines date from the first argument with time from the second argument
+ */
+function combineDateTime(date: moment.MomentInput, time: moment.Moment, tz?: string): moment.Moment {
+    return moment.tz(moment(date).format('YYYY-MM-DD'), tz) // we only want the date part
```
thecalcc, line 282: Format is always `YYYY-MM-DD` on purpose, and we format based on format config later or it's a mistake here?

client/components/fields/editor/EventSchedule.tsx
```diff
@@ -297,3 +270,24 @@ export class EditorFieldEventSchedule extends React.PureComponent<IProps> {
...
+}
+
+/**
+ * Converts datetime in given timezone to date only in utc
+ * eg. 2024-12-10T23:00:00-05:00 -> 2024-12-11T00:00:00Z
+ */
+function localDateToUtc(date: moment.MomentInput, tz?: string): moment.Moment {
+    return moment.utc(tz ?
+        moment.tz(date, tz).format('YYYY-MM-DD') :
+        moment(date).format('YYYY-MM-DD')
```
thecalcc, line 295: Same here

client/components/fields/editor/base/dateTime.tsx
```diff
@@ -43,10 +44,8 @@ export class EditorFieldDateTime extends React.PureComponent<IProps> {
     render() {
         const field = this.props.field;
         const value = get(this.props.item, field, this.props.defaultValue);
-        const momentValue = value != null ?
-            moment(value) :
-            null;
-        const error = get(this.props.errors ?? {}, field);
+        const momentValue = value != null ? moment(value) : undefined; // use moment.utc?
+        const error = get(this.props.errors ?? {}, field) || undefined;
```
thecalcc: Redundancy here - if `get` has to be used might be better to do 
`get(this.props.errors, field, undefined)` if value isn't found, fall back value is returned, in this case would be `undefined`

e2e/cypress/e2e/events/event_all_day.cy.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    it('can create single day event without start and end time', () => {
+        const event = {
+            ...baseEvent, 
+            'dates.start.date': moment().format(CLIENT_FORMAT),
+        };
+
+        const expectedEvent = {
+            ...event,
+            'dates.end.date': event['dates.start.date'],
```
thecalcc, line 47: should there be validation if an event starts & ends at the same time or we automatically treat it as a single day event? Maybe if there's time and start & end have the same values that's invalid?

thecalcc (approved):

client/components/fields/editor/EventSchedule.tsx
petrjasek: there can be no time - `all_day` is `true`, or no end time - `no_end_time` is `true`, or there must be both.
we can't have just end time

client/components/fields/editor/EventSchedule.tsx
petrjasek: here it's on purpose to only take the date from timestamp and ignore the time

e2e/cypress/e2e/events/event_all_day.cy.ts
petrjasek: in general that's ok, there can be same values. we don't need to avoid it. but if there is a usecase that's another question :smile:

thecalcc (approved):

PR #2156, opened 2024-12-12
Refactor structure & categories field

thecalcc:
STT-63

## Front-end checklist


client/components/planning-editor-standalone/field-definitions/place-field.ts
```diff
@@ -0,0 +1,38 @@
+import {IDropdownConfigVocabulary, IAuthoringFieldV2, IVocabularyItem} from 'superdesk-api';
+import {superdeskApi} from '../../../superdeskApi';
+import {IFieldGetter} from '.';
+
+export const getPlaceField: IFieldGetter = () => ({
```
tomaskikutis: `IFieldGetter` doesn't server any purpose here compared to
`const getPlaceField = (): IFieldDefinition => {...}`

client/components/planning-editor-standalone/profile.ts
```diff
@@ -1,220 +1,8 @@
...
-                    name: gettext('Coverages'),
-                    fieldType: 'coverages',
-                    fieldConfig: fieldConfig,
-                };
-
-                return field;
-            },
-        },
-        {
-            fieldId: 'agendas',
```
tomaskikutis, line 178: Should this be present in the refactored code too?

thecalcc: Yes, I've deleted it on accident

tomaskikutis (approved):

PR #2157, opened 2024-12-13
[SDESK-7461] - Assignment Delivery async resource & service

BrianMwangi21:
### Purpose

This PR creates a new async resource and service for Assignment Delivery and registers it to the planning module.

Solves [SDESK-7461](https://sofab.atlassian.net/browse/SDESK-7461)

[SDESK-7461]: https://sofab.atlassian.net/browse/SDESK-7461

### DIFF
--- a/server/planning/assignments/module.py
+++ b/server/planning/assignments/module.py
@@ -5,8 +5,9 @@ from superdesk.core.resources import (
     ElasticResourceConfig,
 )
 
-from planning.types import AssignmentResourceModel
+from planning.types import AssignmentResourceModel, DeliveryResourceModel
 from .service import AssignmentsAsyncService
+from .delivery_service import DeliveryAsyncService
 
 assignments_resource_config = ResourceConfig(
     name="assignments",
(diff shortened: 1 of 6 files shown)


eos87 (approved):
Looks good

server/planning/types/delivery.py
```diff
@@ -0,0 +1,18 @@
+from pydantic import Field
+from typing import Annotated
+from datetime import datetime
+
+from superdesk.core.resources import fields
+from superdesk.core.resources.validators import validate_data_relation_async
+from .base import BasePlanningModel
+
+
+class DeliveryResourceModel(BasePlanningModel):
```
MarkLark86: The `BasePlanningModel` adds `original_creator` and `version_creator` to the model. Delivery resources are managed internally, and therefor do not have any user creator information (see the original schema).

Because of this, you could use core resource & service directly instead.
i.e.

```python
# file: planning/types/delivery.py
from superdesk.core.resources import ResourceModelWithObjectId

class DeliveryResourceModel(ResourceModelWithObjectId):
    ...

# file: planning/assignments/delivery_service.py
from superdesk.core.resources.service import AsyncResourceService
from planning.types import DeliveryResourceModel

class DeliveryAsyncService(AsyncResourceService[DeliveryResourceModel]):
    ...
```

server/planning/assignments/module.py
```diff
@@ -34,3 +35,34 @@
...
+delivery_resource_config = ResourceConfig(
+    name="delivery",
+    data_class=DeliveryResourceModel,
+    service=DeliveryAsyncService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(
+                name="planning_id_1",
+                keys=[("planning_id", 1)],
+                background=True,
```
MarkLark86, line 48: All 4 of these indexes should also have `unique=False` added to them

server/planning/assignments/module.py
```diff
@@ -34,3 +35,34 @@
...
+            ),
+            MongoIndexOptions(
+                name="item_id_1",
+                keys=[("item_id", 1)],
+                background=True,
+            ),
+        ],
+    ),
+    elastic=ElasticResourceConfig(),
```
MarkLark86: Elasticsearch is not enabled for this resource (we only use Mongo for this one).

One way you can tell if Elastic is enabled for a resource, in the old resource class we have a `datasource` dict attribute that has `"search_backend": "elastic"` defined ([example](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/assignments/assignments.py#L1531)).

If the `search_backend` attribute is not defined, or not set to `elastic`, then Elasticsearch not is enabled for that resource

MarkLark86 (commented):
@BrianMwangi21 Just a couple small changes suggested, then we can merge this one

MarkLark86 (approved):

PR #2158, opened 2024-12-13
Subject field

thecalcc:
STT-63

## Front-end checklist


client/components/planning-editor-standalone/field-definitions/subject.ts
```diff
@@ -0,0 +1,63 @@
...
+
+                const allSubjects: Array<ISubjectCode> = (planningApi.redux.store.getState().subjects ?? [])
+                    .filter((x) => operationalValue.includes(x.qcode));
+
+                return {
+                    ...item,
+                    subject: allSubjects.map(({qcode, name, parent}) => {
+                        var itemToStore: IStorageFormat = {qcode, name, parent};
+
+                        if (parent != null) {
```
tomaskikutis: When would this condition make a difference?

```
const a = {x: 1}
const b = {...a}

if (b.x != null) { // <-- would never make a difference because if a.x and b.x are the same
    b.x = a.x;
}
```

thecalcc: Doesn't, I had it copied over from the subject in client-core

client/components/planning-editor-standalone/field-definitions/subject.ts
```diff
@@ -36,26 +36,12 @@ export function getSubjectField(): IFieldDefinition {
...
-                    subject: allSubjects.map(({qcode, name, parent}) => {
-                        var itemToStore: IStorageFormat = {qcode, name, parent};
-
-                        if (parent != null) {
-                            itemToStore.parent = parent;
-                        }
-
-                        return itemToStore;
-                    }),
+                    subject: allSubjects,
```
tomaskikutis: `allSubjects` is confusing naming, because if you filter it, you can't call it "all" anymore

tomaskikutis (approved):

client/components/planning-editor-standalone/field-definitions/subject.ts
```diff
@@ -36,12 +36,12 @@ export function getSubjectField(): IFieldDefinition {
                 return (item.subject ?? []).map(({qcode}) => qcode);
             },
             storeValue: (item, operationalValue: Array<ISubjectCode['qcode']>) => {
-                const allSubjects: Array<ISubjectCode> = (planningApi.redux.store.getState().subjects ?? [])
+                const filteredSubjects: Array<ISubjectCode> = (planningApi.redux.store.getState().subjects ?? [])
```
tomaskikutis: `filteredSubjects` isn't much better. It doesn't make sense to mention programming operations (filter,map,reduce etc.) that you performed - but what is being achieved by those programming operations. When it's something small and obvious like here I'd call it `subjects` simply without elaborating. If you wanted more details, then maybe `subjectsFull` since the purpose of the operation is to return full objects instead of qcodes only.

tomaskikutis (approved):

PR #2162, opened 2024-12-17
[SDESK-7462] - Events, Planning & History resource, service and REST API

BrianMwangi21:
### Purpose
This PR creates new async events and planning history resources and services and exposes their rest API by :

- Implementing a base history service async resource service class
- Implementing Event History resource model, resource config and service
- Implementing Planning History resource model, resource config and service
- Registering these new resources with the Planning module

Solves: [SDESK-7462](https://sofab.atlassian.net/browse/SDESK-7462)

[SDESK-7462]: https://sofab.atlassian.net/browse/SDESK-7462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/types/event_history.py
```diff
@@ -0,0 +1,11 @@
+from typing import Annotated, Any
+
+from superdesk.core.resources import ResourceModelWithObjectId, fields
+from superdesk.core.resources.validators import validate_data_relation_async
+
+
+class EventsHistoryResourceModel(ResourceModelWithObjectId):
```
MarkLark86: I think these 2 resources could be defined in the one source file. Also We could also use a common base model for these two, as `user_id`, `operation` and `update` are the same across both resources.

server/planning/history_async_service.py
```diff
@@ -0,0 +1,119 @@
...
+    "planning_ids",
+    "_updates_schedule",
+    "_planning_schedule",
+    "_planning_date",
+    "_reschedule_from_schedule",
+    "versioncreated",
+]
+
+
+class HistoryAsyncService(AsyncResourceService):
```
MarkLark86: Resource type missing with AsyncResourceService (this provides resource model type when using this service)

server/planning/history_async_service.py
```diff
@@ -0,0 +1,119 @@
...
+        original_keys = set(original.keys())
+        updates_keys = set(updates.keys())
+        intersect_keys = original_keys.intersection(updates_keys)
+        modified = {o: updates[o] for o in intersect_keys if original[o] != updates[o]}
+        added_keys = updates_keys - original_keys
+        added = {a: updates[a] for a in added_keys}
+        modified.update(added)
+        return await self._remove_unwanted_fields(modified)
+
+    async def _remove_unwanted_fields(self, update: dict[str, Any]):
```
MarkLark86: I'm not sure if this function needs to be async, as there is no I/O to await in the body

server/planning/events/module.py
```diff
@@ -28,3 +30,10 @@
     ),
     elastic=ElasticResourceConfig(),
 )
+
+events_history_resource_config = ResourceConfig(
+    name="events_history",
+    data_class=EventsHistoryResourceModel,
+    service=EventsHistoryAsyncService,
+    rest_endpoints=RestEndpointConfig(resource_methods=["GET"], item_methods=["GET"]),
```
MarkLark86, line 47: If we're going to enable the REST API endpoints, then we should probably add `internal_resource = True` to the `EventsHistoryResource` and `PlanningHistoryResource` legacy resources, otherwise they'll both be trying to register the same endpoint with Flask

server/planning/planning/module.py
```diff
@@ -24,3 +26,10 @@
     ),
     elastic=ElasticResourceConfig(),
 )
+
+planning_history_resource_config = ResourceConfig(
```
MarkLark86, line 30: I know it's not in the original, but we should probably have an index for the Item ID for this and Event history resources.

Example PlanningHistory index:
```python
mongo=MongoResourceConfig(
    indexes=[
        MongoIndexOptions(
            name="planning_id",
            keys=[("planning_id", 1)],
            unique=False
        ),
    ],
)
```

server/planning/planning/planning_history_async_service.py
```diff
@@ -0,0 +1,203 @@
...
+
+        return diff
+
+    async def _save_coverage_history(self, updates: dict[str, Any], original):
+        """Save the coverage history for the planning item"""
+        item = deepcopy(original)
+        original_coverages = {c.get("coverage_id"): c for c in (original or {}).get("coverages") or []}
+        updates_coverages = {c.get("coverage_id"): c for c in (updates or {}).get("coverages") or []}
+        added, deleted, updated = [], [], []
+        planning_service = get_resource_service("planning")
```
MarkLark86: Looks like the old service is used here, for the [is_coverage_planning_modified](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/planning/planning.py#L1251) and [is_coverage_assignment_modified](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/planning/planning.py#L1265).

These two functions are quite small, and can probably be placed in a util file somewhere and used here (so we are not using the old legacy service)

server/planning/types/planning_history.py
```diff
@@ -0,0 +1,11 @@
+from typing import Annotated, Any
+
+from superdesk.core.resources import ResourceModelWithObjectId, fields
+from superdesk.core.resources.validators import validate_data_relation_async
+
+
+class PlanningHistoryResourceModel(ResourceModelWithObjectId):
+    planning_id: Annotated[fields.ObjectId, validate_data_relation_async("planning")]
```
MarkLark86: Event and Planning IDs are pure strings, not ObjectId.

MarkLark86 (commented):
It would also be good if you can add types to function signatures, which enables type checking by mypy and autocomplete/error checking from within IDs

MarkLark86:
Good job @BrianMwangi21 Just a few small things

MarkLark86:
There are mypy/type errors, please check those out as well

server/planning/events/events_history_async_service.py
```diff
@@ -0,0 +1,83 @@
...
+from superdesk.resource_fields import ID_FIELD
+from planning.utils import get_related_planning_for_events
+from planning.history_async_service import HistoryAsyncService
+from planning.item_lock import LOCK_ACTION
+
+logger = logging.getLogger(__name__)
+
+
+class EventsHistoryAsyncService(HistoryAsyncService[EventsHistoryResourceModel]):
+    async def on_item_created(self, items: list[EventResourceModel | Any], operation: str | None = None):
```
MarkLark86: Instead of using `Any`, it might be good to use `dict[str, Any]`

server/planning/history_async_service.py
```diff
@@ -0,0 +1,122 @@
...
+    "_planning_date",
+    "_reschedule_from_schedule",
+    "versioncreated",
+]
+
+
+class HistoryAsyncService(AsyncResourceService[Generic[HistoryResourceModelType]]):
+    """Provide common async methods for tracking history of Creation, Updates and Spiking to collections"""
+
+    async def on_item_created(self, items: list[Any], operation: str | None = None):
```
MarkLark86: Same here, use `list[dict[str, Any]]` instead of `list[Any]`

MarkLark86 (commented):
Looks good, just a couple improvements to the types. Please check wherever you're using just `Any` and replace with `dict[str, Any]`, so type checking/IDE knows the type of the variable

server/planning/events/events_history_async_service.py
```diff
@@ -0,0 +1,83 @@
...
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+"""Superdesk Files"""
+
+from copy import deepcopy
+import logging
+from typing import Any
```
eos87: nit:
```suggestion
import logging

from copy import deepcopy
from typing import Any
```

server/planning/events/events_service.py
```diff
@@ -1,3 +1,4 @@
+from planning.events.events_history_async_service import EventsHistoryAsyncService
```
eos87: Let's move this one down under the other `from ...` imports.

server/planning/history_async_service.py
```diff
@@ -0,0 +1,122 @@
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+"""Superdesk Files"""
+
```
eos87: nit: 
```suggestion
```

server/planning/planning/planning_history_async_service.py
```diff
@@ -0,0 +1,207 @@
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+"""Superdesk Files"""
+
```
eos87:
```suggestion
```

server/planning/planning/planning_history_async_service.py
```diff
@@ -0,0 +1,207 @@
...
+logger = logging.getLogger(__name__)
+update_item_actions = ["assign_agenda", "add_featured", "remove_featured"]
+
+
+class PlanningHistoryAsyncService(HistoryAsyncService[PlanningHistoryResourceModel]):
+    async def on_item_created(self, items: list[dict[str, Any]], operation=None):
+        add_to_planning = False
+        if request and hasattr(request, "args"):
+            add_to_planning = strtobool(request.args.get("add_to_planning", "false"))
```
eos87, line 29: Is there any way to avoid coupling with `flask.request` here?

server/planning/planning/planning_history_async_service.py
```diff
@@ -0,0 +1,207 @@
...
+        await super().on_item_created(items, "add_to_planning" if add_to_planning else None)
+
+    async def _save_history(self, item, update: dict[str, Any], operation: str | None = None):
+        user = await self.get_user_id()
+        # confirmation could be from external fulfillment, so set the user to the assignor
+        if operation == ASSIGNMENT_HISTORY_ACTIONS.CONFIRM and user is None:  # type: ignore
```
eos87: You can convert/create an enum for `ASSIGNMENT_HISTORY_ACTIONS`, use it to type the parameter `operation` and then avoid the `# type: ignore` here.

server/planning/planning/planning_history_async_service.py
```diff
@@ -0,0 +1,207 @@
...
+        await self._save_history(
+            {ID_FIELD: str(parent[ID_FIELD])},
+            {"duplicate_id": str(duplicate[ID_FIELD])},
+            "duplicate",
+        )
+
+    async def on_duplicate_from(self, item: dict[str, Any], duplicate_id: str):
+        new_plan = deepcopy(item)
+        new_plan["duplicate_id"] = duplicate_id
+        await self._save_history({ID_FIELD: str(item[ID_FIELD])}, new_plan, "duplicate_from")
```
eos87, line 186: It would be good to add some docstrings to at least some of these methods to clarify their purposes.

eos87 (commented):
I left a few minor comments.

MarkLark86 (approved):
Looks good to me

PR #2165, opened 2024-12-18
Show in embedded editor field config option

thecalcc:
STT-76

## Front-end checklist

### DIFF
--- a/client/components/planning-editor-standalone/profile-fields.ts
+++ b/client/components/planning-editor-standalone/profile-fields.ts
@@ -23,8 +23,9 @@ const unimplementedFields = new Set<string>([
 
 /**
  * A function that handles planning profile field types so they can be used in authoring react.
+ * @embeddedOnly defaults to false
  */
-export const getPlanningProfileFields = (): Array<IFieldConverted> => {
+export const getPlanningProfileFields = (options: {embeddedOnly?: boolean}): Array<IFieldConverted> => {
     const planningProfile = planningApi.contentProfiles.get('planning');
     const planningGroups = getEditorFormGroupsFromProfile(planningProfile);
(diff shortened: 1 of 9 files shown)


thecalcc:
- [x] TODO: Make this field config option `readOnly={true}` & toggled on if `required` config option is set to true 

Note: Since we haven't yet worked on the embedded event editor the condition: [`
For each field in the Event form and Planning item form manager there will be a checkboc “Show in embedded form” next to the checkbox “Required“`](https://sofab.atlassian.net/browse/STT-76#:~:text=For%20each%20field%20in%20the%20Event%20form%20and%20Planning%20item%20form%20manager%20there%20will%20be%20a%20checkboc%20%E2%80%9CShow%20in%20embedded%20form%E2%80%9D%20next%20to%20the%20checkbox%20%E2%80%9CRequired%E2%80%9C) has to be implemented there as well once we have started working on that

client/components/planning-editor-standalone/profile-fields.ts
```diff
@@ -35,6 +35,14 @@ export const getPlanningProfileFields = (): Array<IFieldConverted> => {
     for (const fieldId of planningFieldIds) {
         const fieldSchema = planningProfile.schema[fieldId];
 
+        /**
+         * If a field does not have show_in_embedded_editor or required toggled on
+         * we must not show it in the embedded editor
+         */
+        if (!(fieldSchema.show_in_embedded_editor || fieldSchema.required)) {
```
tomaskikutis: We'll eventually use `getPlanningProfileFields` for non-embedded editors too. Add a argument which would tell if you want to get fields for embedded editor or not.

tomaskikutis (approved):

PR #2169, opened 2024-12-24
[SDESK-7463] - Create new async PlanningTypes resource and service

BrianMwangi21:
### Purpose
This PR creates new async planning types resource and service and exposes their rest API:

- Implementing a new resource model for planning_types
- Implementing new async resource service for planning_types
- Implementing resource config for planning_types and register it in planning module
- Disable the api in old planning_types resource

Solves: [SDESK-7463](https://sofab.atlassian.net/browse/SDESK-7463)

[SDESK-7463]: https://sofab.atlassian.net/browse/SDESK-7463

### DIFF
--- a/server/planning/content_profiles/planning_types_async_service.py
+++ b/server/planning/content_profiles/planning_types_async_service.py
@@ -2,8 +2,8 @@ from typing import Any
 from copy import deepcopy
 
 from superdesk.core.resources import AsyncResourceService
-from superdesk.core.types import SearchRequest, ProjectedFieldArg
-from superdesk.utils import ListCursor
+from superdesk.core.resources.cursor import ElasticsearchResourceCursorAsync
+from superdesk.core.types import SearchRequest, ProjectedFieldArg, SortParam
 
 from planning.common import planning_link_updates_to_coverage, get_config_event_related_item_search_provider_name
 from planning.types import PlanningTypesResourceModel
@@ -50,7 +50,7 @@ class PlanningTypesAsyncService(AsyncResourceService[PlanningTypesResourceModel]
             # lookup name from either **lookup of planning_item(if lookup has only '_id')
             lookup_name = lookup.get("name")
             if not lookup_name and planning_type:
-                lookup_name = planning_type.to_dict().get("name")
+                lookup_name = planning_type.name
 
             default_planning_type = deepcopy(
                 next(
@@ -59,16 +59,39 @@ class PlanningTypesAsyncService(AsyncResourceService[PlanningTypesResourceModel]
                 )
             )
             if not planning_type:
-                await self._remove_unsupported_fields(default_planning_type)
+                self._remove_unsupported_fields(default_planning_type)
                 return PlanningTypesResourceModel(**default_planning_type)
 
-            await self.merge_planning_type(planning_type.to_dict(), default_planning_type)
+            self.merge_planning_type(planning_type.to_dict(), default_planning_type)
             return planning_type
         except IndexError:
             return None
 
-    async def get(self, req, lookup) -> ListCursor:
-        cursor = await super().search(lookup)
+    async def find(
+        self,
+        req: SearchRequest | dict,
+        page: int = 1,
+        max_results: int = 25,
+        sort: SortParam | None = None,
+        projection: ProjectedFieldArg | None = None,
+        use_mongo: bool = False,
+    ) -> ElasticsearchResourceCursorAsync[PlanningTypesResourceModel]:
+        """
+        Overrides the base `find` to return a cursor containing planning types
+        with default configurations merged into the results from the database. If a planning
+        type is not present in the database, a default configuration is added.
+        """
+
+        if isinstance(req, SearchRequest):
+            req = {
+                "where": req.where,
+                "page": req.page,
+                "max_results": req.max_results,
+                "sort": req.sort,
+                "projection": req.projection,
+            }
+
+        cursor = await super().find(req, page, max_results, sort, projection, use_mongo)
         planning_types = await cursor.to_list_raw()
         merged_planning_types = []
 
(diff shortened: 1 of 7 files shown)


server/planning/content_profiles/planning_types_async_service.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2021 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
```
eos87: Let's get rid of this. The license is already in the repository and the license link here is broken anyways.

server/planning/content_profiles/planning_types_async_service.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    async def find_one(
+        self,
+        req: SearchRequest | None = None,
+        projection: ProjectedFieldArg | None = None,
+        use_mongo: bool = False,
+        version: int | None = None,
+        **lookup,
+    ) -> PlanningTypesResourceModel | None:
```
eos87, line 28: Can you please add some docstrings here as to why we are overriding the `find_one` method?

eos87 (commented):
Looks good, I just left a couple of comments.

server/planning/content_profiles/planning_types_async_service.py
```diff
@@ -0,0 +1,132 @@
...
+                    version=version,
+                )
+            )
+
+            planning_type = await super().find_one(search_request)
+
+            # lookup name from either **lookup of planning_item(if lookup has only '_id')
+            lookup_name = lookup.get("name")
+            if not lookup_name and planning_type:
+                lookup_name = planning_type.to_dict().get("name")
```
MarkLark86: Would it be possible to get the name attribute directly, as opposed to converting to a dict first?
i.e.
`lookup_name = planning_type.name`

server/planning/content_profiles/planning_types_async_service.py
```diff
@@ -0,0 +1,132 @@
...
+            if not planning_type:
+                await self._remove_unsupported_fields(default_planning_type)
+                return PlanningTypesResourceModel(**default_planning_type)
+
+            await self.merge_planning_type(planning_type.to_dict(), default_planning_type)
+            return planning_type
+        except IndexError:
+            return None
+
+    async def get(self, req, lookup) -> ListCursor:
```
MarkLark86: For the new async services, this is the [find](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/service.py#L593) method instead (which is also used by the [REST get endpoint](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/resource_rest_endpoints.py#L388))

server/planning/content_profiles/planning_types_async_service.py
```diff
@@ -0,0 +1,132 @@
...
+        version: int | None = None,
+        **lookup,
+    ) -> PlanningTypesResourceModel | None:
+        """
+        Overrides the `find_one` method to merge default planning type configurations
+        with database entries. If no entry exists in the database, it returns a default
+        planning type configuration.
+        """
+        try:
+            search_request = (
```
MarkLark86, line 35: Would it be possible to pass `req`, `projection`, `use_mongo`, `version` and `lookup` directly to the parent function, as opposed to re-constructing the req object here (as the parent function does this for us already)?
Or does this cause any issues, such as types?

server/planning/content_profiles/planning_types_async_service.py
```diff
@@ -0,0 +1,132 @@
...
+            if planning_type is None:
+                await self._remove_unsupported_fields(default_planning_type)
+                merged_planning_types.append(default_planning_type)
+            else:
+                await self.merge_planning_type(planning_type, default_planning_type)
+                merged_planning_types.append(planning_type)
+
+        return ListCursor(merged_planning_types)
+
+    async def merge_planning_type(self, planning_type: dict[str, Any], default_planning_type: dict[str, Any]):
```
MarkLark86: These next 2 methods do not wait on any I/O, therefor these don't need to be async

server/planning/types/planning_types.py
```diff
@@ -0,0 +1,48 @@
...
+from superdesk.core.resources.validators import validate_iunique_value_async, validate_data_relation_async
+
+
+class PlanningTypesResourceModel(ResourceModel):
+    # The name identifies the form in the UI to which the type relates
+    name: Annotated[fields.Keyword, validate_iunique_value_async("planning_types", "name")]
+
+    editor: dict[str, Any] = Field(
+        default_factory=dict,
+        description="Editor controls which fields are visible in the UI",
```
MarkLark86, line 17: :+1:  Nice use of description on the field, this will be handy in the future (when we auto-generate API docs)

server/planning/types/planning_types.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+    schema_config: dict[str, Any] = Field(
+        alias="schema",
```
MarkLark86, line 20: Is `schema` not allowed as a name (wondering why use an alias here)?

server/planning/types/planning_types.py
```diff
@@ -0,0 +1,48 @@
...
+        default_factory=list,
+        description="Fields visible in exports or downloads for events/planning",
+    )
+
+    # Audit information
+    created_by: Annotated[ObjectId, validate_data_relation_async("users")] | None = None
+    updated_by: Annotated[ObjectId, validate_data_relation_async("users")] | None = None
+    firstcreated: datetime = Field(default_factory=utcnow)
+    versioncreated: datetime = Field(default_factory=utcnow)
+    init_version: int
```
MarkLark86: I think this field is optional, so it could be None as well

server/planning/content_profiles/planning_types_async_service.py
BrianMwangi21: I had actually tried that initially but it was causing typing issues because of incompatibility with the base class signature.

server/planning/types/planning_types.py
BrianMwangi21: I was getting an error of overiding `schema` in the BaseModel and mismatch of types. So I decided to use another name but set an alias for it.

server/planning/content_profiles/planning_types_async_service.py
MarkLark86: That's what I thought, oh well

server/planning/types/planning_types.py
MarkLark86: sounds good

MarkLark86 (approved):

PR #2172, opened 2025-01-07
[SDESK-7470] - Create new async Agenda resource, service & REST API

BrianMwangi21:
### Purpose
This PR creates new async agenda resource and service and exposes their rest API:

- Implementing a new resource model for agendas
- Implementing new async resource service for agenda
- Implementing resource config for agenda and register it in planning module
- Disable the API in old agenda resource

Solves: [SDESK-7470](https://sofab.atlassian.net/browse/SDESK-7470)

[SDESK-7470]: https://sofab.atlassian.net/browse/SDESK-7470

### DIFF
--- a/server/planning/agendas_async/agendas_async_service.py
+++ b/server/planning/agendas_async/agendas_async_service.py
@@ -1,0 +1,60 @@
+from typing import Any
+from planning.core.service import BasePlanningAsyncService
+from planning.types import AgendasResourceModel
+from superdesk.resource_fields import ID_FIELD, ITEMS
+from planning.common import set_original_creator
+from apps.auth import get_user_id
+from superdesk import get_resource_service
+from superdesk.errors import SuperdeskApiError
+from superdesk.notification import push_notification
+
+
+class AgendasAsyncService(BasePlanningAsyncService[AgendasResourceModel]):
+    async def _generate_planning_info(self, docs):
+        # TODO-ASYNC: Change this when async planning service is done
+        planning_service = get_resource_service("planning")
+        for doc in docs:
+            doc["plannings"] = planning_service.get_planning_by_agenda_id(doc.get(ID_FIELD)).docs
+
+    async def on_fetched(self, docs):
+        await self._generate_planning_info(docs.get(ITEMS))
+
+    async def on_fetched_item(self, doc):
+        await self._generate_planning_info([doc])
+
+    async def on_create(self, docs: list[AgendasResourceModel]) -> None:
+        for doc in docs:
+            doc_dict = doc.to_dict()
+            set_original_creator(doc_dict)
+
+    async def on_created(self, docs: list[AgendasResourceModel]) -> None:
+        for doc in docs:
+            doc_dict = doc.to_dict()
+            push_notification(
+                "agenda:created",
+                item=str(doc_dict[ID_FIELD]),
+                user=str(doc_dict.get("original_creator", "")),
+            )
+
+    async def on_update(self, updates: dict[str, Any], original: AgendasResourceModel) -> None:
+        user_id = get_user_id()
+        if user_id:
+            updates["version_creator"] = get_user_id()
+
+    async def on_updated(self, updates: dict[str, Any], original: AgendasResourceModel) -> None:
+        await self._generate_planning_info([updates])
+        push_notification(
+            "agenda:updated",
+            item=str(original.to_dict()[ID_FIELD]),
+            user=str(updates.get("version_creator", "")),
+        )
+
+    async def on_delete(self, doc: AgendasResourceModel):
+        # TODO-ASYNC: Change this when async planning service is done
+        if get_resource_service("planning").get_planning_by_agenda_id(doc.get(ID_FIELD)).count() > 0:
+            raise SuperdeskApiError.badRequestError(
+                message="Agenda is referenced by Planning items. " "Cannot delete Agenda"
+            )
+
+    async def on_deleted(self, doc: AgendasResourceModel):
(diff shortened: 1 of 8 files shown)


server/planning/agendas_async/agendas_async_service.py
```diff
@@ -0,0 +1,60 @@
...
+        await self._generate_planning_info([doc])
+
+    async def on_create(self, docs: list[AgendasResourceModel]) -> None:
+        for doc in docs:
+            doc_dict = doc.to_dict()
+            set_original_creator(doc_dict)
+
+    async def on_created(self, docs: list[AgendasResourceModel]) -> None:
+        for doc in docs:
+            doc_dict = doc.to_dict()
```
MarkLark86: No need to convert it to a dict, as you can use the Pydantic model directly here.
```python
push_notification(
    "agenda:created",
    item=str(doc.id),
    user=str(doc.original_creator) if doc.original_creator else None
)
```

server/planning/agendas_async/agendas_async_service.py
```diff
@@ -0,0 +1,60 @@
...
+from superdesk.resource_fields import ID_FIELD, ITEMS
+from planning.common import set_original_creator
+from apps.auth import get_user_id
+from superdesk import get_resource_service
+from superdesk.errors import SuperdeskApiError
+from superdesk.notification import push_notification
+
+
+class AgendasAsyncService(BasePlanningAsyncService[AgendasResourceModel]):
+    async def _generate_planning_info(self, docs):
```
MarkLark86: These next 3 are for the REST API and not the on the data layer. Though we don't yet have `on_fetched` mechanism in the new REST API layer. 

Instead, could you please comment these out and I will add this functionality into the base REST API class, and add a TODO-ASYNC comment

server/planning/agendas_async/agendas_async_service.py
```diff
@@ -0,0 +1,60 @@
...
+    async def on_update(self, updates: dict[str, Any], original: AgendasResourceModel) -> None:
+        user_id = get_user_id()
+        if user_id:
+            updates["version_creator"] = get_user_id()
+
+    async def on_updated(self, updates: dict[str, Any], original: AgendasResourceModel) -> None:
+        await self._generate_planning_info([updates])
+        push_notification(
+            "agenda:updated",
+            item=str(original.to_dict()[ID_FIELD]),
```
MarkLark86: You can use the Pydantic model directly here, i.e. `item=str(original.id)`, makes it much easier to read too

MarkLark86 (commented):
Looks good, just a couple small things

server/planning/agendas_async/agendas_async_service.py
```diff
@@ -0,0 +1,57 @@
...
+    # async def on_fetched(self, docs):
+    #     await self._generate_planning_info(docs.get(ITEMS))
+    #
+    # async def on_fetched_item(self, doc):
+    #     await self._generate_planning_info([doc])
+
+    async def on_create(self, docs: list[AgendasResourceModel]) -> None:
+        for doc in docs:
+            doc_dict = doc.to_dict()
+            set_original_creator(doc_dict)
```
eos87: I'm not sure this is going to work. Once you convert the pydantic model instance into a dictionary, there are two different objects, the changes made to the dict won't be reflected in the model instance, therefore the data won't be saved into the database. Or maybe I'm missing something?

server/planning/agendas_async/agendas_async_service.py
```diff
@@ -0,0 +1,57 @@
...
+            item=str(original.id),
+            user=str(updates.get("version_creator", "")),
+        )
+
+    async def on_delete(self, doc: AgendasResourceModel):
+        # TODO-ASYNC: Change this when async planning service is done
+        if get_resource_service("planning").get_planning_by_agenda_id(doc.id).count() > 0:
+            raise SuperdeskApiError.badRequestError(
+                message="Agenda is referenced by Planning items. " "Cannot delete Agenda"
+            )
```
eos87, line 48: I think we need another ticket to take care of `get_planning_by_agenda_id` method, I was not yet migrated.

server/planning/types/agendas.py
```diff
@@ -0,0 +1,11 @@
...
+from typing import Annotated
+
+from .base import BasePlanningModel
+from superdesk.core.resources import fields
+from superdesk.core.resources.validators import validate_iunique_value_async
+
+
+class AgendasResourceModel(BasePlanningModel):
+    name: Annotated[fields.Keyword, validate_iunique_value_async("agendas", "name")]
+    is_enabled: bool = Field(default=True)
```
eos87: I think you can simplify this like:
```suggestion
    is_enabled: bool = True
```

eos87 (commented):
I left a couple of comments

server/planning/agendas_async/agendas_async_service.py
MarkLark86: This should be added in the base Planning service, as we do in [Newshub](https://github.com/superdesk/newsroom-core/blob/async/newsroom/core/resources/service.py#L22), for both create and update hooks

server/planning/agendas_async/agendas_async_service.py
MarkLark86: IMO, these sort of methods should be util methods and not a method on the service. Historically we added these onto the service, and they became bloated and hard to tell what functions are used where.

It would be ideal to keep resource services as small as possible

BrianMwangi21: Noted. I'll leave it as is but with a todo when the `get_planning_by_agenda_id` is added to a utils file

MarkLark86 (approved):

PR #2175, opened 2025-01-07
avoid overriding user changes via event ingest

petrjasek:
check event history what keys were updated
by users and ignore those when updating from
ingest.

SDCP-861

## Front-end checklist

### DIFF
--- a/server/planning/events/events.py
+++ b/server/planning/events/events.py
@@ -138,17 +138,15 @@ def is_event_updated(new_item: Event, old_item: Event) -> bool:
     return False
 
 
-def get_user_updated_keys(id: str) -> set[str]:
-    # Placeholder function definition
-    # Replace this with the actual implementation
+def get_user_updated_keys(event_id: str) -> set[str]:
     history_service = get_resource_service("events_history")
-    updates = history_service.get_by_id(id)
+    updates = history_service.get_by_id(event_id)
     updated_keys = set()
     for update in updates:
         if not update.get("user_id"):
             continue
-        for key in update.get("update", {}):
-            updated_keys.add(key)
+        if update.get("update"):
+            updated_keys.update(update["update"].keys())
     return updated_keys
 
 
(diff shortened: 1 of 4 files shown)


server/planning/events/events_history.py
```diff
@@ -85,3 +98,15 @@ def on_update_repetitions(self, updates, event_id, operation):
 
     def on_update_time(self, updates, original):
         self.on_item_updated(updates, original, "update_time")
+
+    def get_by_id(self, id: str) -> list[EventHistoryRecord]:
```
MarkLark86: As this is for v2.8.x, are we able to use `list` as a type in Python v3.8, or do we have to continue using `List` from the `typing` module?

server/planning/events/events.py
```diff
@@ -128,6 +138,20 @@ def is_event_updated(new_item: Event, old_item: Event) -> bool:
     return False
 
 
+def get_user_updated_keys(id: str) -> set[str]:
+    # Placeholder function definition
```
MarkLark86: I'm guessing this comment is no longer relevant?

server/planning/events/events_history.py
MarkLark86: `id` shadows built-in name `id`, it would be good to use non built in names for variables

server/planning/events/events.py
```diff
@@ -128,6 +138,20 @@ def is_event_updated(new_item: Event, old_item: Event) -> bool:
...
+    # Replace this with the actual implementation
+    history_service = get_resource_service("events_history")
+    updates = history_service.get_by_id(id)
+    updated_keys = set()
+    for update in updates:
+        if not update.get("user_id"):
+            continue
+        for key in update.get("update", {}):
```
MarkLark86: Instead of using an iterator, you could use `updated_keys.update(update.get("update", {}).keys())`

server/planning/events/events.py
```diff
@@ -128,6 +138,20 @@ def is_event_updated(new_item: Event, old_item: Event) -> bool:
     return False
 
 
+def get_user_updated_keys(id: str) -> set[str]:
```
MarkLark86: Same here with using `set` for type, would need to use `Set` from `typing` module to support Python v3.8

setup.cfg
```diff
@@ -21,7 +21,7 @@ input_file = po/server.pot
 output_dir = server/planning/translations
 
 [mypy]
-python_version = 3.8
+python_version = 3.10
```
MarkLark86, line 24: I know Python v3.8 is EOL'd, but this branch is for a bugfix/improvement release for versions that are already in Production. Not sure if we can assume an OS/environment update along with a minor release

Should we instead look at a v2.9 as a change in supported Python version?

server/planning/events/events_history.py
petrjasek: I wouldn't try to support 3.8 anymore, it seems like the builds got failing there due to dependencies, our production servers are using 3.10 now

setup.cfg
petrjasek: I've created release/2.9 for it now, so this won't be part of 2.8

MarkLark86 (approved):

PR #2176, opened 2025-01-07
[SDESK-7456] Migrate Planning resource service to async

eos87:
### Purpose 
Extend the new Planning async resource service by copying across functionality from the old service that is specific to the data layer

### What has changed
- Moved `on_create`, `on_created`, `on_update` & `on_updated` method with all the connected logic
- Refactored some of the methods with the intention to make them easier to maintain 
- Adjusted some of the tests to use the new async service
- Added as many docstrings as possible trying to describe the intention of the methods to make them easier to understand for future devs and ourselves
- Added a new `planning_utils` module so we can start moving some utils functions there
- `flag_expired_items_test.py` test data has been restored and fixed accordingly
- Rename planning `service.py` to `planning_service.py` following the project file naming convention 

### Pending
- The Planning service is big and kinda entangled. It would be good to have another round to refactor in into separate modules according to their concern
- Not all the references to old service have been removed. I figured we can continue updating them as we move forward.

Thanks for checking!


server/planning/types/common.py
```diff
@@ -198,6 +197,10 @@ class CoverageAssignedTo:
     state: fields.Keyword | None = None
     contact: fields.Keyword | None = None
 
+    # TODO-ASYNC: double check if we should add validation relation for desk
+    desk: fields.Keyword | None = None
```
MarkLark86: I would say yes, to help ensure data integrity

server/planning/types/common.py
```diff
@@ -127,8 +127,7 @@ class Place:
     rel: fields.Keyword | None = None
 
 
-@dataclass
-class RelatedEvent:
+class RelatedEvent(Dataclass):
```
MarkLark86, line 130: Why the change from `@dataclass` decorator to a `Dataclass` inheritance, as well as `ScheduledUpdate` and `PlanningCoverage`?

server/planning/types/planning.py
```diff
@@ -28,11 +28,12 @@
 @dataclass
 class Flags:
     marked_for_not_publication: bool = False
+    # TODO: double check if we can fix this typo `overide` -> `override`
     overide_auto_assign_to_workflow: bool = False
 
 
 class PlanningResourceModel(BasePlanningModel, LockFieldsMixin):
-    guid: fields.Keyword
+    guid: fields.Keyword | None = None
```
MarkLark86: Wondering if we can do something like [this](https://github.com/superdesk/newsroom-core/blob/async/newsroom/types/agenda.py#L265) and [this](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/model.py#L142) to make this field mandatory. As this model should always have both `_id` and `guid` defined

server/planning/types/planning.py
```diff
@@ -124,7 +125,7 @@ class PlanningResourceModel(BasePlanningModel, LockFieldsMixin):
         default_factory=list, alias="updates_schedule"
     )
 
-    planning_date: datetime
+    planning_date: datetime | None = None
```
MarkLark86: Is there a scenario where `planning_date` **is not** defined? I would have thought this was mandatory

server/planning/planning/planning_service.py
```diff
@@ -0,0 +1,1568 @@
...
+
+            await self.update(
+                item[ID_FIELD],
+                {
+                    "recurrence_id": updates["recurrence_id"],
+                    "related_events": related_events,
+                },
+            )
+
+    async def create(self, _docs: list[PlanningResourceModel | dict[str, Any]]) -> list[str]:
```
MarkLark86, line 173: Is there a reason to use the `create` method directly here? To me, it looks like you can directly implement this in a `on_create` method instead.

server/planning/types/planning.py
MarkLark86: This way we won't need the following in the service
https://github.com/superdesk/superdesk-planning/blob/6388e54328f76e92650ca68a4e22f8f3c8c4df21/server/planning/planning/planning_service.py#L430-L432

server/planning/core/service.py
```diff
@@ -8,5 +8,17 @@
 PlanningResourceModelType = TypeVar("PlanningResourceModelType", bound=BasePlanningModel)
 
 
-class BasePlanningAsyncService(AsyncResourceService[Generic[PlanningResourceModelType]]):
-    pass
+class BasePlanningAsyncService(AsyncResourceService[PlanningResourceModelType]):
+    """Base async service to handle repetitive logic that is used throughout the planning project."""
+
+    async def on_create(self, docs: List[PlanningResourceModelType]) -> None:
```
MarkLark86, line 14: Should we also have an `on_update` handler to update the `version_creator` like we do in [Newshub](https://github.com/superdesk/newsroom-core/blob/async/newsroom/core/resources/service.py#L34)

server/planning/planning/planning_service.py
```diff
@@ -0,0 +1,1568 @@
...
+            "planning_date" in updated_planning and updated_planning["planning_date"] is None
+        ):
+            raise SuperdeskApiError(message="Planning item should have a date")
+
+        # TODO-ASYNC: Move this sanitize logic to the pydantic model instead
+        # sanitize_input_data(updated_planning)
+
+        self._validate_events_links(updated_planning)
+
+        # Validate if agendas being added are enabled agendas
```
MarkLark86, line 472: I suppose we can also add custom async validation for Agendas field directly onto the model, including removing duplicate entries

server/planning/planning/planning_service.py
```diff
@@ -0,0 +1,1568 @@
...
+            docs.extend(generated_planning_items)
+
+    def validate_planning(self, updated_planning: dict[str, Any], original=None):
+        if (not original and not updated_planning.get("planning_date")) or (
+            "planning_date" in updated_planning and updated_planning["planning_date"] is None
+        ):
+            raise SuperdeskApiError(message="Planning item should have a date")
+
+        # TODO-ASYNC: Move this sanitize logic to the pydantic model instead
+        # sanitize_input_data(updated_planning)
```
MarkLark86, line 468: Is this on the model now?

server/planning/planning/planning_service.py
```diff
@@ -0,0 +1,1568 @@
...
+        # set the updates dictionary with the values from the pydantic model
+        # in case the object was edited in-place.
+        # TODO-ASYNC: unify everything to use only pydantic models
+        updates = updated_planning.to_dict()
+
+        if update_method and update_method != UpdateMethods.SINGLE:
+            await self._update_recurring_planning_items(updates, original, update_method)
+
+    async def on_updated(self, updates: dict[str, Any], original_obj: PlanningResourceModel):
+        # TODO-ASYNC: figure out what to do with the `from_ingest` param here
```
MarkLark86, line 200: FYI: `from_ingest` is used from the [patch_in_mongo](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/planning/planning.py#L106), which is executed from an ingest cycle.

We will need to look at this once we work on the async ingest

MarkLark86 (commented):
Apart from the various TODO-ASYNC comments and a couple small things, it looks good

server/planning/core/service.py
```diff
@@ -8,5 +8,17 @@
...
+        """
+        Sets `original_creator` and  `version_creator` by default to avoid repeating this everywhere
+        """
+        await super().on_create(docs)
+
+        user_id = get_user_id()
+        if user_id:
+            for doc in docs:
+                doc.original_creator = user_id
+                doc.version_creator = user_id
```
BrianMwangi21: I added an `on_create` to the Base class in this [PR](https://github.com/superdesk/superdesk-planning/pull/2172) which is merged but I got the user first then the ID. 

When you merge async into this we can keep yours since it's only the ID needed.

server/planning/types/planning.py
eos87: yeah, I think we can. I will try that on Monday.

server/planning/types/common.py
eos87: I think it was because I was using some of the methods from pydantic that we introduced in https://github.com/superdesk/superdesk-core/pull/2764 but then I ended up not using them. I will double check. Anyways, `Dataclass` has advantages over `@dataclass` decorator as per https://github.com/superdesk/superdesk-core/pull/2764

server/planning/types/planning.py
eos87: Hm I think this is a leftover of me trying to figure out things. I'll double check.

server/planning/planning/planning_service.py
eos87: Yes and no. If we use the `on_create` method and apply the transformations to the provided docs (those to be created) this piece of code below fails because the original docs have been appended with the `generated_planning_items`, therefore we have an index error.  
https://github.com/superdesk/superdesk-core/blob/d93734e559933f577aeea8c2a4ec47163282d66c/superdesk/core/resources/service.py#L372-L381

Maybe you remember I've asked about this a couple of weeks ago on Slack, I will ping you again there and maybe we can talk more about it if you like, perhaps there is a better implementation for this if we change the parent `create` method.

server/planning/planning/planning_service.py
eos87: Let's have a separate ticket to address that and the one below about `sanitize_input_data`. What do you think?

BrianMwangi21 (commented):
Looks good to me

MarkLark86 (approved):
Looks good to me

PR #2177, opened 2025-01-08
Location field

thecalcc:
STT-31

## Front-end checklist

### DIFF
--- a/client/components/GeoLookupInput/index.tsx
+++ b/client/components/GeoLookupInput/index.tsx
@@ -3,6 +3,10 @@ import {AddGeoLookupInput} from './AddGeoLookupInput';
 
 import {LineInput, Label} from '../UI/Form';
 import {ILocation} from '../../interfaces';
+import {showModal} from '@sourcefabric/common';
+import {CreateNewGeoLookup} from './CreateNewGeoLookup';
+import {Provider} from 'react-redux';
+import {planningApi} from '../../superdeskApi';
 
 interface IProps {
     field: string;
@@ -24,7 +28,6 @@ interface IProps {
     popupContainer?(): HTMLElement;
     onPopupOpen?(): void;
     onPopupClose?(): void;
-    showAddLocationForm?(props: any): Promise<ILocation | undefined>;
 }
 
 export class GeoLookupInput extends React.PureComponent<IProps> {
@@ -41,7 +44,6 @@ export class GeoLookupInput extends React.PureComponent<IProps> {
             onFocus,
             popupContainer,
             refNode,
-            showAddLocationForm,
             ...props
         } = this.props;
 
@@ -66,7 +68,24 @@ export class GeoLookupInput extends React.PureComponent<IProps> {
                     popupContainer={popupContainer}
                     onPopupOpen={props.onPopupOpen}
                     onPopupClose={props.onPopupClose}
-                    showAddLocationForm={showAddLocationForm}
+                    showAddLocationForm={(props) => {
+                        let newLocation: ILocation;
+
+                        return showModal(({closeModal}) => (
+                            <Provider store={planningApi.redux.store}>
+                                <CreateNewGeoLookup
+                                    {...props}
+                                    onError={() => {
+                                        closeModal();
+                                    }}
+                                    onSuccess={(location) => {
+                                        newLocation = location;
+                                        closeModal();
+                                    }}
+                                />
+                            </Provider>
+                        )).then(() => newLocation);
+                    }}
                 />
             </LineInput>
         );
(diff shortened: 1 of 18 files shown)


client/extension_bridge.ts
```diff
@@ -20,6 +20,8 @@ import {AttachmentsInputStandalone} from './components/AttachmentsInputStandalon
 import {IPropsAttachmentsEditorStandalone} from './components/AttachmentsInputStandalone.interface';
 import {IPropsEditorFieldCoverages} from './components/fields/editor/coverages.interface';
 import {EditorFieldCoverages} from './components/fields/editor/Coverages';
+import {EditorFieldLocation} from './components/fields/editor/Location';
+import {IEditorFieldLocationProps} from './components/fields/editor/Location';
```
tomaskikutis: `import type {IProduct, ICompany, ISection} from 'interfaces';`

FYI you can import types from outside of planning extension if you use `import type` keyword.

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+import {IEditorComponentProps} from 'superdesk-api';
+import {ILocationFieldConfig, ILocationFieldUserPreferences, ILocationValueOperational} from './interfaces';
+import {extensionBridge} from '../../extension_bridge';
+
+type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldConfig, ILocationFieldUserPreferences>;
+
+export class Editor extends React.PureComponent<IProps> {
```
tomaskikutis, line 8: Would it support more than one location field in the same content profile? We should avoid passing the entire item when possible.

client/components/GeoLookupInput/index.tsx
```diff
@@ -66,7 +68,24 @@ export class GeoLookupInput extends React.PureComponent<IProps> {
                     popupContainer={popupContainer}
                     onPopupOpen={props.onPopupOpen}
                     onPopupClose={props.onPopupClose}
-                    showAddLocationForm={showAddLocationForm}
+                    showAddLocationForm={(props) => {
+                        let newLocation: ILocation;
+
+                        return showModal(({closeModal}) => (
```
tomaskikutis: it's a mistake that showModal returns a promise. I looked at the source and it simply returns `Promise.resolve()`. We'll update it to return void. Use a separate promise instead:

```
showAddLocationForm={(props) => {
    return new Promise((resolve) => {
        showModal(({closeModal}) => (
            <Provider store={planningApi.redux.store}>
                <CreateNewGeoLookup
                    {...props}
                    onError={() => {
                        closeModal();
                    }}
                    onSuccess={(location) => {
                        resolve(location);
                        closeModal();
                    }}
                />
            </Provider>
        ))
    });
}}
```

client/components/planning-editor-standalone/profile-fields.ts
```diff
@@ -62,5 +62,9 @@ export const getPlanningProfileFields = (options: {embeddedOnly?: boolean}): Arr
         }
     }
 
-    return convertedFieldIds;
+    return convertedFieldIds.concat({
```
tomaskikutis: why is location getting special treatment here? For example slugline(or text type) isn't mentioned in this file.

thecalcc: I have to mock the content profile because we don't have the embedded event editor setup, this is the simplest way without investing a lot of time

client/planning-extension/src/authoring-react-fields/location/editor.tsx
thecalcc: I don't think so? What do you mean not passing the entire item? It's working on `item` using `field`, I don't think it's worth investing more time since it also supports existing editor

client/components/planning-editor-standalone/profile-fields.ts
tomaskikutis: Wouldn't it work to modify the planning content profile in the database to include location? so you don't need to commit something that looks like test code

client/planning-extension/src/authoring-react-fields/location/editor.tsx
tomaskikutis: > I don't think so?

It would be good to implement it. That's the point of a field **type** - that you can have multiple fields of the same type. See if it can easily be fixed by passing a fake item to the component from the old editor as [I did with coverages](https://github.com/superdesk/superdesk-planning/blob/feature/embedded-events-editor/client/planning-extension/src/authoring-react-fields/coverages/editor.tsx#L26)

thecalcc: a like this sure!

client/components/planning-editor-standalone/profile-fields.ts
thecalcc: Yep that can be also, downside is we're messing with the data - might get into a corrupted state after some backend processing

tomaskikutis: you can backup the database before manually editing it and restore when done with the work. Or simply reinitialize

tomaskikutis (approved):

PR #2179, opened 2025-01-13
Saving planning items

thecalcc:
STT-31

## Front-end checklist


client/components/Main/ItemEditor/ItemManager.ts
```diff
@@ -657,10 +659,12 @@ export class ItemManager {
         }
 
         const promise = !updateStates ?
-            Promise.resolve() :
-            this.setState({
-                submitting: true,
-                submitFailed: false,
+            handleUnsavedChanges(this.props.editorType) :
+            handleUnsavedChanges(this.props.editorType).then((x) => {
```
thecalcc: Had to do this because of types, otherwise TS doesn't infer the right types

thecalcc:
@tomaskikutis let me know if you think it's better to create a wrapper around itemManager and do `handleUnsavedChanges` there instead of tying that to the itemManager's implementation of post,save, etc.

thecalcc:
- [x] expose validation errors from exposed
- [x] save event -> save gets called on all embedded items also
    - [x] validation errors -> first field of 1st embedded item with issues has to be scrolled to (and focused)
    - [x] open toggle box if closed
    - [x] user fixes this error -> repeats the flow
- [x] closing event -> handleUnsavedChanges for embedded items
    - [x] user can discard everything (from main and embedded items both)

client/components/editor-standalone/field-definitions/date-time-config.ts
```diff
@@ -1,19 +1,42 @@
...
+        storageAdapterPlanning: {
+            retrieveStoredValue: (item, fieldId) => {
+                if (typeof item[fieldId] === 'object') {
+                    return (item[fieldId] as moment.Moment).toString();
+                }
 
-    return field;
+                return item[fieldId];
+            },
+            storeValue: (item, operationalValue) => {
```
tomaskikutis: see if this can be handled on ui-framework side

client/components/Main/ItemEditor/EditorHeader.tsx
```diff
@@ -22,14 +22,15 @@ import {StretchBar} from '../../UI/SubNav';
...
+import {IEditorProps, IEditorState, IEventOrPlanningItem, ILockedItems, IPrivileges, ISession} from 'interfaces';
 import {IUser} from 'superdesk-api';
 import {ItemManager} from './ItemManager';
 import {AutoSave} from './AutoSave';
+import {IUIButtonProps} from 'components/UI/Button';
 
 interface IProps {
-    diff: IEditorState['diff'];
-    initialValues: IEditorState['initialValues'];
+    diff: IEventOrPlanningItem;
```
tomaskikutis, line 32: Lots of type casting was done here. Why is that? Can it be avoided?

client/components/editor-standalone/save-handling.ts
```diff
@@ -0,0 +1,84 @@
...
+            return Promise.resolve();
+        }),
+        Promise.resolve(),
+    );
+};
+
+export const embeddedPlanningHasUnsavedChanges = () => {
+    let hasUnsavedChanges = false;
+
+    getEmbeddedPlanningExposed(EDITOR_TYPE.INLINE).forEach((x) => {
```
tomaskikutis: the entire thing can be written in a simpler manner:

```
return getEmbeddedPlanningExposed(EDITOR_TYPE.INLINE).some((x) => x.hasUnsavedChanges());
```

Using `some` also stops iteration when `true` is returned.

client/components/fields/editor/index.tsx
```diff
@@ -125,7 +126,7 @@ export const FIELD_TO_EDITOR_COMPONENT = {
     contacts: EditorFieldContacts,
     event_contact_info: EditorFieldContacts,
     occur_status: EditorFieldEventOccurenceStatus,
-    related_plannings: EditorFieldEventRelatedPlannings,
+    related_plannings: EventRelatedPlanningWrapper,
```
tomaskikutis: I would keep the naming. Isn't this wrapping an internal implementation detail?

client/components/editor-standalone/save-handling.ts
```diff
@@ -0,0 +1,84 @@
...
+        } catch {
+            return null;
+        }
+    })
+        .filter(isNullOrUndefined);
+
+    return exposedAuthoringArray;
+};
+
+const FIRST_ERROR = 0;
```
tomaskikutis: Move it inside the function. It's specific to a single usage.

client/components/editor-standalone/save-handling.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    return Promise.reject();
+};
+
+/**
+ * Function that handles editor changes using editor refs.
+ * Execution is cancelled on the first encounter of an editor error.
+ * If an error occurs the first encountered embedded planning editor and the first error field is focused.
+ */
+export const handleEmbeddedPlannings = (editorType: EDITOR_TYPE, action: IEmbeddedPlanningsActionType = 'SAVE') => {
```
tomaskikutis: Drop the default value for action. It is important argument and should be passed explicitly.

If one of the planing items reports an issue saving - you should probably stop attempting it for the other ones, no?

That reducing promisses approach is quite ugly. A better alternative would be
```
let p = new Promise();

for (const item of list) {
    p = p.then();
}

return p;
```

Or even better, let's start using async/await.

client/components/editor-standalone/save-handling.ts
```diff
@@ -0,0 +1,84 @@
...
+    const embeddedEditorRef = planningApi.editor(editorType).dom.fields['related_plannings']?.current;
+    const relatedPlanningsRefs: IRelatedPlanningRefs = embeddedEditorRef?.relatedPlanningRefs;
+
+    return relatedPlanningsRefs;
+};
+
+const getEmbeddedPlanningExposed = (editorType: EDITOR_TYPE): Array<IExposedFromAuthoring<void>> => {
+    const relatedPlanningsRefs = getEmbeddedAuthoringRefs(editorType);
+    const exposedAuthoringArray = Object.values(relatedPlanningsRefs ?? []).map((x) => {
+        try {
```
tomaskikutis: Can you look into the reason it crashes? I did suggest to throw an error, but I was expecting it would be simply a way to tell the compiler not to worry about things being `null`, because we would only call it when it is not `null`.

client/components/Main/ItemEditor/EditorHeader.tsx
thecalcc: Of course it can, I just had to draw the line somewhere with refactoring, so I thought this is an explicit approach, next time I come I can do more

tomaskikutis: The problem is that it is no longer clear what the types are supposed to be and why are you overwriting them. Even a comment would help there. 

Planning isn't being type-checked on build/lint at the moment. When working on some old code I sometimes leave type issues be if the types don't make sense and can't be easily improved.

thecalcc: You can easily see how this component is made to handle both item types in planning. In case 1 where you have an event item, execution is different than case 2 where you have a planning item.

client/components/editor-standalone/save-handling.ts
thecalcc: I did try to reproduce multiple times, but it didn't crash for me this time. I've removed the catch. We'll see if it crashes for someone in the future.

client/components/Main/ItemEditor/ItemManager.ts
```diff
@@ -830,7 +830,7 @@ export class ItemManager {
             ));
     }
 
-    unlockAndCancel(embeddedEditorAction?: IEmbeddedPlanningsActionType) {
+    unlockAndCancel(embeddedEditorAction: IEmbeddedPlanningsActionType = 'SAVE') {
```
tomaskikutis: do not use a default here

client/components/editor-standalone/field-definitions/date-time-config.ts
thecalcc: I tried addressing it in the UI framework PR I pushed yesterday, but it looks like i'll still need it

tomaskikutis (commented):
all good, I'll approve when you handle the ui-framework comment

tomaskikutis:
also - I fixed unit test and lint in `feature/embedded-events-editor` branch - so update from that branch and make sure it passes on your branch too. Not e2e - those I didn't fix yet.

client/components/editor-standalone/field-definitions/date-time-config.ts
tomaskikutis: by "this" I meant conversion of moment date. I expect moment usage to be dropped from here. If you still need the adapter, that's fine

client/components/editor-standalone/save-handling.ts
```diff
@@ -75,5 +77,5 @@ export const handleEmbeddedPlannings = async(
 };
 
 export const embeddedPlanningHasUnsavedChanges = () => {
-    return getEmbeddedPlanningExposed(EDITOR_TYPE.INLINE).some((x) => x.hasUnsavedChanges());
+    return getEmbeddedPlanningExposed(EDITOR_TYPE.INLINE).some((x) => x?.hasUnsavedChanges?.());
```
tomaskikutis: would it work to return an empty array vs `[undefined, undefined]`?

Also, `.some` should return a boolean, your version sometimes returns `undefined`.

client/components/Main/ItemEditor/tests/ItemManager_test.ts
```diff
@@ -1525,13 +1525,13 @@ describe('components.Main.ItemManager', () => {
             restoreSinonStub(planningApi.locks.unlockItem);
         });
 
-        it('doesnt call unlock if the item isnt locked', (done) => {
+        it.only('doesnt call unlock if the item isnt locked', (done) => {
```
tomaskikutis: remove focusing of a single test via `only`

client/components/Events/EventEditor/index.tsx
```diff
@@ -109,6 +105,10 @@ class EventEditorComponent extends React.PureComponent<IProps> {
     }
 
     renderHeader() {
+        if (appConfig.planning_event_link_method === 'many_secondary') {
```
tomaskikutis: what's up with primary/secondary link types here? Why do they matter?

thecalcc: my bad

client/components/editor-standalone/save-handling.ts
```diff
@@ -77,5 +77,11 @@ export const handleEmbeddedPlannings = async(
 };
 
 export const embeddedPlanningHasUnsavedChanges = () => {
-    return getEmbeddedPlanningExposed(EDITOR_TYPE.INLINE).some((x) => x?.hasUnsavedChanges?.());
+    const planningsExposed = getEmbeddedPlanningExposed(EDITOR_TYPE.INLINE);
+
+    if (planningsExposed.length > 0) {
```
tomaskikutis: `return (planningsExposed ?? []).some((x) => x.hasUnsavedChanges());` is cleaner and avoid this extra condition

client/validators/profile.ts
```diff
@@ -40,8 +37,11 @@ export const formProfile = ({field, value, profile, errors, messages, diff}) =>
             errors[field] = gettext('Too long');
             messages.push(gettext('{{ name }} is too long', {name: fieldLabel}));
         }
-    } else if (schema.required && !schema.multilingual && (
-        typeof fieldValue === 'number' ? !fieldValue : isEmpty(fieldValue))) {
+    } else if (schema.required
+        && !schema.multilingual
+        // lodash's `isEmpty` does not handle date values correctly
```
tomaskikutis: `isEmpty` doesn't support dates as arguments thus we shouldn't comment here about it not working as if it were a lodash issue. The way to check if date is empty is `date == null`

client/validators/profile.ts
```diff
@@ -40,8 +37,11 @@ export const formProfile = ({field, value, profile, errors, messages, diff}) =>
...
+        && ((typeof fieldValue === 'number' || isDate(fieldValue)) ? !fieldValue : isEmpty(fieldValue))
```
tomaskikutis: move testing if it has a value into a variable so it's more readable.

client/validators/profile.ts
```diff
@@ -29,6 +29,9 @@ export const formProfile = ({field, value, profile, errors, messages, diff}) =>
 
     fieldLabel = gettext(fieldLabel).toUpperCase();
 
+    // lodash's `isEmpty` does not handle date values, even though they are technically an object instance
+    const isValueEmpty = (typeof fieldValue === 'number' || isDate(fieldValue)) ? !fieldValue : isEmpty(fieldValue);
```
tomaskikutis: Use `false` instead of `!fieldValue` so it's more clear.

client/validators/profile.ts
```diff
@@ -37,11 +40,7 @@ export const formProfile = ({field, value, profile, errors, messages, diff}) =>
             errors[field] = gettext('Too long');
             messages.push(gettext('{{ name }} is too long', {name: fieldLabel}));
         }
-    } else if (schema.required
-        && !schema.multilingual
-        // lodash's `isEmpty` does not handle date values correctly
-        && ((typeof fieldValue === 'number' || isDate(fieldValue)) ? !fieldValue : isEmpty(fieldValue))
-    ) {
+    } else if (schema.required && !schema.multilingual && isValueEmpty) {
```
tomaskikutis: only function names should start with "is". Use `valueEmpty` here.

client/validators/profile.ts
```diff
@@ -40,7 +48,7 @@ export const formProfile = ({field, value, profile, errors, messages, diff}) =>
             errors[field] = gettext('Too long');
             messages.push(gettext('{{ name }} is too long', {name: fieldLabel}));
         }
-    } else if (schema.required && !schema.multilingual && isValueEmpty) {
+    } else if (schema.required && !schema.multilingual && valueIsValid) {
```
tomaskikutis: actually `valueIsValid` isn't right, because if it is valid, why do you display an error that field is required? It should either be `!valueIsValid` or `!hasValidValue`

tomaskikutis (approved):
:tada:

PR #2180, opened 2025-01-14
fix elastic error when fetching expired events

petrjasek:
fix `Result window is too large, from + size must be less than or equal to: [10000]`

## Front-end checklist

### DIFF
--- a/server/planning/events/events.py
+++ b/server/planning/events/events.py
@@ -790,7 +790,7 @@ class EventsService(superdesk.Service):
         total_received = 0
         total_events = -1
 
-        while True:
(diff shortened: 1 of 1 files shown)


server/planning/events/events.py
```diff
@@ -790,7 +790,7 @@ def get_expired_items(self, expiry_datetime, spiked_events_only=False):
         total_received = 0
         total_events = -1
 
-        while True:
+        while total_received + get_max_recurrent_events() < 10000:  # 10k is max elastic limit
```
MarkLark86, line 793: We should probably set this limit somewhere in the new async code as well, to enforce it there

MarkLark86 (approved):

PR #2181, opened 2025-01-14
[SDESK-7473] - Upgrade the planning_validator to set of async functions

BrianMwangi21:
### Purpose
This PR creates new async planning_validator functions that can be used by other services.

- Convert the `PlanningValidateService` to new set of functions
- Delete the old resource service and it's use cases
- Update event & planning post resources to use the new `validate_docs` async function
- Update test to use new functions

Solves: [SDESK-7473](https://sofab.atlassian.net/browse/SDESK-7473)

[SDESK-7473]: https://sofab.atlassian.net/browse/SDESK-7473


server/planning/events/events_post.py
```diff
@@ -90,20 +92,20 @@ def validate_post_state(new_post_state):
...
+    # TODO-ASYNC: Uncomment method when service is changed to async to allow new async validate_docs function
+    # @staticmethod
+    # def validate_item(doc):
+    #     errors_list = await validate_docs([{"validate_on_post": True, "type": "event", "validate": doc}])
+    #     errors = errors_list[0]
+    #
+    #     if errors:
+    #         # We use abort here instead of raising SuperdeskApiError.badRequestError
+    #         # as eve handles error responses differently between POST and PATCH methods
+    #         abort(400, description=errors)
```
BrianMwangi21, line 104: @MarkLark86 for cases like this where we now have to use the new functions and the old service has been deleted, I have commented the functions out and where they are used until the service is converted to async. 

Thoughts ?

MarkLark86: No worries, the to-do comments should help us search for outstanding work

server/planning/validate/planning_validate.py
```diff
@@ -116,72 +89,80 @@ def _validate_vocabularies(self, vocabularies, field, value):
...
+    return {
+        field: get_validator_schema(field_schema)
+        for field, field_schema in validator["schema"].items()
+        if field in enabled_fields and field_schema and field_schema.get("validate_on_post", False) == validate_on_post
+    }
+
+
+async def get_validator(doc: dict[str, Any]):
+    """Get validators from planning types service."""
+    return get_resource_service("planning_types").find_one(req=None, name=doc[ITEM_TYPE])
```
MarkLark86: Rather than using the old service, you can now use the new one from the [PlanningTypesResourceModel](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/types/planning_types.py#L11), you can get it's service and use it's types throughout

server/planning/validate/planning_validate.py
```diff
@@ -116,72 +89,80 @@ def _validate_vocabularies(self, vocabularies, field, value):
...
-            doc["errors"] = self._validate(test_doc)
+    enabled_fields = get_enabled_fields(validator)
+    return {
+        field: get_validator_schema(field_schema)
+        for field, field_schema in validator["schema"].items()
+        if field in enabled_fields and field_schema and field_schema.get("validate_on_post", False) == validate_on_post
+    }
+
+
+async def get_validator(doc: dict[str, Any]):
```
MarkLark86: Don't forget to add return types. In this case this should be [planning.types.Event](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/types/__init__.py#L85) (alias for dict[str, Any]).

By converting the model instance to a dict means there is less code to be changed.

server/planning/validate/planning_validate.py
```diff
@@ -116,72 +89,80 @@ def _validate_vocabularies(self, vocabularies, field, value):
...
+    """Get validators from planning types service."""
+    return get_resource_service("planning_types").find_one(req=None, name=doc[ITEM_TYPE])
 
-        return [doc["errors"] for doc in docs]
 
-    def _get_validator(self, doc):
-        """Get validators."""
-        return get_resource_service("planning_types").find_one(req=None, name=doc[ITEM_TYPE])
+async def validate_doc(doc: dict[str, Any]):
```
MarkLark86: Instead of passing in a dictionary of arguments (like we did with the old system), what if we expand the required args out into this function.

This might be:
```python
async def validate_doc(
    item: dict,
    item_type: str,
    validate_on_post: bool = False
) -> list: ...
```
Then we can pass in only what's required. It might help make other functions cleaner as well, as you can then pass in item_type to functions instead of the entire dict.

MarkLark86 (commented):
There are also tests failing, due to `validate_item` function not available in tests

MarkLark86 (approved):

PR #2182, opened 2025-01-14
autopost ingested events on creation via rule handler

petrjasek:
SDCP-903

## Front-end checklist

### DIFF
--- a/client/utils/history.tsx
+++ b/client/utils/history.tsx
@@ -61,7 +61,7 @@ const getHistoryRowElement = (text, historyItem, users) => {
     if (text) {
         return (
             <div>
-                <span><strong>{text}</strong>{gettext(' by ')}</span>
+                <span><strong>{text}</strong>{historyItem.user_id != null ? gettext(' by ') : null}</span>
                 <span className="user-name">{self.getDisplayUser(historyItem.user_id, users)}</span>
                 <em> <AbsoluteDate date={historyItem._created} /> </em>
             </div>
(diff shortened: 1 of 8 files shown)


server/planning/io/ingest_rule_handler.py
```diff
@@ -164,13 +164,6 @@ def add_planning_agendas(self, ingest_item: Dict[str, Any], attributes: Dict[str
 
     def process_autopost(self, ingest_item: Dict[str, Any]):
         """Automatically post this item"""
-
```
MarkLark86, line 167: This might lead to double publishes, once here and once in the update event or planning. That is why this was there in the first place.

MarkLark86: For that other [PR](https://github.com/superdesk/superdesk-planning/pull/2009) which added the `pubstatus` on post from ingest, why is this an issue. If an item is formatted for creation, then we don't want any pubstatus attribute, because we want the system to control this initial posting of an item.

This is where the auto-ingest rules come into play, they allow setting up a rule to auto publish.

We have business rules in place that states:
* when an item is updated, if the original is published then auto-publish.
* If the `pubstatsus` field has _any_ value, then the system treats that item as published (without needing to check the planning_published collection).

MarkLark86: Maybe in that ingest rule handler, we can check that published planning collection instead, to truly determine if the item has been published or not

petrjasek: yep I've realized later it might be problematic, I think it would be more reliable to use a different field than pubstatus internally because we need a way to ingest pubstatus too.

petrjasek: but looking at the code it would be probably easier to introduce some `ingest_pubstatus` and handle it from there, there is too much code using that `pubstatus` now

MarkLark86: Yeah let's do that. Quick fix that's solid, reducing risk of bugs

server/planning/io/ingest_rule_handler.py
```diff
@@ -74,15 +74,18 @@ def apply_rule(self, rule: Dict[str, Any], ingest_item: Dict[str, Any], routing_
 
         if updates is not None:
             ingest_item.update(updates)
-
-        if attributes.get("autopost", False):
+        elif attributes.get("autopost", False):
```
MarkLark86, line 77: This might still need to be an `if` statement, because an item can be created, ingest rule add a calendar AND publish it, otherwise the item won't get published on ingest of first version.

server/planning/io/ingest_rule_handler.py
```diff
@@ -174,7 +176,7 @@ def process_autopost(self, ingest_item: Dict[str, Any]):
         item_id = ingest_item.get(config.ID_FIELD)
         update_post_item(
             {
-                "pubstatus": ingest_item.get("pubstatus") or POST_STATE.USABLE,
+                "pubstatus": ingest_item.get("ingest_pubstatus") or POST_STATE.USABLE,
```
MarkLark86, line 179: I would probably add a second fallback here, so it would be `ingest_pubstatus` or `pubstatus` or `POST_STATE.USABLE`

server/planning/planning/planning.py
```diff
@@ -105,7 +106,7 @@ def post_in_mongo(self, docs, **kwargs):
 
     def patch_in_mongo(self, id, document, original):
         """Patch an ingested item onto an existing item locally"""
-
+        prepare_ingested_item_for_storage(document)
```
MarkLark86, line 109: I think the Planning schema is missing the `ingest_pubstatus` field

client/utils/history.tsx
```diff
@@ -61,7 +61,7 @@ const getHistoryRowElement = (text, historyItem, users) => {
     if (text) {
         return (
             <div>
-                <span><strong>{text}</strong>{gettext(' by ')}</span>
+                <span><strong>{text}</strong>{historyItem.user_id != null ? gettext(' by ') : null}</span>
```
MarkLark86, line 64: What was the issue with the history component?

server/planning/io/ingest_rule_handler.py
petrjasek: when it adds calendar it calls `.patch` on event which does the publishing

server/planning/io/ingest_rule_handler.py
petrjasek: `pubstatus` is not set at this point because of `ingest_item['ingest_pubstatus'] = ingest_item.pop('pubstatus')` earlier

client/utils/history.tsx
petrjasek: it would be like `Ingested by Monday ...` when something was done without user action

MarkLark86 (approved):

PR #2185, opened 2025-01-17
fix failing e2e

petrjasek:
## Front-end checklist

### DIFF
--- a/client/components/Contacts/ActionBar.tsx
+++ b/client/components/Contacts/ActionBar.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {gettext} from 'core/utils';
+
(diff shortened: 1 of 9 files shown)


client/components/Contacts/ActionBar.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {gettext} from 'core/utils';
+
+export const ActionBar: React.FunctionComponent<any> = ({svc, readOnly, dirty, valid, onSave, onCancel}) => (
+    <div className="action-bar show">
```
MarkLark86, line 6: @fritzSF @tomaskikutis Are there components in the UI Framework that can be used here instead of this?

tomaskikutis: There is subnav that might be similar, but if existing one works without issues I won't be spending time to replace it.

tomaskikutis (approved):

PR #2186, opened 2025-01-20
[SDESK-7485] - Planning featured async resource, service & REST API

BrianMwangi21:
### Purpose
This PR creates new async PlanningFeatured resource and service and exposes the rest API:

- Implementing a new resource model for PlanningFeatured
- Implementing new async resource service for PlanningFeatured
- Implementing resource config for PlanningFeatured and register it in planning module
- Disable the API in old PlanningFeatured resource

Solves: [SDESK-7485](https://sofab.atlassian.net/browse/SDESK-7485)

[SDESK-7485]: https://sofab.atlassian.net/browse/SDESK-7485


server/planning/types/planning_featured.py
```diff
@@ -0,0 +1,22 @@
...
+class PlanningFeaturedResourceModel(BasePlanningModel):
+    date: datetime = Field(default_factory=utcnow)
+    items: list = Field(default_factory=list)
+    tz: str
+    posted: bool
+    last_posted_time: datetime = Field(default_factory=utcnow)
+    last_posted_by: Annotated[fields.ObjectId, validate_data_relation_async("users")]
+    firstcreated: datetime = Field(default_factory=utcnow)
+    versioncreated: datetime = Field(default_factory=utcnow)
+    item_type: str = Field(
```
MarkLark86, line 20: This won't be needed, as it's provided in base ResourceModel class, and uses the name using for the resource config instance.

server/planning/types/planning_featured.py
```diff
@@ -0,0 +1,22 @@
...
+from .base import BasePlanningModel
+from superdesk.utc import utcnow
+from superdesk.core.resources import fields
+from superdesk.core.resources.validators import validate_data_relation_async
+
+
+class PlanningFeaturedResourceModel(BasePlanningModel):
+    date: datetime = Field(default_factory=utcnow)
+    items: list = Field(default_factory=list)
+    tz: str
```
MarkLark86: I think `tz` should be optional, along with `last_posted_time` and `last_posted_by` (as we don't want these defaulting to `now`, but specifically assigned a value when a publish action happens).

server/planning/types/planning_featured.py
```diff
@@ -0,0 +1,22 @@
...
+
+from .base import BasePlanningModel
+from superdesk.utc import utcnow
+from superdesk.core.resources import fields
+from superdesk.core.resources.validators import validate_data_relation_async
+
+
+class PlanningFeaturedResourceModel(BasePlanningModel):
+    date: datetime = Field(default_factory=utcnow)
+    items: list = Field(default_factory=list)
```
MarkLark86: What's it a list of?

server/planning/planning/planning_featured_async_service.py
```diff
@@ -0,0 +1,113 @@
...
+            search_request = SearchRequest(where={"_id": _id})
+            items = await super().find(search_request)
+            if await items.count() > 0:
+                raise SuperdeskApiError.badRequestError(message="Featured story already exists for this date.")
+
+            await self.validate_featured_attrribute(doc_dict.get("items", []))
+            doc_dict["_id"] = _id
+            await self.post_featured_planning(doc_dict)
+            # set the author
+            set_original_creator(doc_dict)
```
MarkLark86: This is not needed as it's done in the parent [on_create()](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/core/service.py#L23)

server/planning/planning/planning_featured_async_service.py
```diff
@@ -0,0 +1,113 @@
...
+from superdesk.utc import utc_to_local, utcnow
+
+ID_DATE_FORMAT = "%Y%m%d"
+
+
+class PlanningFeaturedAsyncService(BasePlanningAsyncService[PlanningFeaturedResourceModel]):
+    """Async Service class for the planning featured model."""
+
+    async def on_create(self, docs: list[PlanningFeaturedResourceModel]):
+        for doc in docs:
```
MarkLark86, line 29: Don't forget to call the parent `on_create` and await it's response

server/planning/planning/planning_featured_async_service.py
```diff
@@ -0,0 +1,113 @@
...
+
+            await self.validate_featured_attrribute(doc_dict.get("items", []))
+            doc_dict["_id"] = _id
+            await self.post_featured_planning(doc_dict)
+            # set the author
+            set_original_creator(doc_dict)
+
+            # set timestamps
+            update_dates_for(doc_dict)
+            doc = PlanningFeaturedResourceModel(**doc_dict)
```
MarkLark86: If you update this service to use a ResourceModel instance, you can then update the instance values in other functions, rather than doing this (re-assigning while looping is also not ideal)

server/planning/planning/planning_featured_async_service.py
```diff
@@ -0,0 +1,113 @@
...
+
+    async def on_created(self, docs: list[PlanningFeaturedResourceModel]):
+        for doc in docs:
+            await self.enqueue_published_item(doc.to_dict(), doc.to_dict())
+
+    async def on_update(self, updates: dict[str, Any], original: PlanningFeaturedResourceModel):
+        # Find all planning items in the list
```
MarkLark86, line 50: Don't forget to call the parent `on_update` and await it's response

server/planning/planning/planning_featured_async_service.py
```diff
@@ -0,0 +1,113 @@
...
+        added_featured = [item_id for item_id in updates.get("items") or [] if item_id not in original.items or []]
+        await self.validate_featured_attrribute(added_featured)
+        updates["version_creator"] = str(get_user_id())
```
MarkLark86: This is not needed as it's done in the parent [on_update()](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/core/service.py#L34)

MarkLark86 (commented):
Besides a couple small things, looks good

server/planning/types/planning_featured.py
BrianMwangi21: @MarkLark86 the previous schema had list defined as this with no specific schema hence why I left it as is: 

```"items": {"type": "list"},```

File: server/planning/types/planning_featured.py (hunk 1 above)
Comment (BrianMwangi21) [reply server/planning/types/planning_featured.py:20]: @MarkLark86 a question for clarity. 

The base ResourceModel you highlighted has `created` and `updated` but not `firstcreated` and `versioncreated` which were then updated in the `on_create` by the `update_dates_for` call. For now since the default is utcnow, I've removed the call `update_dates_for` to avoid redundancy cause it was also just setting the current time.

I'm wondering however if I should still retain them cause they might be used elsewhere similar to how the Events and Planning models have both fields. 

Or maybe add them to `BasePlanningModel` together with `original_creator` and `version_creator`. 

Thoughts ?

File: server/planning/types/planning_featured.py (hunk 3 above)
Comment (MarkLark86) [reply server/planning/types/planning_featured.py]: I believe this should be a list of strings (the string IDs of the Planning items in it's list). We also have it in [Newshub as a list[str]](https://github.com/superdesk/newsroom-core/blob/async/newsroom/types/featured.py#L11). It would improve type checking etc

File: server/planning/types/planning_featured.py (hunk 1 above)
Comment (MarkLark86) [reply server/planning/types/planning_featured.py:20]: The `firstcreated` and `versioncreated` have a different context to `created` and `updated`.
* `created` & `updated`: Stores the date/time when the item was created/updated in our DB
* `firstcreated` & `versioncreated` may default to utcnow, but can also be used when ingesting items into our system, to indicate when they were initially created (possibly from external system).

Not sure if these fields would be needed by many other models, and probably should just be used ad-hoc when required

File: server/planning/planning/planning_featured_async_service.py [hunk 9]
```diff
@@ -0,0 +1,108 @@
...
+
+    async def on_update(self, updates: dict[str, Any], original: PlanningFeaturedResourceModel):
+        await super().on_update(updates, original)
+        # Find all planning items in the list
+        added_featured = [item_id for item_id in updates.get("items") or [] if item_id not in original.items or []]
+        await self.validate_featured_attrribute(added_featured)
+        await self.post_featured_planning(PlanningFeaturedResourceModel(**updates), original)
+
+    async def on_updated(self, updates: dict[str, Any], original: PlanningFeaturedResourceModel):
+        await self.enqueue_published_item(PlanningFeaturedResourceModel(**updates), original)
```
MarkLark86, line 57: Missing calls to super class' events. Such as `on_created`, `on_updated` etc. It's important to call super methods in case a parent class performs some actions.

Such as [resource signals](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/service.py#L487)

MarkLark86 (approved):

PR #2190, opened 2025-01-24
Coverage fixes

thecalcc:
STT-129
## Front-end checklist


thecalcc:
- [x] Improve coverage adding from advanced mode `CoverageAddAdvancedModal.tsx` - bad condition for disabling "Save" button. If form isn't dirty or coverages have no "assigned_to", Save isn't active. But if you click the checkbox at the left bottom isDirty is set to true -> allowing users to save coverages with no `assigned_to`. Decide whether do add bad value error messages, or just remove `isDirty` check.

client/planning-extension/src/authoring-react-fields/debounced-change-hoc.tsx
```diff
@@ -0,0 +1,75 @@
+import {DebouncedFunc, debounce, cloneDeep, set} from 'lodash';
+import React = require('react');
```
tomaskikutis: use es6 import

client/planning-extension/src/authoring-react-fields/debounced-change-hoc.tsx
```diff
@@ -0,0 +1,75 @@
...
+    ) => JSX.Element;
+    value: Array<IPlanningCoverageItem>;
+    onChange: (newValue: Array<IPlanningCoverageItem>) => void;
+}
+
+interface IDebouncedChangeHOCState {
+    renderedValue: Array<IPlanningCoverageItem>;
+}
+
+export class DebouncedChangeHOC extends React.PureComponent<IDebouncedChangeHOCProps, IDebouncedChangeHOCState> {
```
tomaskikutis: Clean up the implementation so it is not coupled with how coverage field works.

client/components/UI/Form/InputArray/index.tsx
```diff
@@ -44,7 +45,7 @@ interface IProps {
     errors: {[key: string]: any};
     showErrors: boolean;
     testId?: string;
-    getRef?(field: string, value: any): React.RefObject<any>;
+    getRef?(field: string, value: any): React.RefObject<CoverageEditorComponent>;
```
tomaskikutis: input array is used in multiple places so ref won't always be `CoverageEditorComponent`?

client/components/Coverages/CoverageAddButton/index.tsx
```diff
@@ -47,10 +47,7 @@ export class CoverageAddButton extends React.Component<IProps> {
     onChange(field: string, coverages: Array<DeepPartial<IPlanningCoverageItem>>) {
         this.props.onChange(
             field,
-            [
-                ...(this.props.value ?? []),
-                ...coverages,
-            ],
+            coverages,
```
tomaskikutis, line 36: `CoverageAddButton` is used in multiple places, did you check to make sure this change is compatible with other usages?

client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -31,8 +31,7 @@ import {getCoverageFields} from '../../../api/editor/item_planning';
 
 import '../style.scss';
 
-interface IProps {
-    // Values
+interface IOwnProps {
```
tomaskikutis, line 29: Pass required arguments instead of making it optional. You can pass `noop` if needed, but that's a good opportunity to document why is it not being passed.

client/components/Coverages/CoverageEditor/index.tsx
```diff
@@ -349,22 +342,13 @@ export class CoverageEditorComponent extends React.PureComponent<IProps> {
                 diff={diff}
                 index={index}
                 onChange={onChange}
-                newsCoverageStatus={newsCoverageStatus}
```
tomaskikutis, line 338: were those not used?

Be careful removing "unused" things. With react components, always check wheher props aren't being spread to child components before concluding that a prop is not used.

client/components/Coverages/ScheduledUpdate/index.tsx
```diff
@@ -245,7 +245,6 @@ export class ScheduledUpdate extends React.PureComponent<IProps> {
                 newsCoverageStatus={newsCoverageStatus}
                 genres={genres}
                 readOnly={readOnly}
-                invalid={componentInvalid}
```
tomaskikutis, line 248: why are you dropping this?

thecalcc: It's not actually used

client/components/Coverages/CoverageEditor/index.tsx
thecalcc: Good point

thecalcc:
@tomaskikutis I've only fixed the test I broke - that is in `remove_assignment.cy.ts`

client/components/Coverages/CoverageAddAdvancedModal.tsx
```diff
@@ -206,7 +212,7 @@ class CoverageAddAdvancedModalComponent extends React.Component<IProps, IState>
                     desk: get(coverage, 'desk._id'),
                 };
 
-                if (coverage.coverage_id != null && !coverage.enabled) {
+                if (coverage.coverage_id != null && coverage.enabled === false) {
```
tomaskikutis: when dealing with code that is not typed well, it's better to do `coverage.enabled !== true` to ensure `undefined` is treated as false.

client/components/editor-standalone/authoring-storage-planning-http.ts
```diff
@@ -5,7 +5,8 @@ import {superdeskApi} from '../../superdeskApi';
 import {getProfile} from './profile';
 import {omitFields} from './utils';
 import {AutoSaveHttp} from './authoring-autosave';
-import {planningUtils} from '../../utils';
+import {getErrorMessage, planningUtils} from '../../utils';
+import {gettext} from 'core/utils';
```
tomaskikutis: do not import from superdesk-core directly. Take `gettext` from `superdeskApi`

client/components/Coverages/CoverageAddAdvancedModal.tsx
```diff
@@ -215,10 +221,12 @@ class CoverageAddAdvancedModalComponent extends React.Component<IProps, IState>
                 return newCoverage;
             });
 
-        // create coverages
-        this.props.onChange(this.props.field, coverages);
+        // Update coverages
+        // Important note: `spiked` workflow_status is only used on the frontend
```
tomaskikutis, line 225: if that's the case, how about we get rid of that status and remove array item entirely instead of setting 'spiked' status?

e2e/cypress/e2e/assignments/remove_assignment.cy.ts
```diff
@@ -18,7 +18,7 @@ describe('Planning.Assignment: remove assignment', () => {
         editor.waitTillOpen();
     });
 
-    it('can remove Assignment', () => {
+    it.only('can remove Assignment', () => {
```
tomaskikutis: drop test focusing

client/components/UI/Form/InputArray/index.tsx
```diff
@@ -14,6 +14,7 @@ import {
     IPlanningNewsCoverageStatus,
 } from 'interfaces';
 import {IDesk} from 'superdesk-api';
+import {CoverageEditorComponent} from 'components/Coverages/CoverageEditor';
```
tomaskikutis: unused import

client/components/Coverages/CoverageAddAdvancedModal.tsx
thecalcc: I know what you mean, but right now it's used in another place as well and I don't want to spend time there as well

client/components/editor-standalone/authoring-storage-planning-http.ts
```diff
@@ -33,24 +33,26 @@ export const authoringStoragePlanningItemHttp: IAuthoringStorage<IPlanningItem>
...
         const {generatePatch} = superdeskApi.utilities;
 
-        return httpRequestJsonLocal<IPlanningItem>({
-            method: 'PATCH',
-            path: `/planning/${original._id}`,
-            payload: omitFields(
-                generatePatch(
-                    planningUtils.modifyForServer(original),
-                    planningUtils.modifyForServer(current),
+        return handleRemovedAssignments(current, original).then((updatedOriginal) =>
```
tomaskikutis: You should only remove assignments after successful save, not before.

client/components/Assignments/EditCoverageAssignmentModal.tsx
```diff
@@ -61,7 +61,7 @@ export class EditCoverageAssignmentModalComponent extends React.Component<IProps
 
     onSubmit() {
         this.setState({submitting: true}, () => {
-            this.props.modalProps.onChange(this.props.modalProps.field, this.state.diff);
+            this.props.modalProps.onChange?.(this.props.modalProps.field, this.state.diff);
```
tomaskikutis: why is is not defined?

client/components/editor-standalone/utils.ts
```diff
@@ -18,4 +20,57 @@ export function omitFields<T extends IBaseRestApiResponse>(
     }
 
     return {...omit(item, baseApiFields)};
-}
\ No newline at end of file
+}
+
+export async function handleRemovedAssignments(current: Partial<IPlanningItem>, original: Partial<IPlanningItem>) {
```
tomaskikutis: > export async function handleRemovedAssignments

I see you have async tag added here. How about actually using async/await here?

client/actions/planning/tests/ui_test.ts
```diff
@@ -438,7 +438,7 @@ describe('actions.planning.ui', () => {
 
             store.test(done, planningUi.saveFromAuthoring(data.plannings[0]))
                 .then(() => { /* no-op */ }, () => {
-                    expect(services.notify.error.callCount).toBe(1);
+                    expect(services.notify.error.callCount).toBe(2);
```
tomaskikutis, line 441: why does this change?

tomaskikutis (approved):

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
```diff
@@ -5,101 +5,99 @@ import {
...
+import {cloneDeep, omit, set} from 'lodash';
 
 type IProps = IEditorComponentProps<ICoveragesValueOperational, IUrlsFieldConfig, IUrlsFieldUserPreferences>;
 
 export class Editor extends React.PureComponent<IProps> {
     render() {
         const Container = this.props.container;
-
         const {EditorFieldCoverages} = extensionBridge.editor.fields;
+        const itemIsNotSaved = this.props.item._id == null || (this.props.item._id ?? '').includes('temp-');
```
tomaskikutis: Avoid negative naming. Call it `eventSaved` and when you need to use it, you can do `!eventSaved`.

Regarding `.includes('temp-')` - use the helper for this.

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
```diff
@@ -5,101 +5,99 @@ import {
...
-
-                item={{
-                    // coverages are the main value
-                    coverages: this.props.value,
-
-                    // related_events are used if available to prefill coverage fields when adding a new coverage
-                    related_events: this.props.item.related_events,
-                } as IPlanningItem}
+            <DebouncedChangeHOC
+                key={(this.props.value ?? []).map((x) => x.coverage_id).join('')}
```
tomaskikutis: comment on why the key is needed - would be useful to include the example with the bug you were having

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
```diff
@@ -5,101 +5,99 @@ import {
...
-                notifyValidationErrors={(errors) => {
-                    for (const error of errors) {
-                        superdesk.ui.notify.error(error);
-                    }
+                    return itemCopy.coverages;
                 }}
+                onChange={this.props.onChange}
+                value={
+                    cloneDeep(this.props.value ?? []).map((x) => (
+                        omit(x, 'planning._scheduledTime')
```
tomaskikutis: See if you can avoid omitting it here. It should be handled in `modifyForServer` - and if it's not being called - it probably should be.

tomaskikutis:
check e2e in `assignments/remove_assignment.cy.ts` - this likely breaks due to changes in this PR

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
thecalcc: I'm inside the extension, I need to double check whether it can be imported without breaking things

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
thecalcc: Can do it yes, I was originally using this to compare the previous value compared to the current one in the HOC and this field was breaking things

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
thecalcc: I now clicked around the coverage editor again and this fix isn't working. I think a key like either `new Date().toString()` would be good or something else to force a re-render.

thecalcc:
Before merging: Update superdesk-core to latest develop

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
tomaskikutis: you can always expose it via that bridge

client/components/fields/editor/AddCoverageToWorkflow.tsx
```diff
@@ -3,18 +3,24 @@ import {IEditorFieldProps} from '../../../interfaces';
...
 
         return (
-            <EditorFieldToggle
-                {...this.props}
-                field="add_coverage_to_workflow"
-                label={gettext('Add to workflow')}
-                defaultValue={false}
-            />
+            <Tooltip
+                disabled={this.props.disabled === false}
```
tomaskikutis, line 14: Shouldn't it be `disabled={this.props.disabled === true}`?

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
```diff
@@ -5,101 +5,100 @@ import {
...
-                    related_events: this.props.item.related_events,
-                } as IPlanningItem}
+            <DebouncedChangeHOC
 
                 /**
-                 * It looks like this prop is designed to accept a validation message.
-                 * authoring-react field types don't accept validation messages.
-                 * They are rendered higher in the component tree.
-                 * We do handle it in PlanningEditorStandalone component (on save).
+                 * Update component when value from props changes. Otherwise the component won't
```
tomaskikutis: With the comment now it's even more clear that this should be handled inside the component.

When returning components from render method we have an assumption that components will respect props - and that's how it should be. If some components don't respect props - we need to fix it inside the components itself. We have situations when it's best to simply remount it, but let's do a remounting inside a component itself via a wrapper component. Similar how we connect redux by creating a wrapper component.

client/components/editor-standalone/field-definitions/planning-date.ts
```diff
@@ -16,7 +16,7 @@ export const getPlanningDate = (): IFieldDefinition => {
             const field: IAuthoringFieldV2 = {
                 id: id,
                 name: gettext('Planning Date'),
-                fieldType: 'datetime',
+                fieldType: 'datetime-v2',
```
tomaskikutis: what is `v2`?

client/components/fields/editor/AddCoverageToWorkflow.tsx
thecalcc: No, I only want to show the tooltip when the button is disabled

client/components/editor-standalone/field-definitions/planning-date.ts
thecalcc: it's the new field id. I discovered there is also datetime in a client-core extension with `datetime` id. I don't want the two to be in conflict.

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
```diff
@@ -5,101 +5,96 @@ import {
...
+                onChange={this.props.onChange}
+                value={this.props.value}
+            >
+                {(changedValue, onChange) => (
+                    <EditorFieldCoverages
+                        readOnly={!itemSaved}
+                        field="coverages"
+                        item={{
+                            // coverages are the main value
+                            coverages: changedValue.map(modifyCoverageForClient),
```
tomaskikutis: Shouldn't it be passed to the field already modified?

tomaskikutis:
see this comment https://github.com/superdesk/superdesk-planning/pull/2190#discussion_r1952479583

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
thecalcc: it's not, so I'm doing it, otherwise genre doesn't work

client/actions/planning/tests/ui_test.ts
thecalcc: we show 2 messages - one on failing to link, one on failing to save

client/planning-extension/src/authoring-react-fields/coverages/editor.tsx
tomaskikutis: Doesn't genre work if you pass it modified already or you didn't try it?

client/actions/planning/tests/ui_test.ts
tomaskikutis: ok, just making sure you didn't bump the number to make tests pass without understanding the change.

client/utils/planning.tsx
```diff
@@ -1789,7 +1789,7 @@ function addToWorkflowCommon<T extends IPlanningCoverageItem | ICoverageSchedule
     next.workflow_status = COVERAGES.WORKFLOW_STATE.ACTIVE;
 
     // Scheduled update does not have `add_coverage_to_workflow`
-    if ('original_coverage_id' in (item as IPlanningCoverageItem)) {
+    if (!('scheduled_update_id' in item)) {
```
tomaskikutis: use [nameof](https://github.com/superdesk/sourcefabric-common/commit/b2e2e166676c8f4a0e394c68f9727372a2be79c6) to ensure type safety of keys. It was also in superdeskApi, but I found there are sometimes issues with accessing superdeskApi in tests thus I'm moving generic helpers to @sourcefabric/common

thecalcc:
Explaining my last commit for E2E fix:

I removed the "Add Planning Item" bookmark for events that have not been saved. This follows existing functionality and rules - we don't allow adding plannings to unsaved events in the "Related Plannings" box as well. 

Previously it was still possible to add a planning item if the event wasn't saved, and the test was testing whether you can do that. This no longer makes sense given our functionality. In the future writing a more through test on bookmarks would be a good follow up.

tomaskikutis (approved):

PR #2194, opened 2025-02-03
Embedded event list inline add button, embedded event remove button

thecalcc:
STT-31

## Front-end checklist

### DIFF
--- a/client/api/events.ts
+++ b/client/api/events.ts
@@ -160,7 +160,7 @@ function create(updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
             const events = modifySaveResponseForClient(response);
 
             // Update redux store, so changes are reflected in associated_event field
-            planningApi.redux.store.dispatch<any>(eventsApi.receiveEvents([response]));
+            planningApi.redux.store.dispatch<any>(eventsApi.receiveEvents(events));
 
             return planningApi.planning.searchGetAll({
                 recurrence_id: events[0].recurrence_id,
--- a/client/api/events.ts
+++ b/client/api/events.ts
@@ -18,6 +18,7 @@ import {arrayToString, convertCommonParams, cvsToString, searchRaw, searchRawGet
 import {eventUtils} from '../utils';
 import {eventProfile, eventSearchProfile} from '../selectors/forms';
 import planningApis from '../actions/planning/api';
+import eventsApi from '../actions/events/api';
 import {searchPlanning} from './planning';
 
 function convertEventParams(params: ISearchParams): Partial<ISearchAPIParams> {
@@ -157,6 +158,9 @@ function create(updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
     })
         .then((response) => {
             const events = modifySaveResponseForClient(response);
+
+            // Update redux store, so changes are reflected in associated_event field
+            planningApi.redux.store.dispatch<any>(eventsApi.receiveEvents([response]));
 
             return planningApi.planning.searchGetAll({
                 recurrence_id: events[0].recurrence_id,
--- a/client/components/Coverages/CoverageArrayInput.tsx
+++ b/client/components/Coverages/CoverageArrayInput.tsx
@@ -23,6 +23,7 @@ import {InputArray} from '../UI/Form';
 import {CoverageEditor} from './CoverageEditor';
 import {CoverageAddButton} from './CoverageAddButton';
 import planningActions from '../../actions/planning/api';
+import {EmptyState} from 'superdesk-ui-framework/react';
 
 
 interface IOwnProps {
@@ -202,6 +203,16 @@ class CoverageArrayInputComponent extends React.Component<IProps, IState> {
                 value={value}
                 coverages={value}
                 onChange={onChange}
+                emptyStateElement={(
+                    <EmptyState
+                        title={superdeskApi.localization.gettext('No coverages have been added')}
+                        description={
+                            superdeskApi.localization.gettext('To add some, click the plus icon at the top right')
+                        }
+                        illustration="2"
+                        size="small"
+                    />
+                )}
                 addButtonText={addButtonText}
                 addButtonComponent={CoverageAddButton}
                 addButtonProps={{
--- a/client/components/Coverages/CoverageArrayInput.tsx
+++ b/client/components/Coverages/CoverageArrayInput.tsx
@@ -207,7 +207,7 @@ class CoverageArrayInputComponent extends React.Component<IProps, IState> {
                     <EmptyState
                         title={superdeskApi.localization.gettext('No coverages have been added')}
                         description={
-                            superdeskApi.localization.gettext('To add some, click the plus icon at the top right')
+                            superdeskApi.localization.gettext('To add some, click the plus icon')
                         }
                         illustration="2"
                         size="small"
--- a/client/components/Planning/PlanningEditor/index.tsx
+++ b/client/components/Planning/PlanningEditor/index.tsx
@@ -339,6 +339,7 @@ class PlanningEditorComponent extends React.Component<IProps, IState> {
                         files: this.props.files,
                     },
                     associated_event: {
+                        removeEventItem: editor.item.planning.removeEventItem,
                         field: 'related_events',
                         events: (this.props.diff.related_events ?? [])
                             .map((relatedEvent) => this.props.events[relatedEvent._id]),
--- a/client/components/editor-standalone/save-handling.ts
+++ b/client/components/editor-standalone/save-handling.ts
@@ -13,7 +13,7 @@ const getEmbeddedAuthoringRefs = (editorType: EDITOR_TYPE, itemType: ItemType) =
     const embeddedEditorRef = planningApi.editor(editorType).dom.fields[fieldId]?.current;
     const relatedItemRefs: IRelatedItemRefs = embeddedEditorRef?.relatedItemRefs;
 
-    return (relatedItemRefs ?? {}) as IRelatedItemRefs;
+    return relatedItemRefs;
 };
 
 const getEmbeddedItemsExposed = (
@@ -22,9 +22,7 @@ const getEmbeddedItemsExposed = (
 ): Array<IExposedFromAuthoring<void>> => {
     const relatedItemRefs = getEmbeddedAuthoringRefs(editorType, itemType);
 
-
-    // Use Object.values instead of Object.keys because when refs are removed value becomes `null` and keys stay
-    if (Object.values(relatedItemRefs ?? {}).filter(superdeskApi.helpers.notNullOrUndefined).length < 1) {
+    if (Object.values(relatedItemRefs).filter(superdeskApi.helpers.notNullOrUndefined).length < 1) {
         return [];
     }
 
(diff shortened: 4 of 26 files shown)


thecalcc:
UI demo:

https://github.com/user-attachments/assets/6e45aa04-04c6-4408-8811-de666309e20d

client/api/events.ts
```diff
@@ -158,6 +159,9 @@ function create(updates: Partial<IEventItem>): Promise<Array<IEventItem>> {
         .then((response) => {
             const events = modifySaveResponseForClient(response);
 
+            // Update redux store, so changes are reflected in associated_event field
+            planningApi.redux.store.dispatch<any>(eventsApi.receiveEvents([response]));
```
tomaskikutis: shouldn't you call modifySaveResponseForClient before pushing it to redux store?

client/components/fields/editor/AssociatedEvent.tsx
```diff
@@ -58,10 +71,22 @@ export class EditorFieldAssociatedEventComponent extends React.PureComponent<IAs
...
             <div>
-                <label className="InputArray__label side-panel__heading side-panel__heading--big">
-                    {gettext('Related Events')}
-                </label>
-
+                <Spacer h gap="4" justifyContent="space-between" noWrap>
+                    <label className="InputArray__label side-panel__heading side-panel__heading--big">
+                        {gettext('Related Events')}
+                    </label>
+                    {disabled !== true && this.props.item._id.includes('temp') === false && (
```
tomaskikutis: > `this.props.item._id.includes('temp')`

This is too unreliable. For example this would match:

`c6f4ddc7-4fe2-436d-9b88-temp3bf98478`

We're probably using UUID v4 and only a-f letters are possible, but still, not a good practice. Use `isTemporaryId` helper that we have already.

client/components/fields/editor/AssociatedEventWrapper.tsx
```diff
@@ -22,7 +27,7 @@ export class AssociatedEventField extends React.PureComponent<IAssociatedEventFi
     }
 }
 
-const mapStateToProps = (state) => ({
+const mapStateToProps = (state): IReduxStoreProps => ({
```
tomaskikutis: locked items and files appear unused to me. It's not used inside `EditorFieldAssociatedEventComponent` and I don't see that entire props object would be passed to children.

tomaskikutis (commented):
Make plus button always visible, disable it when action is unavailable and add a message in tooltip. See my latest commit in feature branch where I did the same for planning items.

client/components/fields/editor/AssociatedEvent.tsx
```diff
@@ -75,7 +75,11 @@ export class EditorFieldAssociatedEventComponent extends React.PureComponent<IAs
                     <label className="InputArray__label side-panel__heading side-panel__heading--big">
                         {gettext('Related Events')}
                     </label>
-                    {disabled !== true && this.props.item._id.includes('temp') === false && (
+                    <Tooltip
+                        flow="left"
+                        text={gettext('Item not saved or you\'re in preview')}
```
tomaskikutis: If `disabled` is `true` we don't need a message. Main editor will handle adding indicator explaining why all fields are disabled.

The error message isn't clear which item you're referring to since we have related items. I used "Event has to be created before adding related plannings" maybe you could use a similar one.

client/components/fields/editor/AssociatedEventWrapper.tsx
tomaskikutis: seems you still getting files from store when they're not being used?

tomaskikutis:
check unresolved comment

thecalcc:
> check unresolved comment

fixed

client/components/fields/editor/AssociatedEventItem.tsx
```diff
@@ -3,16 +3,19 @@ import {authoringStorageEventItemHttp} from '../../../components/editor-standalo
 import {EventEditorStandalone} from '../../../components/editor-standalone/event-editor-standalone';
 import {RelatedEventListItem} from '../../../components/Events/EventMetadata/RelatedEventListItem';
 import React, {createRef} from 'react';
-import {ToggleBox} from 'superdesk-ui-framework/react';
-import {IPropsAuthoring} from 'superdesk-api';
+import {IconButton, ToggleBox} from 'superdesk-ui-framework/react';
+import {AuthoringReact} from 'apps/authoring-react/authoring-react';
```
tomaskikutis: See if getting the type would work this way:
```
type AuthoringReact<T extends IBaseRestApiResponse> =
    ReturnType<typeof superdeskApi.components.getAuthoringComponent<T>>
```

You might need to update also API definition from:
`getAuthoringComponent: <T extends IBaseRestApiResponse>() => React.ComponentType<IPropsAuthoring<T>>;`
to:
`getAuthoringComponent: <T extends IBaseRestApiResponse>() => React.ComponentClass<IPropsAuthoring<T>>;`

client/components/Planning/PlanningEditor/index.tsx
```diff
@@ -339,6 +339,7 @@ class PlanningEditorComponent extends React.Component<IProps, IState> {
                         files: this.props.files,
                     },
                     associated_event: {
+                        removeEventItem: editor.item.planning.removeEventItem,
```
tomaskikutis, line 342: wouldn't it work to move this to related events compoment itself?

client/components/editor-standalone/save-handling.ts
```diff
@@ -83,5 +86,9 @@ export const handleEmbeddedItems = async(
 export const embeddedItemHasUnsavedChanges = (itemType: ItemType) => {
     const planningsExposed = getEmbeddedItemsExposed(EDITOR_TYPE.INLINE, itemType);
 
-    return (planningsExposed ?? []).some((x) => x.hasUnsavedChanges());
+    if ((planningsExposed ?? []).length < 1) {
```
tomaskikutis: Does this change anything? It shouldn't since:

```
const colors = [];

const hasBlue = colors.some((color) => color === 'blue')

console.log(hasBlue) // false
```

client/components/Coverages/CoverageArrayInput.tsx
```diff
@@ -202,6 +203,16 @@ class CoverageArrayInputComponent extends React.Component<IProps, IState> {
                 value={value}
                 coverages={value}
                 onChange={onChange}
+                emptyStateElement={(
+                    <EmptyState
+                        title={superdeskApi.localization.gettext('No coverages have been added')}
+                        description={
```
tomaskikutis, line 209: It's better not to describe positions of elements in UI. We might change the positions and descriptions might get forgoten. I think users will be able to find the plus icon even without the position hint :smile:

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
```diff
@@ -49,6 +49,32 @@ export class EditorFieldEventRelatedPlanningsComponent extends React.PureCompone
             }
         })();
 
+        const PlanningItemsMetadata = () => {
```
tomaskikutis: No need to make this a component. Store as ReactNode.
```
const planningItemsMetadata = planningItems.length > 0 ? <div>...<div> : <div>...</div>
```

client/components/editor-standalone/save-handling.ts
```diff
@@ -22,9 +22,12 @@ const getEmbeddedItemsExposed = (
 ): Array<IExposedFromAuthoring<void>> => {
     const relatedItemRefs = getEmbeddedAuthoringRefs(editorType, itemType);
 
-    // Crashes whenever there's no embedded items
+    if (Object.values(relatedItemRefs).filter(superdeskApi.helpers.notNullOrUndefined).length < 1) {
```
tomaskikutis: > // Use Object.values instead of Object.keys because when refs are removed value becomes `null` and keys stay

This was a useful comment, I'd keep it. Or you used removeRefs and this is no longer the case? then go back to Object.keys

client/components/fields/editor/AssociatedEventItem.tsx
thecalcc: I don't think it would work. The ref needs a direct reference to the actual class being used. So that it can access it's properties. This definition doesn't include class properties, rather arguments you can pass to it through `props`

client/components/Planning/PlanningEditor/index.tsx
thecalcc: Probably, but I thought I'd follow the existing pattern - like in related plannings

client/components/editor-standalone/save-handling.ts
thecalcc: Correct, it's not needed. Previously I had it to prevent crashes of `undefined.hasUnsavedChanges`

client/components/Coverages/CoverageArrayInput.tsx
thecalcc: Agree, position based comments are bad. What do you suggest instead? Just "Add one by clicking the add button"? I can't think of something good right now that doesn't have repetitiveness

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
thecalcc: I have it for ease of readability

client/components/editor-standalone/save-handling.ts
thecalcc: I agree, I kept this because I didn't get to the bottom of why still some KV pairs with `null` values still showed up, even after adding cleanup functions

client/components/fields/editor/AssociatedEventItem.tsx
tomaskikutis: Ok, let's keep it for now. It would get solved if we manage to extract authoring-react into a separate package so it can be directly imported without having to call that function.

client/components/Planning/PlanningEditor/index.tsx
tomaskikutis: Try moving it then. I try to pass as little as possible through those props. Basically only value/onChange

client/components/Coverages/CoverageArrayInput.tsx
tomaskikutis: > To add some, click the plus icon ~at the top right~

Your message minus the position is fine

client/components/editor-standalone/save-handling.ts
tomaskikutis: I'd say it's fine to use Object.values and not even try to remove anything

client/components/Planning/PlanningEditor/index.tsx
thecalcc: Can't remove it. Editor instance is created at the top of this file, it has to be in sync with everything. That's why you passed it like that as well in related plannings

client/components/editor-standalone/save-handling.ts
thecalcc: OK, will then revert it

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
tomaskikutis: image

Is there a difference in readability here?

thecalcc: Oh this, all good, I can change it

tomaskikutis (approved):

tomaskikutis:
handle tests before merging. I think even e2e should pass here

tomaskikutis:
On target branch, only tests in suite A, file `event_embedded_coverage.cy.ts` are failing. The others probably broke in this PR. Fix them before merging.

PR #2195, opened 2025-02-05
Inline toggle for add to workflow

thecalcc:
## Front-end checklist

### DIFF
--- a/client/components/Coverages/CoverageEditor/CoverageForm.tsx
+++ b/client/components/Coverages/CoverageEditor/CoverageForm.tsx
@@ -364,14 +364,8 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
                 label: assignmentUtils.getContactLabel(this.props.value),
             },
             add_coverage_to_workflow: {
-                field: 'add_coverage_to_workflow',
                 onChange: this.onAddToWorkflowChange,
-                label: superdeskApi.localization.gettext('Add to workflow'),
-                disabled:
-                    planningUtils.canAddCoverageToWorkflow(this.props.value, this.props.diff) !== true
-                    || this.props.value.add_coverage_to_workflow === true
-                    || this.props.value.assigned_to.user == null
-                    || this.props.value.assigned_to.desk == null,
+                disabled: planningUtils.canAddCoverageToWorkflow(this.props.value, this.props.diff) !== true
             },
             g2_content_type: {
                 readOnly: this.props.readOnly || readOnlyFields.g2_content_type,
--- a/client/components/Coverages/CoverageEditor/CoverageForm.tsx
+++ b/client/components/Coverages/CoverageEditor/CoverageForm.tsx
@@ -20,7 +20,6 @@ import {
 } from '../../../interfaces';
 
 import * as selectors from '../../../selectors';
-import * as actions from '../../../actions';
 import {planningUtils, generateTempId, assignmentUtils} from '../../../utils';
 
 import {WORKFLOW_STATE} from '../../../constants';
@@ -50,6 +49,7 @@ interface IProps {
     includeScheduledUpdates?: boolean;
     editorType: EDITOR_TYPE;
     language: IVocabularyItem['qcode'];
+    coverages: Array<IPlanningCoverageItem>;
 
     // Functions
     onChange(field: string, value: any): void;
@@ -110,6 +110,8 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
         this.onAddXmpFile = this.onAddXmpFile.bind(this);
         this.onRemoveXmpFile = this.onRemoveXmpFile.bind(this);
         this.onContentTypeChange = this.onContentTypeChange.bind(this);
+        this.onAddToWorkflowChange = this.onAddToWorkflowChange.bind(this);
+
         this.dom = {
             contentType: React.createRef(),
             popupContainer: null,
@@ -310,6 +312,29 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
         }
     }
 
+    // Only used to add to workflow.
+    // Removing from workflow is done if the user removes the assignee.
+    onAddToWorkflowChange() {
+        const {vocabulary} = superdeskApi.entities;
+        const coverageStatuses =
+            vocabulary.getAll().get('newscoveragestatus').items as Array<IPlanningNewsCoverageStatus>;
+        const updatedCoverage = planningUtils.addCoverageToWorkflow(this.props.value, coverageStatuses);
+
+        this.onChange(
+            'workflow_status',
+            updatedCoverage.workflow_status,
+        );
+        this.onChange(
+            'news_coverage_status',
+            updatedCoverage.news_coverage_status,
+        );
+        this.onChange(
+            'assigned_to.state',
+            updatedCoverage.assigned_to.state,
+        );
+        this.onChange('add_coverage_to_workflow', updatedCoverage.add_coverage_to_workflow);
+    }
+
     render() {
         const contentTypeQcode = this.props.value.planning?.g2_content_type;
         const defaultGenre = (appConfig.default_genre || [{}])[0];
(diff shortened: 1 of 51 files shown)


client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -338,6 +363,16 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
                 assignmentField: 'assigned_to.contact',
                 label: assignmentUtils.getContactLabel(this.props.value),
             },
+            add_coverage_to_workflow: {
+                field: 'add_coverage_to_workflow',
+                onChange: this.onAddToWorkflowChange,
+                label: superdeskApi.localization.gettext('Add to workflow'),
+                disabled:
+                    planningUtils.canAddCoverageToWorkflow(this.props.value, this.props.diff) !== true
+                    || this.props.value.add_coverage_to_workflow === true
```
tomaskikutis: how about moving these checks to `canAddCoverageToWorkflow`?

client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
```diff
@@ -83,7 +83,7 @@ class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
         /*
             Check if:
             1. This view is rendered from AddToPlanning action
-            2. There's an already scheduled update for the coverage
+            2. The item isn't a scheduled update
```
tomaskikutis: I think it'd be cleaner to use variable names as documentation here e.g.
`const itemIsScheduledUpdate = !((value as ICoverageScheduledUpdate).scheduled_update_id)`

client/components/fields/editor/AddCoverageToWorkflow.tsx
```diff
@@ -0,0 +1,20 @@
...
+import {superdeskApi} from '../../../superdeskApi';
+
+export class EditorFieldAddCoverageToWorkflow extends React.PureComponent<IEditorFieldProps> {
+    render() {
+        const {gettext} = superdeskApi.localization;
+
+        return (
+            <EditorFieldToggle
+                {...this.props}
+                field={this.props.field ?? 'add_coverage_to_workflow'}
```
tomaskikutis: `EditorFieldAddCoverageToWorkflow` seems like a very specific field. Don't accept field id or label from props.

client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -338,6 +363,16 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
                 assignmentField: 'assigned_to.contact',
                 label: assignmentUtils.getContactLabel(this.props.value),
             },
+            add_coverage_to_workflow: {
+                field: 'add_coverage_to_workflow',
+                onChange: this.onAddToWorkflowChange,
```
tomaskikutis: would it work to move all props besides ID into the component itself?

client/interfaces.ts
```diff
@@ -646,6 +646,7 @@ export interface IPlanningCoverageItem {
     version_creator: string;
     firstcreated: string;
     versioncreated: string;
+    add_coverage_to_workflow: boolean;
```
tomaskikutis, line 649: Can't it be checked whether coverage is in workflow using other fields? If so, maybe we don't need an extra one that stores computed data?

client/components/Coverages/CoverageEditor/CoverageForm.tsx
thecalcc: Besides `onChange` and `disabled` as those depend on prop values.

client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -310,6 +312,29 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
         }
     }
 
+    // Only used to add to workflow.
+    // Removing from workflow is done if the user removes the assignee.
+    onAddToWorkflowChange() {
```
tomaskikutis: when you toggle from on to off, does this get called too? I don't see conditions whether it's toggling on or off here

client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
```diff
@@ -198,6 +193,8 @@ class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
                                         {
                                             ...value,
                                             assigned_to: {},
+                                            add_coverage_to_workflow: false,
```
tomaskikutis, line 196: add a comment that those 2 fields are for removing a coverage from workflow

client/utils/planning.tsx
```diff
@@ -377,13 +377,14 @@ function canCancelCoverage(
...
-function canAddCoverageToWorkflow(coverage: IPlanningCoverageItem, planning: IPlanningItem): boolean {
+function canAddCoverageToWorkflow(coverage: IPlanningCoverageItem, planning: Partial<IPlanningItem>): boolean {
     return (
         isExistingItem(coverage, 'coverage_id') &&
         isCoverageDraft(coverage) &&
         isCoverageAssigned(coverage) &&
         !appConfig.planning_auto_assign_to_workflow &&
-        !isItemExpired(planning)
+        !isItemExpired(planning) &&
+        this.props.value.add_coverage_to_workflow === true
```
tomaskikutis: `this.props` doesn't exist here

client/components/Coverages/CoverageEditor/CoverageForm.tsx
thecalcc: No, it's not getting called when you want to remove it from workflow. You can do that only by removing the assignment

client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
thecalcc: No need. This will be merged to coverage fixes and this code will be removed. On that branch it's clear what's happening

client/utils/planning.tsx
thecalcc: Fixed this

client/components/Coverages/CoverageEditor/CoverageForm.tsx
tomaskikutis: I though that's the point of the toggle, that you can **toggle** it :laughing: Can you update the code so it works?

client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -310,6 +311,69 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
...
     }
 
+    onAddToWorkflowChange() {
+        const {vocabulary} = superdeskApi.entities;
+        const {gettext} = superdeskApi.localization;
+        const coverageStatuses = vocabulary
+            .getAll()
+            .get('newscoveragestatus').items as Array<IPlanningNewsCoverageStatus>;
+        const updatedCoverage = planningUtils.addCoverageToWorkflow(this.props.value, coverageStatuses);
+        const filteredCoverages = this.props.coverages.filter((x) => x.coverage_id !== updatedCoverage.coverage_id);
```
tomaskikutis: Choose a better variable name here that would better describe the value it holds and what you're trying to achieve with it. I see that you're filtering it here, spreading it there, but I don't get the entire picture of what you're doing.

After implementing actual toggling, would this function also be called when removing item from workflow? It would need to be renamed then.

client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -310,6 +311,69 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
...
+                        },
+                        add_coverage_to_workflow: add_coverage_to_workflow,
+                    },
+                ],
+            );
+        } else {
+            showModal(({closeModal}) => (
+                <Modal
+                    visible
+                    headerTemplate={gettext('Are you sure?')}
```
tomaskikutis: We have a `confirm` method in superdesk-api.d.ts that you could use here. Also it's a good UX practice to phrase modal titles in such a way so it's clear what the modal is about, even if we assume that user forgot which action he clicked to iniate the modal. It also helps to reassure users that they are indeed confirming the intended thing.

tomaskikutis (approved):
you can merge with e2e failing since it's also broken on target branch

tomaskikutis:
after reviewing I forgot to switch branch and I was running this one and noticed that planning items are not saving. Look into that.

Might be related to this error I was seeing:

> ItemManager.ts:121 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

tomaskikutis:
also this 
image

thecalcc:
> also this image

I continue testing this and it's not happening for me. In your case it might've been old state being persisted from some other place

tomaskikutis:
I tried again and indeed I can't reproduce it anymore with new items. It might have been a corrupted item I had.

tomaskikutis (approved):

PR #2196, opened 2025-02-05
[SDESK-7483] - Create Event & Planning Autosave resources, services and REST APIs

BrianMwangi21:
### Purpose
This PR creates new Event & Planning Autosave resources, services and REST APIs:

- Implementing a custom AsyncResourceService for common autosave functionality
- Implementing new async resource service for EventAutosave & PlanningAutosave resources
- Implementing resource config for the new resources and register them in the planning module
- Disable the API in old Autosave resources. Added TODO to remove the old resources where they are used in other services that are yet to be converted to async.

Solves: [SDESK-7483](https://sofab.atlassian.net/browse/SDESK-7483)

[SDESK-7483]: https://sofab.atlassian.net/browse/SDESK-7483


BrianMwangi21:
@MarkLark86, 

The previous EventAutosaveResource and PlanningAutosaveResource made use of the `event_schema` and `planning_schema` respectively as they were. Hence, I've also used the existing `EventResourceModel` and `PlanningResourceModel` for their resource configs instead of creating new types. 

Is that ok ?

server/planning/autosave_async_service.py
```diff
@@ -0,0 +1,37 @@
...
+
+class AutosaveAsyncService(AsyncResourceService):
+    """Async Service class for the Autosave model."""
+
+    async def on_create(self, docs: list[EventResourceModel | PlanningResourceModel]) -> None:
+        await super().on_create(docs)
+
+        for doc in docs:
+            self._validate(doc)
+            delattr(doc, "expired")
```
MarkLark86: I tested this locally, and might cause an issue when accessing that value.
```python
>>> from pydantic import BaseModel
>>> class Person(BaseModel):
...     name: str
...     expired: bool | None = None
... 
>>> p = Person(name="Mark", expired=True)
>>> p
Person(name='Mark', expired=True)
>>> delattr(p, "expired")
>>> p
Person(name='Mark')
>>> p.expired
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "lib/python3.10/site-packages/pydantic/main.py", line 891, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'Person' object has no attribute 'expired'
>>>
```

As the `expired` field is defined as `expired: bool = False`, it would make sense then to assign a value of `False` here instead

server/planning/events/module.py
```diff
@@ -46,3 +47,16 @@
...
     rest_endpoints=RestEndpointConfig(resource_methods=["GET"], item_methods=["GET"]),
 )
+
+events_autosave_resource_config: ResourceConfig = ResourceConfig(
+    name="event_autosave",
+    data_class=EventResourceModel,
+    service=EventsAutosaveAsyncService,
+    mongo=MongoResourceConfig(
+        indexes=[
+            MongoIndexOptions(name="event_autosave_user", keys=[("lock_user", 1)], background=True),
```
MarkLark86: Will also need to explicitly set `unique=False` for these 2, and the 2 for Planning autosave

server/planning/planning/planning_autosave_async_service.py
```diff
@@ -0,0 +1,35 @@
...
+
+        if not item:
+            # Item is not currently being edited (No current autosave item)
+            return
+
+        coverages = item.get("coverages") or []
+        coverage = next((c for c in coverages if c.get("coverage_id") == coverage_id), None)
+
+        if not coverage:
+            logger.warn("Coverage {} not found in autosave for item {}".format(coverage_id, planning_id))
```
MarkLark86: The `warn` method is deprecated, use `warning` instead

server/planning/autosave_async_service.py
```diff
@@ -0,0 +1,37 @@
...
+    def _validate(doc: EventResourceModel | PlanningResourceModel):
+        """Validate the autosave to ensure it contains user/session"""
+
+        if not doc.lock_user:
+            raise SuperdeskApiError.badRequestError(message="Autosave failed, User not supplied")
+
+        if not doc.lock_session:
+            raise SuperdeskApiError.badRequestError(message="Autosave failed, User Session not supplied")
+
+    async def on_session_end(self, user_id, session_id, is_last_session):
```
MarkLark86: It would be good to update the old `on_session_end` callbacks.

**1. Remove the old callback registration:**
* Planning: https://github.com/superdesk/superdesk-planning/blob/async/server/planning/planning/__init__.py#L149
* Events: https://github.com/superdesk/superdesk-planning/blob/async/server/planning/events/__init__.py#L159

**2. Add a single cleanup function:**
Add a single function that cleans up after a session ends by adding an init function in the [Planning module](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/module.py#L14)

```python
from asyncio import gather

from bson import ObjectId

from superdesk.core.module import Module, SuperdeskAsyncApp

from planning.events.event_autosave import EventsAutosaveAsyncService
from planning.planning.planning_autosave_async_service import PlanningAutosaveAsyncService


async def cleanup_on_session_end(user_id: ObjectId, session_id: ObjectId, is_last_session: bool) -> None:
    lookup = {LOCK_USER: user_id} if is_last_session else {LOCK_SESSION: session_id}

    # Using ``asyncio.gather`` allows these 2 to run in the 1 asyncio event loop cycle
    await gather(
        EventsAutosaveAsyncService().delete_many(lookup),
        PlanningAutosaveAsyncService().delete_many(lookup),
    )


def init_planning(app: SuperdeskAsyncApp):
    wsgi_app = app.wsgi.as_any()
    wsgi_app.on_session_end += cleanup_on_session_end


module = Module(
    "planning",
    init=init_planning,
    ...
)
```

The `app.on_` should now handle async callbacks. We have not done something like this yet, but it should work.

MarkLark86:
> @MarkLark86,
> 
> The previous EventAutosaveResource and PlanningAutosaveResource made use of the `event_schema` and `planning_schema` respectively as they were. Hence, I've also used the existing `EventResourceModel` and `PlanningResourceModel` for their resource configs instead of creating new types.
> 
> Is that ok ?

It might not be.
In the ResourceModel we have the `model_resource_name` ClassVar and `get_service` class method. These rely on the name of the registered/configured resource config (which is set when the app runs/resource is registered)

It might be okay to inherit from it though, try adding them to `planning.types.autosave` (and export them in `planning.types.__init__`):
```python
# file: planning/types/autosave.py
from .events import EventResourceModel
from. planning import PlanningResourceModel

class EventAusoaveResource(EventResourceModel):
    pass

class PlanningAutosaveResource(PlanningResourceModel):
    pass
```

And use those 2 resource models in the service and config for registration. Hopefully that should isolate the 4 models

server/planning/autosave_async_service.py
BrianMwangi21: This is actually pretty neat 🔥

MarkLark86 (approved):

PR #2200, opened 2025-02-10
Update empty states

thecalcc:
STT-31

## Front-end checklist

### DIFF
--- a/client/components/Coverages/CoverageArrayInput.tsx
+++ b/client/components/Coverages/CoverageArrayInput.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import {connect} from 'react-redux';
-import {isEmpty, noop} from 'lodash';
+import {isEmpty} from 'lodash';
 
 import {
     EDITOR_TYPE,
@@ -192,7 +192,6 @@ class CoverageArrayInputComponent extends React.Component<IProps, IState> {
 
         const {desks, users, coverageAddAdvancedMode} = this.props;
         const language = this.props.item.language;
-        const {DropZone} = superdeskApi.components;
 
         return (
             <InputArray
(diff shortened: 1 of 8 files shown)


client/components/fields/editor/AssociatedEvent.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-nested-ternary */
```
tomaskikutis: this is probably not needed? I don't see any added nested terneries

client/components/Coverages/CoverageArrayInput.tsx
```diff
@@ -204,14 +204,9 @@ class CoverageArrayInputComponent extends React.Component<IProps, IState> {
...
                 emptyStateElement={(
-                    <EmptyState
-                        title={superdeskApi.localization.gettext('No coverages have been added')}
-                        description={
-                            superdeskApi.localization.gettext('To add some, click the plus icon')
-                        }
-                        illustration="2"
-                        size="small"
-                    />
+                    <div className="sd-border--medium border-dashed p-1-5 radius-md d-flex items-center justify-center">
```
tomaskikutis: use a disabled dropzone here

thecalcc: Fritz doesn't want one, he wants this one specifically, and we don't have it as a component

tomaskikutis: I discussed it with him, he's now fine with a dropzone

client/components/Coverages/CoverageArrayInput.tsx
```diff
@@ -203,15 +203,10 @@ class CoverageArrayInputComponent extends React.Component<IProps, IState> {
...
-                emptyStateElement={(
-                    <EmptyState
-                        title={superdeskApi.localization.gettext('No coverages have been added')}
-                        description={
-                            superdeskApi.localization.gettext('To add some, click the plus icon')
-                        }
-                        illustration="2"
-                        size="small"
-                    />
+                emptyValueElement={(
```
tomaskikutis: what's the purpose to pass it as a prop when dropzone can be imported from superdesk api at any time?

thecalcc: no purpose, I didn't check if `InputArray` was used somewhere else, turns out it isn't.

tomaskikutis (approved):

PR #2202, opened 2025-02-11
[SDESK-7486] - Event update time endpoint

BrianMwangi21:
### Purpose
This PR creates a new async Event update time endpoint:

- Implementing a new async Event update time endpoint as a view function including validations and sending notifications
- Add new auth rule for "planning_event_management". 
- Deleted the API in old EventUpdateTime resources. 

Solves: [SDESK-7486](https://sofab.atlassian.net/browse/SDESK-7486)

[SDESK-7486]: https://sofab.atlassian.net/browse/SDESK-7486


BrianMwangi21:
@MarkLark86 a few questions: 

1. For the auth rules. do I need to implement something similar to the `auth_rules` in newsroom-core so that it can subsequently be used by other blueprints that might be done in this planning module ?

2. I have disabled the old resource for now as it is used in the `events_tests` file and includes some dot notation assignments like the example below and I'm not sure how that translates to an endpoint : 
```
        update_time = get_resource_service("events_update_time")
        update_time.REQUIRE_LOCK = False

```

3. Does the `get_current_app` usage when updating recurring events remain as ? or is there a change when used in async function ? 

Thanks.

MarkLark86:
> 1. For the auth rules. do I need to implement something similar to the `auth_rules` in newsroom-core so that it can subsequently be used by other blueprints that might be done in this planning module ?

You should be able to use the new [required_privilege_rule](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/auth/privilege_rules.py#L9) auth rule. Try it with
`auth=[required_privilege_rul("planning_event_management")]`

MarkLark86:
> 2. I have disabled the old resource for now as it is used in the `events_tests` file and includes some dot notation assignments like the example below and I'm not sure how that translates to an endpoint :
> 
> ```
>         update_time = get_resource_service("events_update_time")
>         update_time.REQUIRE_LOCK = False
> ```

Good point, I missed the use of `get_resource_service` for this action resources. It doesn't make sense to have these developed as a resource/service, as they don't directly manage storage of a resource. That is why I recommended using a web endpoint instead.

What might make sense here is where `get_resource_service("events_update_time")` is used, replace that with a function call (common function between the web endpoint and other usages within the code). The `REQUIRE_LOCK` argument to the old class could simply be an argument in said function.

MarkLark86:
> 3. Does the `get_current_app` usage when updating recurring events remain as ? or is there a change when used in async function ?

No, `get_current_app` is still relevant here. Ideally things like `app.on_updated_events_update_time` should really be updated to use the [new signals](https://superdesk.readthedocs.io/en/async/core/signals.html).

In this case it is the [app.on_updated_events_update_time += events_history_service.on_update_time](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/events/__init__.py#L159) call which would be replaced with a signal.

Might be good to add a file `planning/events/signals`, and then in other parts of the code subscribe to it and react (to add history entries for this action). Does that make sense?

BrianMwangi21:
> Might be good to add a file `planning/events/signals`, and then in other parts of the code subscribe to it and react (to add history entries for this action). Does that make sense?

Could you clarify more on this ?

MarkLark86:
> > Might be good to add a file `planning/events/signals`, and then in other parts of the code subscribe to it and react (to add history entries for this action). Does that make sense?
> 
> Could you clarify more on this ?

Actually we already have a [signals](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/signals.py) file that should be used.

So you would add the following to planning/signals.py:
```python
event_time_updated = AsyncSignal[dict, dict]("events:time_updated")
```

And then in planning/events/views.py, replace
`app.on_updated_events_update_time(new_updates, {"_id": event[ID_FIELD]})`
with
`await signals.event_time_updated.send(new_updates, {"_id": event[ID_FIELD]})`

And then in planning/events/\_\_init__.py, replace
`app.on_updated_events_update_time += events_history_service.on_update_time`
with
`signals.event_time_updated.connect(events_history_service.on_update_time)`

See how that goes. This should be the procedure to replace old signals with new signals (required as the new signals supports async functionality, which would be needed from time to time)

BrianMwangi21:
@MarkLark86 I found one more use of `events_update_time` in `client/actions/events/api.ts` : 

```
        return api.update(
            'events_update_time',
            original,
            {
                update_method: get(updates, 'update_method.value', EVENTS.UPDATE_METHODS[0].value),
                dates: updates.dates,
                [TO_BE_CONFIRMED_FIELD]: false,
            }
        );

```

Does it also need change to `events/update_time` ?

server/planning/events/events_utils.py
```diff
@@ -203,7 +213,17 @@ async def get_recurring_timeline(
...
+            selected_start = datetime.strptime(selected_start, "%Y-%m-%dT%H:%M:%S%z")
+        except ValueError:
+            # Fallback for strings without timezone information:
+            selected_start = datetime.strptime(selected_start, "%Y-%m-%dT%H:%M:%S")
+            # If there's a timezone field in the event, use it:
+            tz_str = selected.get("dates", {}).get("tz")
+            if tz_str:
+                selected_start = pytz.timezone(tz_str).localize(selected_start)
+            else:
+                selected_start = selected_start.replace(tzinfo=pytz.UTC)
```
eos87: I think there is chance of another `ValueError` if for some reason the format does not match with the date string. I would suggest using `arrow` as it is already used in the project. Something like this `arrow.get(selected_start).datetime` would do instead of the `datetime.strptime` that's inside the except. What do you think?

server/planning/events/events_utils.py
```diff
@@ -220,3 +240,107 @@ async def get_recurring_timeline(
...
+    # First find the hour and minute of the start date in local time
+    start_time = utc_to_local(timezone, updates["dates"]["start"]).time()
+
+    # Next convert that to seconds since midnight (which gives us a timedelta instance)
+    delta_since_midnight = datetime.combine(date.min, start_time) - datetime.min
+
+    # And calculate the new duration of the events
+    duration = updates["dates"]["end"] - updates["dates"]["start"]
+
+    app = get_current_app().as_any()
```
eos87: This does not seem to be used. 
```suggestion
```

eos87 (approved):
Looks good, I just left a minor comment.

server/planning/utils.py
```diff
@@ -287,3 +288,10 @@ def is_coverage_assignment_modified(updates: dict[str, Any], original: dict[str,
             return True
 
     return False
+
+
+async def get_json_or_400_async(req: Request):
```
MarkLark86: A return type here would be good, to allow type checkers and IDEs to know it returns a dict

server/planning/events/views.py
```diff
@@ -0,0 +1,39 @@
...
+
+class EventsArgs(BaseModel):
+    event_id: str
+
+
+@blueprint.endpoint(
+    "/update_time/<string:event_id>", methods=["PATCH"], auth=[required_privilege_rule("planning_event_management")]
+)
+async def update_time(args: EventsArgs, params: None, request: Request) -> Response:
+    original = await EventsAsyncService().find_by_id_raw(args.event_id)
```
MarkLark86, line 23: Rather than using the `EventsAsyncService` directly, you can use the model type instead.
i.e.
```python

async def update_time(...):
    original = await EventResourceModel().get_service().find_by_id_raw(args.event_id)
```

Although it's not relevant here, using the type to get the service helps to reduce cyclic imports (such as an endpoint importing a service which imports a util which imports another service which import the first service). Using the types get's around this problem

server/planning/events/events_utils.py
```diff
@@ -220,3 +240,107 @@ async def get_recurring_timeline(
             future.append(event.to_dict())
 
     return historic, past, future
+
+
+async def update_single_event(updates: dict[str, Any]):
```
MarkLark86: These functions are specific to the update time action, and would be a good idea to keep them together in the same file. Also because we will be converting more and more event & planning action endpoint from a resource service to an endpoint

server/planning/events/events_utils.py
```diff
@@ -220,3 +240,107 @@ async def get_recurring_timeline(
...
+    from planning.events.events_service import EventsAsyncService
+
+    events_service = EventsAsyncService()
+
+    # Determine update method, ensuring non-recurring events use UPDATE_SINGLE
+    update_method = updates.pop("update_method", UPDATE_SINGLE)
+    if not original.get("dates", {}).get("recurring_rule"):
+        update_method = UPDATE_SINGLE
+
+    if update_method == UPDATE_SINGLE:
```
MarkLark86: There's functionality missing from the original implementation.

* Sending [notifications](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/events/events_base_service.py#L109)
* [top level validation](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/events/events_base_service.py#L130).

Please have a look over the original EventsUpdateTimeService and EventsBaseService, specifically the `on_update`, `update` and `on_updated` methods to make sure your implementation has all the required pieces.

Without these, I fear things like event locks will not be checked.

MarkLark86 (approved):
Looks good

PR #2203, opened 2025-02-11
Coverage validation for embedded events

tomaskikutis:
1. Validates coverages when editing an embedded planning item
2. `handleEmbeddedItems` function is updated not to handle scrolling to invalid fields anymore since it is now done by authoring-react

DEPENDS ON: https://github.com/superdesk/superdesk-client-core/pull/4751

## Front-end checklist

### DIFF
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -567,32 +567,9 @@ function updateLinkedPlanningsForEvent(
 
         const toUnlink: Array<IPlanningItem> = currentlyLinked
             .filter((item) => {
-                const createdAt = new Date(item._created);
-                const now = new Date();
+                const needToUnlink = associatedPlannings.find(({_id}) => _id === item._id) == null;
 
-                const ageSeconds = (now.getTime() - createdAt.getTime()) / 1000;
-                const tooRecent = ageSeconds < 30;
-
-                if (tooRecent) {
-                    /**
-                     * This is a hack to workaround our existing "fake ID" workaround.
-                     * In event editor it is possible to create a planning item and relate it to current event at once.
-                     * It would happen only after saving, thus while it's not saved yet, we use a fake ID
-                     * - which will not remain the same after saving.
-                     * This function computes a list of planning items which have to be linked
-                     * and which have to be unlinked based on a desired outcome
-                     * which is that only items specified in {@link associatedPlannings} must remain linked.
-                     * The problem arises that when item with a fake ID is saved, and ID changes,
-                     * that item will immediately get unlinked by this function, because there is no way that
-                     * the new ID could have been a part of {@link associatedPlannings}
-                     * (which is computed before saving).
-                     */
-                    return false;
-                } else {
-                    const needToUnlink = associatedPlannings.find(({_id}) => _id === item._id) == null;
-
-                    return needToUnlink;
-                }
+                return needToUnlink;
             });
 
         return Promise.all(
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -566,11 +566,7 @@ function updateLinkedPlanningsForEvent(
             associatedPlannings.filter(({_id}) => currentLinkedIds.has(_id) !== true);
 
         const toUnlink: Array<IPlanningItem> = currentlyLinked
-            .filter((item) => {
-                const needToUnlink = associatedPlannings.find(({_id}) => _id === item._id) == null;
-
-                return needToUnlink;
-            });
+            .filter((item) => associatedPlannings.find(({_id}) => _id === item._id) == null);
 
         return Promise.all(
             [
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -20,6 +20,7 @@ import {
     isPublishedItemId,
     isTemporaryId,
     gettext,
+    planningUtils,
 } from '../../utils';
 
 import planningApis from '../planning/api';
@@ -661,7 +662,7 @@ const save = (original, updates) => (
                 eventUpdates.update_method?.value ?? eventUpdates.update_method;
 
             const [planningsToCreate, planningsToUpdate] = partition(
-                eventUpdates.associated_plannings ?? [],
+                (eventUpdates.associated_plannings ?? []).map((planning) => planningUtils.modifyForServer(planning)),
                 (item) => item._id.startsWith(TEMP_ID_PREFIX),
             );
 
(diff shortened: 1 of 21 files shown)


client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
```diff
@@ -139,6 +140,20 @@ export class RelatedPlanningItem extends React.PureComponent<IProps> {
...
+                            if (this.toggleBoxRef.current.isOpen()) {
+                                return Promise.resolve();
+                            } else {
+                                return new Promise((resolve) => {
+                                    this.toggleBoxRef.current.toggle();
+
+                                    // PR-TODO: improve toggleBox so `toggle` method returns a promise
+                                    setTimeout(() => {
+                                        resolve();
+                                    }, 500);
```
thecalcc, line 153: Why do you need promises at all, since the toggling itself is synchronous? I see you're doing `makeVisible.then` in the client-core PR but it would work for things to run synchronously

client/validators/planning.ts
```diff
@@ -26,10 +29,35 @@ const validatePlanningScheduleDate = ({getState, field, value, errors, messages,
...
+        value: value,
+        diff: {
+            coverages: value,
+        },
+        item: {},
+        errors: errors,
+        messages: messages,
+    });
+
+    return {errors, messages};
```
thecalcc, line 55: I'm looking at the client core code as well, and I wonder why do you need messages and errors both? Also I guess you don't want to refactor `validateCoverages` for compatibility reasons? Maybe it's better if we validate only the first coverage, since you always take `[0]` from the errors object.

client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
tomaskikutis: Toggling isn't synchronous. I need to wait for animation to finish before scrolling into view. Otherwise scrolling will interfere with animation and get scrolled to a wrong place.

client/validators/planning.ts
tomaskikutis: > I'm looking at the client core code as well, and I wonder why do you need messages and errors both?

That's how the validator I'm wrapping works. And it is useful. `errors` is an object with issues by field. `messages` is an array of strings with validation issues. I'm using `messages` now, while it would be a good improvement to use `errors`. Those improvements are time consuming so I can't do them now.

> Maybe it's better if we validate only the first coverage, since you always take [0] from the errors object.

I'm displaying the first validation error from the array. It might be coming from any coverage.

thecalcc:
Also don't forget to update with latest client-core develop

client/actions/events/api.ts
```diff
@@ -661,7 +662,7 @@ const save = (original, updates) => (
                 eventUpdates.update_method?.value ?? eventUpdates.update_method;
 
             const [planningsToCreate, planningsToUpdate] = partition(
-                eventUpdates.associated_plannings ?? [],
+                (eventUpdates.associated_plannings ?? []).map((planning) => planningUtils.modifyForServer(planning)),
```
thecalcc, line 633: why do you modify for server here? Wouldn't they save without this?

tomaskikutis: Yes, it wouldn't save. It would complain that field `_scheduledTime` does not exist on the server, which is correct.

tomaskikutis:
I've ran all e2e tests failing at `ca376831`. The only one that reproduced was `planning/featured_planning.cy.ts`

The issue was that HTTP request to `http://localhost:5000/api/planning_featured/20250215` was returning 404

It might be related to browser sending a request for `0215` - February 15, while I have 14th selected in planning view.

thecalcc:
> I've ran all e2e tests failing at `ca376831`. The only one that reproduced was `planning/featured_planning.cy.ts`
> 
> The issue was that HTTP request to `http://localhost:5000/api/planning_featured/20250215` was returning 404
> 
> It might be related to browser sending a request for `0215` - February 15, while I have 14th selected in planning view.

The same tests are occasionally failing on my branches, and sometimes get fixed when I run them again

thecalcc (approved):

client/actions/events/api.ts
```diff
@@ -567,32 +567,9 @@ function updateLinkedPlanningsForEvent(
...
-                     * which is that only items specified in {@link associatedPlannings} must remain linked.
-                     * The problem arises that when item with a fake ID is saved, and ID changes,
-                     * that item will immediately get unlinked by this function, because there is no way that
-                     * the new ID could have been a part of {@link associatedPlannings}
-                     * (which is computed before saving).
-                     */
-                    return false;
-                } else {
-                    const needToUnlink = associatedPlannings.find(({_id}) => _id === item._id) == null;
+                const needToUnlink = associatedPlannings.find(({_id}) => _id === item._id) == null;
```
thecalcc: maybe a better name would be `toBeUnlinked` or `needToBeUnlinked`

client/components/editor-standalone/save-handling.ts
```diff
@@ -3,6 +3,7 @@ import {IExposedFromAuthoring} from 'superdesk-api';
 import {planningApi, superdeskApi} from '../../superdeskApi';
 import {RelatedPlanningItem} from '../../components/fields/editor/EventRelatedPlannings/RelatedPlanningItem';
 import {AssociatedEventItem} from 'components/fields/editor/AssociatedEventItem';
+import {notNullOrUndefined} from 'core/helpers/typescript-helpers';
```
thecalcc: do not import from superdesk use `superdeskApi`

client/components/editor-standalone/save-handling.ts
```diff
@@ -30,9 +31,11 @@ const getEmbeddedItemsExposed = (
         return [];
     }
 
-    const exposedAuthoringArray = Object.values(relatedItemRefs).map((x) =>
-        x?.authoringRef?.current?.getExposed?.() as unknown as IExposedFromAuthoring<void>
-    );
+    const exposedAuthoringArray = Object.values(relatedItemRefs)
```
thecalcc: what's the bug here? I thought this is handled on line 29. If not it might be good to do that filtering right away on line 24 where `getEmbeddedAuthoringRefs` is called

tomaskikutis: Ref itself wasn't `null`, but `ref.current` was `null`.

client/components/editor-standalone/save-handling.ts
```diff
@@ -23,15 +23,7 @@ const getEmbeddedItemsExposed = (
...
-    // Use Object.values instead of Object.keys because when refs are removed value becomes null and keys stay
-    if (
-        relatedItemRefs == null
-        || Object.values(relatedItemRefs).filter(superdeskApi.helpers.notNullOrUndefined).length < 1
-    ) {
-        return [];
-    }
-
-    const exposedAuthoringArray = Object.values(relatedItemRefs)
+    const exposedAuthoringArray = Object.values(relatedItemRefs ?? {})
```
thecalcc, line 26: nice improvement

thecalcc (approved):

PR #2206, opened 2025-02-14
Fix `CORS` issue and `hits` missing data

eos87:
- Fixes a CORS problem generated when the frontend tried to access `agenda` and `planning_types` resources from the planning screen
- Fixes missing `hits` data provided to the `ElasticsearchResourceCursorAsync` in `planning_types` service

image

### DIFF
--- a/server/planning/agendas_async/module.py
+++ b/server/planning/agendas_async/module.py
@@ -10,5 +10,9 @@ agendas_resource_config = ResourceConfig(
     name="agenda",
     data_class=AgendasResourceModel,
     service=AgendasAsyncService,
-    rest_endpoints=RestEndpointConfig(resource_methods=["GET", "POST"], item_methods=["GET", "PATCH", "DELETE"]),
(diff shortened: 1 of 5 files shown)


server/planning/agendas_async/module.py
```diff
@@ -10,5 +10,9 @@
     name="agenda",
     data_class=AgendasResourceModel,
     service=AgendasAsyncService,
-    rest_endpoints=RestEndpointConfig(resource_methods=["GET", "POST"], item_methods=["GET", "PATCH", "DELETE"]),
+    rest_endpoints=RestEndpointConfig(
+        resource_methods=["GET", "POST"],
+        item_methods=["GET", "PATCH", "DELETE"],
+        enable_cors=True,
```
MarkLark86, line 16: We can leave these `enable_cores=True` in here, though in my PR I decided it's best we enable by default and disable where required (as more resources would use them then not)

MarkLark86 (approved):

PR #2207, opened 2025-02-14
[SDESK-7487] Implement async `/planning-locks` endpoint 

eos87:
### Purpose
Convert the existing `/planning_locks` endpoint to async framework. 

### What has changed
- Converted the `/planning_locks` endpoint to use async `EndpointGroup`
- Updated the endpoint to use available async services (`EventsAsyncService`, `PlanningAsyncService`, and `AssignmentsAsyncService`) to query the list of locks.
- Implemented concurrent async database queries to fetch locks.
- Added necessary imports and refactored code to support async functionality.

Resolves: [SDESK-7487]

Thanks for checking!

[SDESK-7487]: https://sofab.atlassian.net/browse/SDESK-7487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/planning_locks.py
+++ b/server/planning/planning_locks.py
@@ -56,8 +56,6 @@ class PlanningLocksParams(BaseModel):
 
 @endpoint("planning_locks", methods=["GET"])
 async def get_planning_locks(_: None, params: PlanningLocksParams, _r: None):
-    print("*" * 100)
-    print(params.repos)
     resp = await _get_planning_module_locks(params.repos)
     return Response(resp)
 
--- a/server/planning/planning_locks.py
+++ b/server/planning/planning_locks.py
@@ -7,8 +7,8 @@ from pydantic import BaseModel, Field, field_validator
 
 from superdesk.core import json
 from superdesk import get_resource_service
-from superdesk.core.web import EndpointGroup
-from superdesk.core.types import Response, SearchRequest, SearchArgs, ESQuery
+from superdesk.core.web import endpoint
+from superdesk.core.types import Response, SearchRequest, ESQuery, ESBoolQuery
 
 from planning.events import EventsAsyncService
 from planning.planning import PlanningAsyncService
@@ -16,9 +16,6 @@ from planning.assignments import AssignmentsAsyncService
 from planning.core.service import BasePlanningAsyncService
 from planning.utils import get_first_related_event_id_for_planning
 from planning.search.queries.elastic import ElasticQuery, field_exists
-
-
-planning_locks_endpoints = EndpointGroup("/planning_locks", __name__, url_prefix="api")
 
 
 @unique
@@ -47,7 +44,7 @@ PROJECTED_FIELDS = [
 
 
 class PlanningLocksParams(BaseModel):
-    repos: Annotated[list[PlanningLockRepos], Field(default=DEFAULT_REPOS)]
+    repos: Annotated[list[PlanningLockRepos], Field(default_factory=lambda: DEFAULT_REPOS)]
 
     @field_validator("repos", mode="before")
     def parse_repos(cls, value: list[PlanningLockRepos] | str) -> list[PlanningLockRepos]:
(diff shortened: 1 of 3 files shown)


eos87:
This PR depends on https://github.com/superdesk/superdesk-core/pull/2832

server/planning/planning_locks.py
```diff
@@ -1,45 +1,38 @@
...
 from planning.utils import get_first_related_event_id_for_planning
 from planning.search.queries.elastic import ElasticQuery, field_exists
 
 
-class PlanningLocksResource(Resource):
-    resource_methods = ["GET"]
-    item_methods = []
-    endpoint_name = "planning_locks"
-    allow_unknown = True
+planning_locks_endpoints = EndpointGroup("/planning_locks", __name__, url_prefix="api")
```
MarkLark86: The `url_prefix` should not be here, as it's controlled via the app factory's [register_endpoint](https://github.com/superdesk/superdesk-core/blob/async/superdesk/factory/app.py#L302) function, and the [api_prefix](https://github.com/superdesk/eve/blob/sync-upstream/eve/flaskapp.py#L610) property

MarkLark86: Do we really need an EndpointGroup here anyway, considering we're registering a single endpoint?

server/planning/planning_locks.py
```diff
@@ -53,32 +46,49 @@ class PlanningLockRepos(Enum):
 ]
 
 
-bp = Blueprint("planning_locks", __name__)
+class PlanningLocksParams(BaseModel):
+    repos: Annotated[list[PlanningLockRepos], Field(default=DEFAULT_REPOS)]
```
MarkLark86: `default` here is mutable, therefor it would be safer to use default_factory to provide a copy of the `DEFAULT_REPOS` variable

server/planning/planning_locks.py
```diff
@@ -111,6 +121,54 @@ def _get_planning_module_locks():
...
+            - Projection fields from PROJECTED_FIELDS
+            - Page 1 with 1000 results per page
+    """
+    elastic_query = ESQuery()
+    elastic_query.query.must.append(field_exists("lock_session"))
+
+    return SearchRequest(
+        args=SearchArgs(
+            source=elastic_query.generate_query_dict(),
+            projections=json.dumps(PROJECTED_FIELDS),
```
MarkLark86: You can provide the [projection argument](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/types/search.py#L163) directly to the SearchRequest top level variable (the args.projections is there for compatibility with the way Eve did things).

And there you can simply use a list of strings, rather than converting to a json string.

MarkLark86: You can specifically use the [elastic attribute](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/types/search.py#L169) of the SearchRequest directly as well, and the elastic layer will use that. This makes it a bit cleaner and easier to read, i.e.

```python
req = SearchRequest(
    projections=PROJECTED_FIELDS,
    page=1,
    max_results=1000,
    elastic=ESQuery(
        query=ESBoolQuery(
            must=[field_exists("lock_session")]
        )
    ),
)
```

We could probably clean it up a bit, so you don't have to use ESQuery nor ESBoolQuery, but for now the above should do the trick

MarkLark86 (commented):
Just a couple small things, otherwise looks good

server/planning/planning_locks.py
eos87: > as it's controlled via the app factory's [register_endpoint](https://github.com/superdesk/superdesk-core/blob/async/superdesk/factory/app.py#L302) function, and the [api_prefix](https://github.com/superdesk/eve/blob/sync-upstream/eve/flaskapp.py#L610) property

right, that's what I thought but, for some reason I does not pick it up. Do you think I might have some missing configuration/setting? I've checked the URL_PREFIX env var and it's there so I'm not sure what could be.

eos87: > Do we really need an EndpointGroup here anyway, considering we're registering a single endpoint?

indeed, we do not need a group in this case. I'll change it.

server/planning/planning_locks.py
eos87: > the `args.projections` is there for compatibility with the way Eve did things

ahh good to know. I didn't know which one should be the preferred one to use. I will add a comment there because I think it is not really clear why we have both, or maybe I've missed that. Thanks!

MarkLark86 (approved):

BrianMwangi21 (approved):

PR #2209, opened 2025-02-14
check ci 250214

tomaskikutis:
## Front-end checklist


tomaskikutis:
Latest commit `915dbaa` in `feature/embedded-events-editor` has 1 e2e test file failing. For this PR, 7 e2e test files fail. Mostly the same 7 are failing on my other PR https://github.com/superdesk/superdesk-planning/pull/2203 - there, I've ran them locally - where [all but one passed](https://github.com/superdesk/superdesk-planning/pull/2203#issuecomment-2659963008). I suspect something on the server broke. Can you check it out @petrjasek ?

PR #2211, opened 2025-02-19
Add Agendas to list of available Content Filter fields

MarkLark86:
### Purpose
Prior to https://github.com/superdesk/superdesk-core/pull/2829 PR, Agendas was added into the list of available content filters in core. That PR removed it and replaced it with a signal for other repos/modules to register their own.

### What has changed
Add a signal listener to add Agendas as an available field for Content Filters (for publishing).
[Original code](https://github.com/superdesk/superdesk-core/blob/develop/apps/content_filters/filter_condition/filter_condition_parameters.py#L167)


eos87 (approved):
Looks good

PR #2212, opened 2025-02-19
[SDESK-7499] - Upgrade Event Spike & Unspike resources to async endpoints

BrianMwangi21:
### Purpose
This PR upgrades event spike & unspike resources to async endpoints

- Create endpoints for spike events and unspike events
- Applied auth rules to the new endpoints
- Updated the pure functions for spike and unspike events based off old implementation.

For the failing tests, this PR depends on this [PR](https://github.com/superdesk/superdesk-planning/pull/2211)

Solves: [SDESK-7499](https://sofab.atlassian.net/browse/SDESK-7499)

[SDESK-7499]: https://sofab.atlassian.net/browse/SDESK-7499


server/planning/events/events_spike.py
```diff
@@ -1,320 +1,335 @@
...
+                "id": event[ID_FIELD],
+                "etag": item["_etag"],
+                "state": event.get("revert_state", WORKFLOW_STATE.DRAFT),
+            }
+        )
+
+    updates["_unspiked_items"] = notifications
+
+
+async def process_spike_event(updates: dict[str, Any], original: dict[str, Any]) -> dict[str, Any] | None:
```
MarkLark86: We can safely assume this function will return a `dict`, and never `None`, this is because your're returning the result of a find_by_id and we have already done checks to make sure the Event exists.

Therefor the return type can simply be a dict, and you can maybe place an assert after `spiked_event = await events_service.find_by_id_raw(original[ID_FIELD])`

MarkLark86 (commented):
Overall it looks good, just need to update those endpoint URLs and add a name there

eos87 (approved):

MarkLark86 (approved):

PR #2214, opened 2025-02-24
Address PR-TODOs

thecalcc:
STT-31

## Front-end checklist

### DIFF
--- a/client/components/fields/editor/AssociatedEventItem.tsx
+++ b/client/components/fields/editor/AssociatedEventItem.tsx
@@ -1,10 +1,11 @@
+import React, {createRef} from 'react';
+import {IAuthoringReact} from 'superdesk-api';
 import {superdeskApi} from '../../../superdeskApi';
 import {authoringStorageEventItemHttp} from '../../../components/editor-standalone/authoring-storage-event-http';
 import {EventEditorStandalone} from '../../../components/editor-standalone/event-editor-standalone';
 import {RelatedEventListItem} from '../../../components/Events/EventMetadata/RelatedEventListItem';
-import React, {createRef} from 'react';
+import {CustomHeaderToggleBox} from 'superdesk-ui-framework/react/components/ToggleBox/CustomHeaderToggleBox';
 import {IconButton, ToggleBox} from 'superdesk-ui-framework/react';
-import {IAuthoringReact} from 'superdesk-api';
 
 interface IProps{
     removeEventItem(item: DeepPartial<IEventItem>): void;
@@ -14,7 +15,7 @@ interface IProps{
 }
 
 export class AssociatedEventItem extends React.PureComponent<IProps> {
-    public toggleBoxRef: React.RefObject<any>;
+    public toggleBoxRef: React.RefObject<CustomHeaderToggleBox>;
     public authoringRef: React.RefObject<IAuthoringReact<IEventItem>>;
 
     constructor(props) {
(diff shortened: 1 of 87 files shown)


client/components/fields/editor/AssociatedEventItem.tsx
```diff
@@ -61,18 +61,11 @@ export class AssociatedEventItem extends React.PureComponent<IProps> {
...
-                                return new Promise((resolve) => {
-                                    this.toggleBoxRef.current.toggle();
-
-                                    // PR-TODO: improve toggleBox so `toggle` method returns a promise
-                                    setTimeout(() => {
-                                        resolve();
-                                    }, 500);
-                                });
+                            if (this.toggleBoxRef.current.isOpen() === false) {
+                                return this.toggleBoxRef.current.toggle() as unknown as Promise<void>;
```
tomaskikutis: Could you type `toggleBoxRef` so you don't need to cast the result of `current.toggle` here?

Also it'd be good to first handle the positive condidion i.e. `if (condition) ... else ...` instead of
`if (!condition) ... else ...`

client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
```diff
@@ -141,18 +141,11 @@ export class RelatedPlanningItem extends React.PureComponent<IProps> {
...
-                            } else {
-                                return new Promise((resolve) => {
-                                    this.toggleBoxRef.current.toggle();
-
-                                    // PR-TODO: improve toggleBox so `toggle` method returns a promise
-                                    setTimeout(() => {
-                                        resolve();
-                                    }, 500);
-                                });
+                            if (this.toggleBoxRef.current.isOpen() === false) {
```
tomaskikutis: same comment as in other related item

client/components/fields/editor/AssociatedEventItem.tsx
thecalcc: togglebox is typed, but `toggle` returns a `Promise<boolean>`

thecalcc: and typing `makeVisible` would be useless rather as we don't and probably won't make use of the promise result in the authoring react internals

tomaskikutis: `toggleBoxRef: React.RefObject<any>;` doesn't look typed to me

thecalcc: My bad, it is typed in the other file for related planning item -`client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx`. I'll type it here as well

client/components/fields/editor/AssociatedEventItem.tsx
```diff
@@ -61,11 +62,11 @@ export class AssociatedEventItem extends React.PureComponent<IProps> {
...
                         authoringStorage={authoringStorageEventItemHttp}
                         makeVisible={() => {
-                            if (this.toggleBoxRef.current.isOpen() === false) {
-                                return this.toggleBoxRef.current.toggle() as unknown as Promise<void>;
+                            if (this.toggleBoxRef.current.isOpen()) {
+                                return Promise.resolve();
                             }
 
-                            return Promise.resolve();
+                            return this.toggleBoxRef.current.toggle() as unknown as Promise<void>;
```
tomaskikutis: if you're using the latest ui-framework `toggle` should return a promise already and you wouldn't need casting.

thecalcc: Yes, it returns a promise of boolean

tomaskikutis: I see. Do not do cast `as unknown` if you can avoid it, it's like ts-ignore and drops all typechecking. For this case if it returns a `Promise<boolean>` and you want `Promise<void>` you can always do `toggle().then(() => null)`

tomaskikutis (approved):

tomaskikutis:
@petrjasek could you check ~back-end  e2e~ nose test failure?

thecalcc:
> @petrjasek could you check back-end e2e failure?

I think he fixed it here - https://github.com/superdesk/superdesk-planning/pull/2217, let me know if I should port it

PR #2215, opened 2025-02-24
[SDESK-7498] - Upgrade Planning Spike & Unspike resources to async endpoints

BrianMwangi21:
### Purpose
This PR upgrades planning items spike & unspike resources to async endpoints

- Create endpoints for spike planning items and unspike planning items
- Applied auth rules to the new endpoints
- Updated the pure functions for spike and unspike planning items based off old implementation.

Solves: [SDESK-7498](https://sofab.atlassian.net/browse/SDESK-7498)

[SDESK-7498]: https://sofab.atlassian.net/browse/SDESK-7498


BrianMwangi21:
@MarkLark86, a question. 

In `planning/__init__.py` we have this : 

```
    app.on_updated_planning_spike += planning_history_service.on_spike
    app.on_updated_planning_unspike += planning_history_service.on_unspike
```

But from the old implementation and the overall code, I haven't seen anywhere they are called. So I'm wondering if signals for planning_spike and planning_unspike are needed too ? 

Thanks

MarkLark86:
> @MarkLark86, a question. 
> 
> In `planning/__init__.py` we have this : 
> 
> ```
>     app.on_updated_planning_spike += planning_history_service.on_spike
>     app.on_updated_planning_unspike += planning_history_service.on_unspike
> ```
> 
> But from the old implementation and the overall code, I haven't seen anywhere they are called. So I'm wondering if signals for planning_spike and planning_unspike are needed too ? 
> 
> Thanks

Yes they are. These signals were raised by the Eve library, and by our deprecated base resource & service classes (look at the code in Superdesk core)

server/planning/planning/planning_spike_async.py
```diff
@@ -0,0 +1,192 @@
...
+            coverage["workflow_status"] = WORKFLOW_STATE.DRAFT
+            coverage["assigned_to"] = {}
+
+    updates["coverages"] = coverages
+
+    # Mark item as unlocked directly in order to avoid more queries and notifications
+    # coming from lockservice.
+    remove_lock_information(updates)
+
+    remove_autosave_on_spike(original)
```
MarkLark86, line 125: I looked up [this function](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/common.py#L391), and it's used only by Event and Planning spike functions.
We should now be able to update this function so it is async and uses the async autosave resources

server/planning/planning/planning_spike_async.py
```diff
@@ -0,0 +1,192 @@
...
+
+    id = original[ID_FIELD]
+    await planning_service.update(id, updates)
```
MarkLark86: We should probably use `planning_service.system_update(id, updates)` here, so that the base services' signals are not called (keeping the same functionality as the original implementation)

server/planning/planning/planning_spike_async.py
```diff
@@ -0,0 +1,192 @@
...
+
+    updates["coverages"] = coverages
+
+    # Mark item as unlocked directly in order to avoid more queries and notifications
+    # coming from lockservice.
+    remove_lock_information(updates)
+
+    remove_autosave_on_spike(original)
+
+    id = original[ID_FIELD]
```
MarkLark86: `id` is a python built-in function, please use a meaning name here

server/planning/planning/planning_spike_async.py
```diff
@@ -0,0 +1,192 @@
...
+
+    notify_user_on_failed_assignment_deletes = True
+    first_event_id = get_first_related_event_id_for_planning(original, "primary")
+
+    if first_event_id:
+        event = await events_service.find_by_id_raw(first_event_id)
+        notify_user_on_failed_assignment_deletes = not event or event.get("state") != WORKFLOW_STATE.SPIKED
+
+    # TODO-ASNYC: Convert this to async call when function is added to PlanningAsyncService
+    get_resource_service("planning").delete_assignments_for_coverages(
```
MarkLark86: This looks like a small enough function, would you be able to re-create this function in async in the `planning_utils` file in this PR and use it here?

server/planning/planning/planning_spike_async.py
MarkLark86: This is because the original used `self.backend.update`, which I believe bypasses the on_update/on_updated of the base class

server/planning/planning/planning_spike_async.py
```diff
@@ -0,0 +1,192 @@
...
+        if event and event.get("state") == WORKFLOW_STATE.SPIKED:
+            raise SuperdeskApiError.badRequestError(message="Unspike failed. Associated event is spiked.")
+
+    updates[ITEM_STATE] = original.get("revert_state", WORKFLOW_STATE.DRAFT)
+    updates["revert_state"] = None
+    updates[ITEM_EXPIRY] = None
+    remove_lock_information(updates)
+
+    id = original[ID_FIELD]
+    await planning_service.update(id, updates)
```
MarkLark86: Same here, use system_update to update the item

MarkLark86 (commented):
Looks good, just a couple small improvements

server/planning/planning/planning_spike_async.py
BrianMwangi21: This is true.

However, this function is also used by `PlanningSpikeService` which is used by `EventsPostService` which is yet to be converted to async. 

Hence why I removed the `PlanningUnspikeService` only cause it wasn't used anywhere else. But retained the planning spike as an internal resource for this. 

I've added more context to the todo. I think it will be safe to convert it when `EventPostService` is changed to async.

MarkLark86: Fair enough :+1:

server/planning/planning/planning_spike_async.py
```diff
@@ -0,0 +1,187 @@
...
+            etag=updates.get("_etag"),
+            event_ids=get_related_event_ids_for_planning(original),
+            recurrence_id=original.get("recurrence_id") or None,
+            type=original.get("type"),
+        )
+
+
+async def post_planning_item_spike_actions(updates: dict[str, Any], original: dict[str, Any]):
+    post_update_planning_item_actions(updates, original)
+    events_service = EventResourceModel.get_service()
```
eos87: Does using the service class here produce circular references? Just wondering if it would be better to use the service class if possible

server/planning/planning/planning_spike_async.py
```diff
@@ -0,0 +1,187 @@
...
+
+    if first_event_id:
+        event = await events_service.find_by_id_raw(first_event_id)
+        notify_user_on_failed_assignment_deletes = not event or event.get("state") != WORKFLOW_STATE.SPIKED
+
+    await delete_assignments_for_coverages(assignments_to_delete, notify_user_on_failed_assignment_deletes)
+
+
+async def notify_draft_coverage_on_spike(coverage: dict[str, Any]):
+    assignment_service = AssignmentResourceModel.get_service()
```
eos87: Same here

eos87 (approved):
It looks good to me. I just left a comment regarding the usage of `Model.get_service()`. Wondering if we could use the service class directly (if possible) for more clarity.

MarkLark86:
> It looks good to me. I just left a comment regarding the usage of `Model.get_service()`. Wondering if we could use the service class directly (if possible) for more clarity.

Hi @eos87 & @BrianMwangi21, Not sure what is meant by `for more clarity`?. The `get_service` returns an instance of AsyncResourceService.

Previously our resource services were full of code that had nothing to do with the management of the resource storage, so I would like to keep these resource services mean and clean, focusing just on the CRUD and validation functionality. Anything else can be a util that is imported.

As for why I recommend to use the `Model.get_service()`, is because it drastically reduces the likelihood of a cyclic import. I saw this all too often in the Newshub code. 

If you can, please use the Model to get a handle to it's data management service if possible.

eos87:
> Not sure what is meant by `for more clarity`?

`PlanningAsyncService().find` vs `Planning.get_service().find`. I’d rather see services stand alone than hide behind models. But for the sake of "clarity" (readability) it is not a big deal honestly, it was just a quick thought.

> Previously our resource services were full of code that had nothing to do with the management of the resource storage, so I would like to keep these resource services mean and clean, focusing just on the CRUD and validation functionality. Anything else can be a util that is imported.

I 100% agree with this :) 

> As for why I recommend to use the Model.get_service(), is because it drastically reduces the likelihood of a cyclic import. I saw this all too often in the Newshub code.
If you can, please use the Model to get a handle to it's data management service if possible.

I have to disagree here. A few thoughts: 
- Model classes are representation of a resource in our database collection/table (at least how I see it) and they should not know how to fetch a service, plus they should not be paired 1:1 to a single service class as we are locking ourselves in a rigid structure and at the same time it is counterproductive with our new architecture/design.
- Circular references **_should not_** happen in the first place. They often suggest a dependency tangle. To me they are like a warning about something that's becoming too tight in our code. If you have a circular reference in the code you're writing, you should think about it twice and try as much as possible to avoid such circular reference, considering refactoring/improving the code.   
- I understand the existence of the `get_service` method, as there are times, when it's almost impossible to avoid the circular reference given the complexity of the project and the tight timeline but, it should only be used as a last resort instead of using it more often.

PR #2216, opened 2025-02-26
Fix qa reported bugs

thecalcc:
STT-31
STT-153
STT-157
STT-156

## Front-end checklist


thecalcc:
TODO:
- [ ] autosave should work for temporary embedded events. 
- [ ] add all fields for events
- [ ] handle input type fields config - multiline, editor3, etc. 
- [x] change description of tooltips and icons for unlinking an embedded item
- [x] fix empty labels
- [x] fix saving & linking of events

Note: Look into `_temporary` prop of planning item. Should be able to store temp events there.

client/components/Main/ItemEditor/ItemManager.ts
```diff
@@ -661,12 +664,18 @@ export class ItemManager {
...
 
-        const promise: Promise<any> = handleEmbeddedItems(this.props.editorType, 'SAVE', this.props.itemType)
-            .then(() =>
-                !updateStates ? Promise.resolve({}) : this.setState({submitting: true, submitFailed: false}),
+        /**
+         * Calls internal save of each embedded item - the one of the storage adapter, then returns the saved items.
+         */
+        const promise = handleEmbeddedItems(this.props.editorType, 'SAVE', this.props.itemType)
+            .then((res) =>
+                Promise.all([
```
tomaskikutis, line 671: `Promise.all` accepts an array of promises as arguments. `res` is not a promise, neither is `{}` at line 674.

client/components/editor-standalone/save-handling.ts
```diff
@@ -37,24 +41,28 @@ const getEmbeddedItemsExposed = (
...
  */
-export const handleEmbeddedItems = async(
+export const handleEmbeddedItems = async<T extends IEventItem | IPlanningItem>(
     editorType: EDITOR_TYPE,
     action: IEmbeddedPlanningsActionType,
     itemType: ItemType,
-): Promise<void> => {
-    for (const exposed of getEmbeddedItemsExposed(editorType, itemType)) {
+): Promise<Array<T>> => {
+    const updatedItems: Array<T> = [];
```
tomaskikutis, line 49: what's the purpose of adding this?

client/components/editor-standalone/field-definitions/index.ts
```diff
@@ -18,9 +18,14 @@ import {getPriorityField} from './priority-field';
 import {getLocationsField} from './locations-field';
 import {getLinksField} from './link-field';
 import {getContactsField} from './contacts';
+import {getOccurStatusField} from './occurence-status';
+import {getLanguageField} from './lanugage';
```
tomaskikutis: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker :upside_down_face: 

image

client/components/Main/ItemEditor/ItemManager.ts
```diff
@@ -32,6 +33,8 @@ import {
     handleEmbeddedItems,
     IEmbeddedPlanningsActionType,
 } from '../../../components/editor-standalone/save-handling';
+import eventsApi from '../../../actions/events/api';
+import {initialValues} from 'selectors/forms';
```
tomaskikutis: unused imports

client/components/Main/ItemEditor/ItemManager.ts
```diff
@@ -693,14 +702,23 @@ export class ItemManager {
...
-                    });
+                        updates.associated_plannings?.forEach((planningItem) => {
+                            if (planningUpdateMethods[planningItem._id] != null) {
+                                planningItem.update_method = planningUpdateMethods[planningItem._id];
+                            }
+                        });
                 }
             }
 
+            if (updates.type === 'planning' && (updates.related_events ?? []).length > 0) {
```
tomaskikutis: What are you doing here that wasn't working with the previous version of code?

client/components/editor-standalone/field-definitions/calendars.ts
```diff
@@ -0,0 +1,52 @@
...
+    return {
+        fieldId: 'calendars',
+        getField: ({required, id}) => {
+            const options = vocabularyFromStore.map(
+                (option) => ({
+                    id: option.qcode,
+                    label: getVocabularyItemFieldTranslated(
+                        option,
+                        'label',
+                        'en',
```
tomaskikutis: language shouldn't be hardcoded

client/components/editor-standalone/field-definitions/calendars.ts
```diff
@@ -0,0 +1,52 @@
...
+                        'name'
```
tomaskikutis: use `nameof` to preserve type safety for interface members

client/components/editor-standalone/field-definitions/all-day.ts
```diff
@@ -0,0 +1,56 @@
...
+                        .endOf('day');
+                } else {
+                    // If allDay is disabled, then set the new dates to
+                    // the initial values since last save and time to empty
+                    newEnd = moment(dates?.end ?? newStart)
+                        .hour(0)
+                        .minute(1);
+                }
+
+                // TODO: Figure out support for _startTime _endTime, _time_to_be_confirmed fields
```
tomaskikutis: should this be PR-TODO?

client/components/editor-standalone/field-definitions/lanugage.ts
```diff
@@ -0,0 +1,43 @@
...
+    return {
+        fieldId: 'language',
+        getField: ({required, id}) => {
+            const options = vocabularyFromStore.map(
+                (option) => ({
+                    id: option.qcode,
+                    label: getVocabularyItemFieldTranslated(
+                        option,
+                        'label',
+                        'en',
```
tomaskikutis: language shouldn't be hardcoded

client/components/editor-standalone/field-definitions/occurence-status.ts
```diff
@@ -0,0 +1,53 @@
...
+    fieldId: 'occur_status',
+    getField: ({required, id}) => {
+        const vocabularyFromStore = eventOccurStatuses(planningApi.redux.store.getState());
+        const options = vocabularyFromStore.map(
+            (option) => ({
+                id: option.qcode,
+                label: getVocabularyItemFieldTranslated(
+                    option,
+                    'label',
+                    'en',
```
tomaskikutis: language shouldn't be hardcoded

client/components/editor-standalone/field-definitions/lanugage.ts
```diff
@@ -0,0 +1,43 @@
+import {IAuthoringFieldV2, IDropdownConfigManualSource} from 'superdesk-api';
+import {planningApi, superdeskApi} from '../../../superdeskApi';
+import {getLanguages} from '../../../selectors/vocabs';
+import {getVocabularyItemFieldTranslated} from '../../../utils/vocabularies';
+
+export const getLanguageField = () => {
+    const vocabularyFromStore = getLanguages(planningApi.redux.store.getState());
```
tomaskikutis: `languages` would be a better name.

client/components/editor-standalone/save-handling.ts
thecalcc: I use the result of `handleEmbeddedItems` in `ItemManager.tsx` to update `related_events`.

client/components/Main/ItemEditor/ItemManager.ts
thecalcc: I'm updating `related_events`

client/components/editor-standalone/save-handling.ts
tomaskikutis: Could you respond to everything in one review?

client/components/AttachmentsInputStandalone.tsx
```diff
@@ -33,11 +33,11 @@ export class AttachmentsInputStandalone extends React.PureComponent<IProps, ISta
             <Spacer v gap="16" noWrap>
                 <WithLiveResources
                     resources={[
-                        {resource: 'planning_files', ids: this.props.value},
+                        {resource: 'planning_files', ids: this.props.value ?? []},
                     ]}
                 >
                     {(res) => {
-                        const files: Array<IFile> = res[0]._items;
+                        const files: Array<IFile> = res?.[0]?._items ?? [];
```
tomaskikutis: after https://github.com/superdesk/superdesk-client-core/pull/4762 you can revert this line to how it was

client/api/editor/item_planning.ts
```diff
@@ -81,6 +82,49 @@ export function getPlanningInstance(type: EDITOR_TYPE): IEditorAPI['item']['plan
...
+                    }
+                });
+        };
+
+        // On saving of a temporary event from the embedded form itself,
+        // we must also remove it from _unsaved_related_events
+        // otherwise trying to save the same item twice would happen
+        if (isTemporaryId(original._id) && !isTemporaryId(updates._id)) {
+            editor.form.changeField(
+                '_unsaved_related_events',
```
tomaskikutis: use `nameof`

client/components/editor-standalone/authoring-storage-in-memory.ts
```diff
@@ -1,17 +1,31 @@
...
+import {autosave} from '../../api/autosave';
 
-export function getAuthoringStorageInMemory<T>(
+export function getAuthoringStorageInMemory<T extends {_id: string}>(
     profile: 'event' | 'planning',
     item: T,
     onSave: (current: T, original: T) => Promise<T>,
 ): IAuthoringStorage<T> {
+    /**
+     * Timeout for 500 seconds to let newly added embedded events finish autosaving.
```
tomaskikutis: Timing can't be relied upon. We should do a better solution. Could you elaborate what issue you're trying to address here?

Why did you drop my usage of spread operator to handle validation?

client/components/editor-standalone/authoring-storage-in-memory.ts
```diff
@@ -1,17 +1,31 @@
...
+     */
     class NoAutoSave implements IAuthoringAutoSave<T> {
         get(id: string) {
-            // return a different reference so authoring-react sees it as having unsaved changes.
-            // otherwise, it will consider all saved and won't attempt to save - which won't trigger validation.
-            return Promise.resolve({...item});
+            return new Promise<T>((resolve) => {
+                setTimeout(() => {
+                    autosave
+                        .getById(profile === 'event' ? 'event' : 'planning', id)
```
tomaskikutis: > profile === 'event' ? 'event' : 'planning'

It's a trap to do it like that. If more items are added to the union type the code will fail silently. And we want it to fail loud and clear :D Use if/else if/assert never

client/interfaces.ts
```diff
@@ -750,6 +750,7 @@ export interface IPlanningItem extends IBaseRestApiResponse {
     // added by client - should be dropped before sending to server
     event?: IEventItem;
 
+    _unsaved_related_events?: Array<IPlanningRelatedEventLink>;
```
tomaskikutis, line 757: add a comment on how it is mean to be used. sent to autosave, but not save?

server/planning/planning/planning_schema.py
```diff
@@ -218,6 +218,7 @@
         "mapping": not_analyzed,
         "nullable": True,
     },
+    "_unsaved_related_events": {"type": "list", "nullable": True},
```
tomaskikutis: can this be added to autosave only?

client/components/editor-standalone/field-definitions/recurring-rules.ts
```diff
@@ -0,0 +1,52 @@
...
+            };
+
+            return field;
+        },
+        storageAdapterEvent: {
+            retrieveStoredValue: (item) => {
+                const clonedValue = cloneDeep(item.dates.recurring_rule);
+
+                if (clonedValue?.until != null) {
+                    set(clonedValue, nameof<typeof clonedValue>('until'), moment(clonedValue.until));
```
tomaskikutis: since value update path is static rather than dynamic - no need to use `lodash.set` here. It'd be cleaner to use the usual immutable updating via spreading.

tomaskikutis:
can you resolve the conflicts?

client/components/editor-standalone/field-definitions/calendars.ts
```diff
@@ -0,0 +1,55 @@
+import {ICalendar} from 'interfaces';
+import {IAuthoringFieldV2, IDropdownConfigManualSource} from 'superdesk-api';
+import {planningApi, superdeskApi} from '../../../superdeskApi';
+import {getVocabularyItemFieldTranslated} from '../../../utils/vocabularies';
+import {calendars} from '../../../selectors/events';
+import {nameof} from 'core/helpers/typescript-helpers';
```
tomaskikutis: don't import from core directly. It's available in planningApi

client/components/editor-standalone/field-definitions/calendars.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    return {
+        fieldId: 'calendars',
+        getField: ({required, id, language}) => {
+            const options = vocabularyFromStore.map(
+                (option) => ({
+                    id: option.qcode,
+                    label: getVocabularyItemFieldTranslated(
+                        option,
+                        'label',
```
tomaskikutis: use nameof in all places where getVocabularyItemFieldTranslated is used

client/components/editor-standalone/base-editor-standalone.tsx
```diff
@@ -44,16 +44,31 @@ function validate<T extends IPlanningItem | IEventItem>(
...
-            messages: messages,
-            diff: latestItem,
-        });
+    const profileFields = getPlanningProfileFields({
+        profile: entityType,
+        embeddedOnly: true,
+    });
+
+    for (const {fieldId} of profileFields) {
+        if (fieldId === 'recurring_rules') {
```
tomaskikutis, line 57: would it not better to put that condition inside formProfile function? How was it working before? Or it wasn't?

client/components/editor-standalone/profile-fields.ts
```diff
@@ -22,6 +31,17 @@ const unimplementedFields = new Set<string>([
     'associated_event',
 ]);
 
+export const TEXT_FIELDS_MULTIPLE_TYPES = new Set<string>([
```
tomaskikutis: use nameof. What does"multiple types" mean?

client/extension_bridge.ts
```diff
@@ -55,10 +68,14 @@ interface IExtensionBridge {
...
             EditorFieldLocation: React.ComponentType<IEditorFieldLocationProps>;
             EditorFieldCoverages: React.ComponentType<IPropsEditorFieldCoverages>;
+            EditorFieldEventRecurringRules: React.ComponentType<IEditorFieldEventRecurringRulesProps>;
+            EditorFieldEventSchedule: React.ComponentType<IEventScheduleFieldProps>;
         },
     }
     ui: {
         utils: {
+            planning_event_link_method: 'one_primary' | 'many_secondary' | 'one_primary_many_secondary';
+            getItemProfile: (type: 'planning' | 'event') => IPlanningContentProfile;
```
tomaskikutis, line 78: if it returns `IPlanningContentProfile`, how come is it possible to pass `type: 'event'`?

thecalcc: types for profiles are messed up in a lot of places. I chose not to spend time refactoring them, so I followed this pattern. There's no generic support for `IEditorFieldProps` and other interfaces used for both `event` & `planning`, and in the case of `IEditorFieldProps` there's a prop `profile: IPlanningContentProfile`, so a type cast would be needed if I add another return type to this function. And also - there's really no interface for event content profiles, only for event form profile.

tomaskikutis: ok, I resolve the comment then

thecalcc: I'd typecast just to follow the box pattern, but there's no other interface for event content profiles. Another option is to rename it to some generic name that covers event & planning item both.

client/components/editor-standalone/base-editor-standalone.tsx
thecalcc: No, because in the other places we use this `formProfile` function we pass the whole date, and then the validator runs its logic. If I do that I'll have to refactor existing usages

client/components/editor-standalone/profile-fields.ts
thecalcc: means that this field has the option for its input type to be configured. I've changed the name and added a comment to clarify it

client/components/editor-standalone/field-definitions/lanugage.ts
thecalcc: resolved

tomaskikutis:
I've checked it out - all good. There's one more comment you didn't address that I left unresolved https://github.com/superdesk/superdesk-planning/pull/2216#discussion_r1981695333

tomaskikutis (approved):

e2e/cypress/e2e/events/event_action_create_planning.cy.ts
```diff
@@ -30,15 +30,18 @@ describe('Planning.Events: create planning action', () => {
...
             definition_short: 'Desc.',
             occur_status: {
                 name: 'Planned, occurs certainly',
                 label: 'Confirmed',
                 qcode: 'eocstat:eos5',
             },
             dates: {
-                start: start.utc().format("YYYY-MM-DDTHH:mm:ss+0000"),
-                end: start.clone().utc().add(1, 'h').format('YYYY-MM-DDTHH:mm:ss+0000'),
+                start: start.utc().format('YYYY-MM-DDTHH:mm:ss+0000'),
```
tomaskikutis, line 41: why did you change dates?

thecalcc: I haven't changed the values, this is just auto formatting

tomaskikutis (approved):

PR #2217, opened 2025-02-27
Feat coverage profiles

thecalcc:
STT-111

## Front-end checklist


client/interfaces.ts
```diff
@@ -2178,7 +2189,9 @@ export interface IEditorAPI {
             };
             removeEventItem(item: DeepPartial<IEventItem>): void;
             getRelatedEventsDomRef(eventId: IEventItem['_id']): React.RefObject<any>;
-            getCoverageFields(): ISearchProfile;
+            getCoverageFields(
```
tomaskikutis, line 2192: To get coverage fields - shouldn't you pass `profile` as an argument? How else will it be determined which fields to return?

client/interfaces.ts
```diff
@@ -2328,7 +2340,17 @@ export interface IPlanningAPI {
         };
         getDefaultLanguage(profile: IPlanningContentProfile): IVocabularyItem['qcode'];
         getDefaultValues(profile: IPlanningContentProfile): DeepPartial<IEventOrPlanningItem | IPlanningCoverageItem>;
-        patch(original: IPlanningContentProfile, updates: IPlanningContentProfile): Promise<IPlanningContentProfile>;
+        patch(
+            original: IPlanningContentProfile,
+            updates: IPlanningContentProfile,
+        ): Promise<IPlanningContentProfile>;
+        coverages: {
```
tomaskikutis, line 2347: what about instead of nesting this in `contentProfiles` - create a separate `coverageProfiles` interface? I'm not insisting on one way or another, rather thinking what would be a better structure.

client/components/ContentProfiles/CoverageProfileModal.tsx
```diff
@@ -0,0 +1,289 @@
...
+    profile: Partial<ICoverageContentProfile> & IEditorProfile;
+    selectedType: ICoverageType;
+    allProfiles: Array<ICoverageContentProfile>;
+}
+
+interface IProps {
+    closeModal(): void;
+}
+
+const coverageType: Array<ICoverageType> = [
```
tomaskikutis: Would be good to use typescript to enforce all of them to be listed. Here's an example:
```
const allCoverageTypesObject: {[key in ICoverageType]: 1} = {
    text: 1,
    picture: 1,
    video: 1,
    audio: 1,
    infographics: 1,
    liveBlog: 1,
    liveVideo: 1,
};

const allCoverageTypes: Array<ICoverageType> = Object.keys(allCoverageTypesObject);
```

client/components/ContentProfiles/CoverageProfileModal.tsx
```diff
@@ -0,0 +1,289 @@
...
+                        gettext('Failed to save the profile!')
+                    )
+                );
+
+                this.setState({saving: false});
+            });
+    }
+
+    updateField(item: IProfileFieldEntry) {
+        const profileCloned = cloneDeep(this.state.profile);
```
tomaskikutis: React decides whether to re-render based on referential equality. If you use `cloneDeep` you get brand new references which will be different even if data is the same. In short - you disable the most important rendering optimization - that is not to re-render if references haven't changed.

I did suggest to use it previously in combination with `lodash.set` , but that's because due to having a dynamic update path we were forced to use `lodash.set`.

If you're not working with a dynamic update path, do not use a clone-and-mutate approach. The idea itself is not bad, immer uses it, but they keep references the same to things that didn't change.

client/components/ContentProfiles/CoverageProfileModal.tsx
```diff
@@ -0,0 +1,289 @@
...
+                        error,
+                        gettext('Failed to save the profile!')
+                    )
+                );
+
+                this.setState({saving: false});
+            });
+    }
+
+    updateField(item: IProfileFieldEntry) {
```
tomaskikutis, line 146: What about dropping `updateField` and only keeping `updateFields`. If you want to update only one, you pass an array with one item?

client/components/ContentProfiles/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    requiredFields.forEach((fields) => {
+        const result = fields.some(
+            (field) => isProfileFieldEnabled(profile, field, includeGroupCheck)
+        );
+
+        if (!result) {
+            valid = false;
+            if (fields.length === 1) {
+                notify.error(gettext('"{{field}}" field is required by the system', {
```
tomaskikutis: choose such placeholder name so it doesn't repeat things already present in translation string.

How about `"{{fieldName}}" field is required`
or `field "{{name}}" is required`

client/components/ContentProfiles/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+                    field: getFieldNameTranslated(fields[0]).toUpperCase()
+                }));
+            } else {
+                notify.error(gettext('At least one "{{fields}}" fields are required by the system', {
```
tomaskikutis: `field {{name}} is required by the system` would be better

client/selectors/coverageProfiles.ts
```diff
@@ -0,0 +1,8 @@
+import {ICoverageContentProfile} from 'interfaces';
+import {get} from 'lodash';
+
+export const coverageProfiles: (state: any) => Array<ICoverageContentProfile>
+    = (state) => get(state, 'coverageProfiles.profiles', []);
```
tomaskikutis: do not use lodash.get

tomaskikutis (commented):
While the task requires being able to configure a distinct content profile for each content type e.g. text, video, audio - the implementation itself at the database level or at data-structure level - should not be coupled to content type. If a new requirement came in tomorrow - that we want multiple profiles for a single content type - that shouldn't require us to do changes to database structure or data structures. It should be enough to adjust the UI in the modal.

I remember discussing this - that we'd retrieve profiles by ID - not content type with @petrjasek. Not sure if you @thecalcc wasn't participating in the discussion or you guys changed the approach.

client/interfaces.ts
thecalcc: no, this returns the old configured coverage profile, which we use as a fallback

client/interfaces.ts
thecalcc: I can do it if you think it's better, imo we don't need it now, maybe later, but when that comes we'll restructure since easily since it's got 2 functions

client/components/ContentProfiles/utils.ts
thecalcc: this code was actually kind of bad. I began reworking it, but it was tied to other bad code, then I realized that some stuff related to `embeddedProfile` was no longer used now that I decoupled the coverage profiles. So I now have reworked it to work without this case, while achieving the same functionality in a better and more obvious way for the user - by notifying them about each required field

thecalcc:
> While the task requires being able to configure a distinct content profile for each content type e.g. text, video, audio - the implementation itself at the database level or at data-structure level - should not be coupled to content type. If a new requirement came in tomorrow - that we want multiple profiles for a single content type - that shouldn't require us to do changes to database structure or data structures. It should be enough to adjust the UI in the modal.
> 
> I remember discussing this - that we'd retrieve profiles by ID - not content type with @petrjasek. Not sure if you @thecalcc wasn't participating in the discussion or you guys changed the approach.

It's not actually coupled to the `content_type` I think. Now we use that as just a property right @petrjasek ? We can always make the creation of new coverage types also dynamic which would pass on the backend because in the schema it's defined like this: `"content_type": {"type": "string", "required": True},`. Then we can adjust frontend according to the requirements - instead of working with a static type we'll use ids and list all coverage profiles by fetching them all. Just like we do with article profiles probably. For now how I understand it - we should list profiles for all coverage types, regardless whether they were actually created by the user or not - if not - we render the old profile. If it was created after this update - we show the new data.

thecalcc:
To add to my comment - now I think we should keep things as they are in the UI. Otherwise we need to either make some workaround (mock some static profiles so users can access the in the profile config) to prepare for a feature that may never come. And when it comes we can add more UI - a plus button to add coverage profiles dynamically, and endless `content_types`. And a text input for `content_type` field itself and then fetch everything by id.

petrjasek:
> While the task requires being able to configure a distinct content profile for each content type e.g. text, video, audio - the implementation itself at the database level or at data-structure level - should not be coupled to content type. If a new requirement came in tomorrow - that we want multiple profiles for a single content type - that shouldn't require us to do changes to database structure or data structures. It should be enough to adjust the UI in the modal.
> 
> I remember discussing this - that we'd retrieve profiles by ID - not content type with @petrjasek. Not sure if you @thecalcc wasn't participating in the discussion or you guys changed the approach.

we're ready for multiple on the backend, nothing changed in that regard

tomaskikutis:
[image](https://github.com/user-attachments/assets/8d9b2270-785c-41de-8e38-371f5be2b3b5)


@dzonidoo could you improve this PR to make button content left-aligned?

See `client/components/ContentProfiles/CoverageProfileModal.tsx`.
You'll need to add that capability to ui-framework buttons. Use RTL friendly names for the prop
`textAlign: 'start' | 'end' | 'center' // defaults to 'center'`

PR #2219, opened 2025-03-04
[SDESK-7516] - Upgrade Locations resource & service to async

BrianMwangi21:
### Purpose
This PR creates new Location resources, services and REST APIs:

- Implementing a Locations resource model
- Implementing async resource service for Locations
- Implementing resource config for the new resources and register them in the planning module
- Disable the API in old Locations resources as it is still used by JSON event ingest parser

Solves: [SDESK-7516](https://sofab.atlassian.net/browse/SDESK-7516)

[SDESK-7516]: https://sofab.atlassian.net/browse/SDESK-7516

### DIFF
--- a/server/planning/locations/locations_service_async.py
+++ b/server/planning/locations/locations_service_async.py
@@ -1,7 +1,6 @@
 from typing import Any
 
 from planning.core.service import BasePlanningAsyncService
-from planning.common import format_address
 from planning.events.events_service import EventsAsyncService
 from planning.types import LocationResourceModel
 
@@ -18,27 +17,17 @@ class LocationsAsyncService(BasePlanningAsyncService[LocationResourceModel]):
             if not doc.guid:
                 doc.guid = generate_guid(type=GUID_NEWSML)
 
-    # TODO-ASYNC: Confirm on `on_fetched` function from base AsyncResourceService
-    def on_fetched(self, docs):
-        for doc in docs["_items"]:
-            self._enhance_item(doc)
-
-    # TODO-ASYNC: Confirm on `on_fetched_item` function from base AsyncResourceService
-    def on_fetched_item(self, doc):
-        self._enhance_item(doc)
-
-    def _enhance_item(self, doc):
-        format_address(doc)
-
     async def delete_many(self, lookup: dict[str, Any]) -> list[str]:
-        """If the document to be deleted is reference in an event then flag it as inactive otherwise just delete it."""
+        """
+        If the document to be deleted is reference in an event then flag it as inactive otherwise just delete it.
+        """
         if lookup:
             location = await self.find_by_id(lookup.get(ID_FIELD, ""))
             if location:
                 search_request = SearchRequest(where={"location.qcode": str(location.guid)})
                 events = await EventsAsyncService().find(search_request)
                 if await events.count() > 0:
-                    # Patch the unique name in case the location get recreated
+                    # Update the unique name in case the location get recreated
                     await self.update(
                         location.id,
                         {
--- a/server/planning/locations/locations_service_async.py
+++ b/server/planning/locations/locations_service_async.py
@@ -4,19 +4,10 @@ from planning.core.service import BasePlanningAsyncService
 from planning.events.events_service import EventsAsyncService
 from planning.types import LocationResourceModel
 
-from superdesk.core.types import SearchRequest
-from superdesk.core.utils import generate_guid, GUID_NEWSML
 from superdesk.resource_fields import ID_FIELD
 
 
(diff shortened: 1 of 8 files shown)


server/planning/locations/locations_service_async.py
```diff
@@ -0,0 +1,40 @@
...
+
+    async def delete_many(self, lookup: dict[str, Any]) -> list[str]:
+        """
+        If the document to be deleted is referenced in an event, flag it as inactive otherwise just delete it.
+        """
+        if lookup:
+            location = await self.find_by_id(lookup[ID_FIELD])
+            if location:
+                search_request = SearchRequest(where={"location.qcode": str(location.guid)})
+                events = await EventsAsyncService().find(search_request)
```
MarkLark86: There is a count function available on the service that can be used here instead
```python
await EventsAsyncService().count({"location.qcode": location.guid)})
```

server/planning/locations/locations_service_async.py
```diff
@@ -0,0 +1,40 @@
...
+from superdesk.resource_fields import ID_FIELD
+
+
+class LocationsAsyncService(BasePlanningAsyncService[LocationResourceModel]):
+    async def on_create(self, docs: list[LocationResourceModel]) -> None:
+        await super().on_create(docs)
+
+        for doc in docs:
+            if not doc.guid:
+                doc.guid = generate_guid(type=GUID_NEWSML)
```
MarkLark86: Seems like a location should ALWAYS have a `guid` value. In this case I would use Pydantic to do this for us, and we can remove this `on_create` entirely from this service.

On the model you would replace
```python
guid: str | None = None
```
with
```python
guid: Annotated[str, Field(default_factory=lambda: generate_guid(type=GUID_NEWSML))]
```

And this way Pydantic will ensure that the `guid` field will always have a value, and never be `None`

server/planning/locations/module.py
```diff
@@ -0,0 +1,21 @@
...
+
+locations_resource_config: ResourceConfig = ResourceConfig(
+    name="locations",
+    data_class=LocationResourceModel,
+    service=LocationsAsyncService,
+    elastic=ElasticResourceConfig(),
+    rest_endpoints=RestEndpointConfig(
+        item_methods=["GET", "PATCH", "PUT", "DELETE"],
+        resource_methods=["GET", "POST"],
+        endpoints_class=LocationsRestEndpoints,
```
MarkLark86, line 20: Missing auth config, using https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/auth/privilege_rules.py#L44

server/planning/types/locations.py
```diff
@@ -0,0 +1,78 @@
...
+    country: str | None = None
+    locality: str | None = None  # Aka city
+    area: str | None = None
+    external: dict | None = None
+    boundingbox: list | None = None
+    address_type: str | None = Field(alias="type", default=None)
+
+
+class LocationResourceModel(BasePlanningModel):
+    guid: str | None = None
```
MarkLark86: As this resource uses Elasticsearch, we must ensure that the elastic mapping is correct.
To this end, anything in the old schema that has `"mapping": not_analyzed` needs to be replaced with [fields.Keyword](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/fields.py#L126)

MarkLark86: Also any field that has `"unique": True` in the old schema requires the [validate_iunique_value_async](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/validators.py#L230) annotated for that field

MarkLark86 (commented):
:+1:  Nicely done. There's a couple improvements suggested for the service, and some elastic mappings that need to be fixed on the model (please go over all `not_analyzed` and `unique` configs in the old schema)

server/planning/types/locations.py
```diff
@@ -0,0 +1,79 @@
...
+    country: str | None = None
+    locality: str | None = None  # Aka city
+    area: str | None = None
+    external: dict | None = None
+    boundingbox: list | None = None
+    address_type: str | None = Field(alias="type", default=None)
+
+
+class LocationResourceModel(BasePlanningModel):
+    guid: Annotated[fields.Keyword, Field(default_factory=lambda: generate_guid(type=GUID_NEWSML))]
```
MarkLark86: This field also the `validate_iunique_value_async` validation, to ensure data integrity that this field will always be unique, as we support this field being provided outside the model (aka not using the default factory)

MarkLark86:
@BrianMwangi21 Just the one last validation issue, then we can merge this one

MarkLark86 (approved):

PR #2220, opened 2025-03-06
[SDESK-7519] - Migrate the Planning Duplicate action from a resource/service to a web endpoint

BrianMwangi21:
### Purpose
This PR upgrades planning duplicate action resources to async endpoints

- Create endpoints for planning duplicate action
- Applied auth rules to the new endpoint
- Updated the pure functions for planning duplicate action based off old implementation.
- Regitered endpoint to planning module
- Extra: Added async `on_duplicated` function to PlanningAsyncService as it is used by planning duplicate and was in old planning service.

Solves: [SDESK-7519](https://sofab.atlassian.net/browse/SDESK-7519)

[SDESK-7519]: https://sofab.atlassian.net/browse/SDESK-7519

### DIFF
--- a/server/planning/planning/views.py
+++ b/server/planning/planning/views.py
@@ -10,14 +10,14 @@ from superdesk.core.web import EndpointGroup
 from superdesk.core.types import Request, Response
 
 
-planning_blueprint: EndpointGroup = EndpointGroup("planning", __name__)
+planning_endpoint_group: EndpointGroup = EndpointGroup("planning", __name__)
 
 
 class PlanningArgs(BaseModel):
     planning_id: str
 
 
-@planning_blueprint.endpoint(
+@planning_endpoint_group.endpoint(
     "planning/spike/<string:planning_id>",
     name="planning_spike",
     methods=["PATCH"],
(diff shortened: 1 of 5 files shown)


server/planning/planning/views.py
```diff
@@ -48,3 +49,19 @@ async def unspike_planning_item(args: PlanningArgs, params: None, request: Reque
...
+    auth=[required_privilege_rule("planning_planning_management")],
+)
+async def duplicate_planning_item(args: PlanningArgs, params: None, request: Request) -> Response:
+    original = await PlanningAsyncService().find_by_id_raw(args.planning_id)
+    if not original:
+        await request.abort(404, "Planning Item not found")
+
+    duplicated_planning_item = await process_planning_item_duplicate(original)
+
+    return Response(duplicated_planning_item)
```
MarkLark86, line 67: The old resource actually returns the entire new item, that way the front-end can use the newly created item without sending another API request for the new item.

In the old eve resources, when the `create` method returns IDs, Eve will get those items with that ID and use that in the response to the client.

MarkLark86 (commented):
Looks good, just one small bug with the endpoint response

MarkLark86 (approved):

PR #2225, opened 2025-03-09
[SDESK-7497] Partially fixes behave tests

eos87:
### Purpose
To fix as many behave tests as possible. 

### What has changed
- Adjusted and implemented missing functionality in the new async planning service (all planning.feature tests should pass now, although old service is still in place)
- Fixed issues with databases connections leak (`Too many files open` error) 
- Went through many tests cases and fixed them accordingly 
- Implemented `hypercorn` service for e2e tests 
- Commented and skipped some tests 
- Implemented logging configuration in behave tests (using same as e2e one)
- Set some of the signals back in place as some tests depend on them (until new ones are fully implemented)

### What's pending
- Fix the remaining broken tests
- Fix pytest as some of them got broken
- Many tests are broken because of this https://github.com/superdesk/superdesk-core/pull/2853#discussion_r1986440299. They will most likely be fixed by some work Mark's doing.

Thanks for checking!

### DIFF
--- a/server/features/planning.feature
+++ b/server/features/planning.feature
@@ -50,7 +50,7 @@ Feature: Planning
         """
             {"_items": [{
                 "guid": "__any_value__",
-                "_type": "planning",
+                "type": "planning",
                 "original_creator": "__any_value__",
                 "item_class": "item class value",
                 "headline": "test headline",
@@ -143,7 +143,7 @@ Feature: Planning
         """
             {"_items": [{
                 "guid": "__any_value__",
-                "_type": "planning",
+                "type": "planning",
                 "original_creator": "__any_value__",
                 "item_class": "item class value",
                 "headline": "test headline",
@@ -3775,7 +3775,7 @@ Feature: Planning
         {
             "slugline": "file test",
             "planning_date": "2016-01-02",
-            "_type": "planning",
+            "type": "planning",
             "coverages": [
                 {
                     "workflow_status": "draft",
@@ -3829,7 +3829,7 @@ Feature: Planning
         """
         {
             "slugline": "file test",
-            "_type": "planning",
+            "type": "planning",
             "coverages": [
                 {
                     "workflow_status": "draft",
@@ -3869,7 +3869,7 @@ Feature: Planning
         {
             "slugline": "file test",
             "planning_date": "2016-01-02",
-            "_type": "planning",
+            "type": "planning",
             "coverages": [
                 {
                     "workflow_status": "draft",
@@ -3923,7 +3923,7 @@ Feature: Planning
         """
         {
             "slugline": "file test",
-            "_type": "planning",
+            "type": "planning",
             "coverages": [
                 {
                     "workflow_status": "draft",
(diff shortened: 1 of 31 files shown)


MarkLark86:
@eos87 There are some pytest failing. One of them due to the EventHistory async service [not awaiting the user_id](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/events/events_history_async_service.py#L56).

It would be good if we can get all the pytest passing

server/planning/planning/planning_service.py
```diff
@@ -176,6 +178,75 @@ async def create(self, _docs: list[PlanningResourceModel | dict[str, Any]]) -> l
...
+                session=session_id,
+                event_ids=get_related_event_ids_for_planning(doc_dict, "primary"),  # Event IDs for primary events
+            )
+
+            await self._update_event_history(doc_dict)
+            await signals.planning_created.send([doc_dict])
+
+            first_primary_event_id = get_first_related_event_id_for_planning(doc_dict, "primary")
+            if first_primary_event_id and post_planning_with_event:
+                event = get_resource_service("events").find_one(req=None, _id=first_primary_event_id)
```
MarkLark86: We can use the Event async service here

server/planning/planning/planning_service.py
```diff
@@ -611,39 +698,41 @@ def validate_and_remove_coverage_entity(
...
+        ]
 
         for coverage in updated_planning.coverages:
             coverage_id = coverage.coverage_id or ""
             is_temporal_coverage = TEMP_ID_PREFIX in coverage_id
 
             if not coverage_id or is_temporal_coverage or coverage_id not in original_coverage_ids:
-                if "duplicate" in coverage_id or coverage.original_coverage_id:
-                    coverage.planning.xmp_file = self.duplicate_xmp_file(coverage.to_dict())
+                # if "duplicate" in coverage_id or coverage.original_coverage_id:
```
MarkLark86: What was wrong with this one?

server/planning/types/common.py
```diff
@@ -127,14 +128,20 @@ class Place:
...
+    """
+    Inherit from `ResourceModel` otherwise `get_model_annotations` won't pick it up
+    to run the async validators
+    """
+
+    id: Annotated[fields.Keyword | None, validate_data_relation_async("events")] = Field(alias="_id")
     recurrence_id: fields.Keyword | None = None
     link_type: LinkType | None = None
 
+    model_resource_name = "events"
```
MarkLark86: I would make this `related_event`, so that the name of this resource is unique

server/features/planning.feature
```diff
@@ -50,7 +50,7 @@ Feature: Planning
         """
             {"_items": [{
                 "guid": "__any_value__",
-                "type": "planning",
```
MarkLark86: `type` is actually required, as it is used by the publishing system (`type` is a field available through all content resources, and so we're using it from Planning based resources too).

`_type`: Resource Type
`type`: Item Type - In content authoring this is the sub-type, such as `text`, `picture`, `video` etc

MarkLark86 (commented):
Looks good, just a couple things

server/planning/planning/planning_service.py
eos87: debug leftover, sorry.

MarkLark86 (approved):

MarkLark86:
Closing and created another PR based on these changes

PR #2226, opened 2025-03-11
Assignment emails: Adding the 'News Coverage Status' in the assignment [SDBELGA-937]

devketanpro:
## Front-end checklist


server/planning/planning/planning.py
```diff
@@ -875,6 +875,12 @@ def _create_update_assignment(
         translated_value = {}
         translated_name = planning.get("name", planning.get("headline", ""))
         doc.setdefault("planning", {})
+
+        # SDBELGA-937
+        doc["planning"]["news_coverage_status"] = updates.get("news_coverage_status") or original.get(
```
MarkLark86, line 880: I don't see `news_coverage_status` in the Assignment schema (or rather, the Coverage.Planning schema). Looks like this will need to be added

MarkLark86 (approved):

PR #2236, opened 2025-03-18
Coverage profiles: UI issue when a contact is selected

dzonidoo:
STT-182

## Front-end checklist

### DIFF
--- a/client/components/Contacts/ContactField.tsx
+++ b/client/components/Contacts/ContactField.tsx
@@ -10,6 +10,7 @@ import {IContact, Omit} from 'superdesk-api';
 import {showModal} from '@sourcefabric/common';
 import {IContactFieldProps, IContactReduxStateProps, IContactReduxDispatchProps} from './ContactField.interface';
 import {Row} from './../../components/UI/Form';
+import {Spacer} from 'superdesk-ui-framework/react';
 
 const mapStateToProps = (state) => ({
     contacts: selectors.general.contacts(state),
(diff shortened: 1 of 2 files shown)


tomaskikutis (approved):

client/components/Contacts/ContactField.tsx
```diff
@@ -107,11 +106,7 @@ class ContactFieldComponent extends React.Component<IContactFieldProps> {
         }
 
         return (
-            <div
-                ref={refNode}
-                className={paddingTop ? 'contact-field--padding-top' : null}
-                data-test-id={this.props.testId}
```
tomaskikutis, line 113: can you keep the existing data-test-id and ref?

tomaskikutis (changes requested):

tomaskikutis:
keep existing ref too

tomaskikutis (approved):

tomaskikutis:
Ensure CI passes before merging. Eiher restart CI to check if it's an unstable test that's failing or fix it if it's a genuine failure.

dzonidoo:
@tomaskikutis CI failing on feature/embedded-events-editor as well... do i need to fix it in this PR?

tomaskikutis:
I will first fix `feature/embedded-events-editor` so it passes there and then we'll re-run tests on your branch and see if there's anything to fix

tomaskikutis:
fixed it on https://github.com/superdesk/superdesk-planning/tree/feature/embedded-events-editor you can merge it and see if it passes here

PR #2237, opened 2025-03-18
[SDESK-7515] - Upgrade the Event and Planning search endpoint to async

BrianMwangi21:
### Purpose
This PR upgrades events and planning search endpoints to async

- Convert EventsPlanningService to use the AsyncBaseService
- Convert PlanningSearchService to use the AsyncBaseService
- Updated code using `planning_search` and `events_planning_search` to use await where possible

**NOTE:** Makes use of new search with projections added in this [PR SDESK-7531](https://github.com/superdesk/superdesk-core/pull/2862/files) which needs to be merged first.

Solves: [SDESK-7515](https://sofab.atlassian.net/browse/SDESK-7515)

[SDESK-7515]: https://sofab.atlassian.net/browse/SDESK-7515

### DIFF
--- a/server/planning/search/eventsplanning_search.py
+++ b/server/planning/search/eventsplanning_search.py
@@ -18,13 +18,15 @@ from copy import deepcopy
 from werkzeug.datastructures import MultiDict, ImmutableMultiDict
 from eve.utils import ParsedRequest
 
-from superdesk import Resource, Service, get_resource_service
+from superdesk import Resource, get_resource_service
+from superdesk.eve_async.service import AsyncBaseService
 from superdesk.resource_fields import ITEMS
 from superdesk.resource import build_custom_hateoas
 from superdesk.errors import SuperdeskApiError
 
-from planning.planning.planning import planning_schema
 from planning.events.events_schema import events_schema
+from planning.planning.planning import planning_schema
+from planning.search.eventsplanning_filters_service import EventsPlanningFiltersAsyncService
 
 from .queries.planning import PLANNING_PARAMS, PLANNING_SEARCH_FILTERS
 from .queries.events import EVENT_PARAMS, EVENT_SEARCH_FILTERS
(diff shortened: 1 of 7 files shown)


server/planning/search/eventsplanning_search.py
```diff
@@ -296,9 +299,9 @@ def search_by_filter_id(self, filter_id, args=None, page=1, page_size=None, proj
 
         args["filter_id"] = filter_id
 
-        return self.search_repos(search_filter["item_type"], args, page, page_size, projections)
+        return await self.search_repos(search_filter["item_type"], args, page, page_size, projections)
 
-    def get_locked_items(self, repo=None, page=None, page_size=None, projections=None):
+    async def get_locked_items(self, repo=None, page=None, page_size=None, projections=None):
```
MarkLark86: As this function is only used if client arg `lock_state` is used, we can keep it sync here and comment to upgrade to async when adding support for [search param async callbacks](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/search/queries/common.py#L487)

server/planning/search/planning_search.py
```diff
@@ -63,7 +108,7 @@ def get(self, req, lookup):
 
             params["_source"] = fields
 
-        docs = self.elastic.search(query, types, params)
+        docs = await self.elastic.search_async(query, types, params)
```
MarkLark86: Elastic async support is not yet added to Eve datalayer. What you could use instead is the Events resource to get a client handle for Elastic (both Events & Planning run on the same es instance).

```python

async def get_async(self, req, lookup):
    ...
   docs = await EventResourceModel.get_service().elastic.search(query, types, params)
```

See how that goes

MarkLark86 (commented):
Looks good, just a couple small things

server/planning/search/planning_search.py
BrianMwangi21: I checked the definition of the search and it takes a query, and indexes : 
```    
async def search(self, query: Dict[str, Any], indexes: Optional[List[str]] = None) -> Any:
```

I can get the indexes using the `_get_index` in the class but I'm not so sure what to do with the `types` and `fields` as they were required in the previous search implementation. 

Thoughts ?

MarkLark86: **types**
This is the indexes to use, either provided by the client or a default. Try using:

```python
def get_indexes_for_search(repos: list[str]) -> list[str]:
    indexes = []
    if "events" in repos:
        indexes += EventResourceModel.get_service().elastic.config.index
    if "planning" in repos:
        indexes += PlanningResourceModel.get_service().elastic.config.index
    return indexes
```

**params**
In the current code all this is doing is providing field projection. I have added it in a PR in core now, which accepts a list of field names to use.

So you can use:
```python
def get_projection(req) -> list[str]:
    fields = self._get_projected_fields(req)
    projection = None if not fields else fields.split(",")

    if projection and "type" not in projection:
        # Make sure `type` is always included in the projection
        projection.append("type")

    return projection
```

**Putting that together**
you would get something like:
```python

def get(self, req, lookup):
    query = self._get_query(req)
    indexes = get_indexes_for_search(self._get_types(req))
    projection = get_projection(req)
    elastic = EventResourceModel.get_service().elastic
    docs = await elastic.search(query, indexes, projection)
    self._format_docs(docs)
```

Or something along those lines

server/planning/search/planning_search.py
```diff
@@ -118,19 +95,56 @@ def _format_docs(self, docs):
...
+
+    async def get_async(self, req, lookup):
+        """Run the query against events and planning indexes"""
+        query = self._get_query(req)
+        types = self._get_types(req)
+        indexes = self.get_indexes_for_search(types)
+        projection = self.get_projection(req)
+
+        elastic = EventResourceModel.get_service().elastic
+        docs = await elastic.search(query, indexes, projection)
```
MarkLark86: I have been working on something similar on the content authoring side (upgrading global content search to async), and there may need to be a couple more fixes here.

```python
from eve_elastic.elastic import fix_query

async def get_async(self, req, lookup):
    ...
    hits = await elastic.search(fix_query(query), indexes, projection)
    docs = self.elastic._parse_hits(hits, types[0])
```

You'll see 2 changes here:
* The use of `fix_query` from the `eve_elastic` library (as some front-end code still used Elastic version 1 and/or  2 queries  :face_exhaling: 
* Using the `_parse_hits` from the `eve_elastic` library to process the response

I found these 2 extra steps were needed so the response is in the correct format etc

MarkLark86 (approved):

PR #2238, opened 2025-03-19
Assignment mails - Language specific mail based on the language of the coverage (including the related articles) [SDBELGA-940]

devketanpro:


server/planning/assignments/assignments.py
```diff
@@ -480,6 +480,12 @@ def send_assignment_notification(self, updates, original=None, force=False):
             else None
         )
 
+        coverage_language = assignment.get("planning", {}).get("language")
```
MarkLark86: Seems like `coverage_language` is only used inside the following `if` statement, might be good to localise it there to indicate it's only used in that statement

MarkLark86 (approved):

PR #2253, opened 2025-03-28
Improve drag and drop for content profile fields, improve options menu

thecalcc:
STT-187

## Front-end checklist

### DIFF
--- a/client/components/ContentProfiles/FieldTab/FieldList.tsx
+++ b/client/components/ContentProfiles/FieldTab/FieldList.tsx
@@ -5,9 +5,10 @@ import {superdeskApi} from '../../../superdeskApi';
 
 import {getFieldNameTranslated, getProfileGroupNameTranslated} from '../../../utils/contentProfiles';
 
-import {Button, IconButton, ToggleBox, Menu, Label} from 'superdesk-ui-framework/react';
+import {Button, IconButton, ToggleBox, Label, TreeMenu} from 'superdesk-ui-framework/react';
 import * as List from '../../UI/List';
-import SortItems from '../../SortItems';
+import {arrayMove, WithSortable} from '@sourcefabric/common';
+import {shouldNotStartDragging} from '../utils';
 
 interface IProps {
     profile: IEditorProfile;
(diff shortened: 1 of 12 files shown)


client/components/ContentProfiles/FieldTab/FieldList.tsx
```diff
@@ -150,40 +164,31 @@ export class FieldList extends React.PureComponent<IProps> {
...
                     items={this.props.fields}
-                    getListElement={this.getListElement}
-                    useCustomStyle={true}
-                    lockAxis="y"
-                    lockToContainerEdges={true}
-                    distance={10}
+                    getId={(item) => item.name}
+                    itemTemplate={(item) => <>{this.getListElement(item.item)}</>}
+                    options={{
+                        shouldCancelStart: shouldNotStartDragging,
```
tomaskikutis: It would be cleaner to use `distance` here so it wouldn't trigger dragging if you're simply clicking on something. It would work regardless of classnames then which is nicer.

client/components/ContentProfiles/GroupTab/GroupList.tsx
```diff
@@ -55,7 +57,10 @@ export class GroupList extends React.PureComponent<IProps> {
                         icon="plus-large"
                         shape="round"
                         type="primary"
-                        onClick={() => this.props.insertGroup(group.index - 0.1)}
+                        onClick={(e) => {
+                            e.stopPropagation();
```
tomaskikutis: isn't preventDefault enough?

client/components/ContentProfiles/FieldTab/FieldList.tsx
```diff
@@ -28,6 +29,48 @@ export class FieldList extends React.PureComponent<IProps> {
         super(props);
 
         this.getListElement = this.getListElement.bind(this);
+        this.getTreeMenu = this.getTreeMenu.bind(this);
+    }
+
+    getTreeMenu(options: Array<{value: IProfileFieldEntry; onSelect: () => void;}>, buttonLabel: string) {
```
tomaskikutis: If you're returning a JSX - it should be a separate react component - not a class method.

client/components/ContentProfiles/GroupTab/GroupList.tsx
thecalcc: yep, I don't even need that since I got `distance` option to work

thecalcc:
I made `distance` work in both places, so no need for any other function. I had a different setup for the two places we use the sortable hoc.

client/components/ContentProfiles/utils.ts
```diff
@@ -1,3 +1,4 @@
+import {SyntheticEvent} from 'react';
```
tomaskikutis: remove unused import

client/components/ContentProfiles/GroupTab/GroupList.tsx
tomaskikutis: what about other usages you added in this PR?

client/components/ContentProfiles/utils.ts
thecalcc: For some reason I can't respond to the other comment, but I've dropped the import and removed other usages

tomaskikutis (approved):

e2e/cypress/support/planning/manageContentProfiles.ts
```diff
@@ -76,6 +77,6 @@ export class ManageContentProfiles extends Modal {
     getHeaderButton(label: string) {
         return this.element.find('.side-panel__header')
             .contains(label);
-            // .should('exist')
+        // .should('exist')
```
tomaskikutis: I guess this shouldn't be commented out?

PR #2258, opened 2025-04-02
Fix: When a coverage is reassigned to another desk and user, the assignment email is not sent [SDBELGA-943]

devketanpro:


server/planning/assignments/assignments.py
```diff
@@ -637,9 +640,11 @@ def send_assignment_notification(self, updates, original=None, force=False):
                         slugline=slugline,
                         client_url=client_url,
                         assignment_id=assignment_id,
-                        assignor="by " + user.get("display_name", "")
-                        if str(user.get(config.ID_FIELD, None)) != assigned_to.get("user", "")
-                        else "to yourself",
+                        assignor=(
+                            "by " + user.get("display_name", "")
+                            if str(user.get(config.ID_FIELD, None)) != assigned_to.get("user", "")
+                            else "to yourself"
```
petrjasek: we will need those strings translated

petrjasek (commented):
test pls

petrjasek (approved):

PR #2259, opened 2025-04-03
Fix item load and lock

thecalcc:
STT-189

## Front-end checklist

### DIFF
--- a/client/components/Main/ItemEditor/Editor.tsx
+++ b/client/components/Main/ItemEditor/Editor.tsx
@@ -200,6 +200,7 @@ export class EditorComponent extends React.Component<IEditorProps, IEditorState>
         this.autoSave.flushAutosave().then(() => {
             const {openCancelModal, itemId, itemType, addNewsItemToPlanning} = this.props;
             const {dirty, errorMessages, initialValues} = this.state;
+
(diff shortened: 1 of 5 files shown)


tomaskikutis (approved):

thecalcc:
@tomaskikutis test changes explained:

I removed `prevProps` from `onItemIDChanged` in `client/components/Main/ItemEditor/ItemManager.ts:211`, because I saw it's unused. This explains why I removed argument checks in a couple tests (something like `expect(manager.**onItemIDChanged**.args[0])`). 

Besides this I passed the `'edit'` to `lockItem` in `client/components/Main/ItemEditor/ItemManager.ts:375` in order to fix the lint errors there. And my change is also correct there because on line 352 in the same file we check if `itemAction === 'edit'`. So I had to change tests from this:

```
expect(planningApi.locks.lockItem.args[0]).toEqual([testData.events[0]]);
```

to something like:

```
expect(planningApi.locks.lockItem.args[0]).toEqual([testData.events[0], 'edit']); <-- includes 'edit' argument
```

All changes are of this type, so that should be all.

tomaskikutis (approved):

PR #2260, opened 2025-04-03
Add the (Belga keywords) metadata to the preview of the related articles (from the archive) in the Planning [SDBELGA-932]

devketanpro:


petrjasek:
afaik belga keywords is a custom vocabularies, not the keywords core field

devketanpro:
> afaik belga keywords is a custom vocabularies, not the keywords core field

yep but in related article its stored under keywords attribute

petrjasek:
shouldn't we rather change the parser than? that seems like wrong data

devketanpro:
My bad, it's actually stored under `subject`. Both keywords and Belga keywords have the same values, so I just got a bit confused—never mind!

client/components/fields/editor/EventRelatedArticles/PreviewArticle.tsx
```diff
@@ -95,6 +102,30 @@ export class PreviewArticle extends React.PureComponent<IProps, IState> {
                                     </div>
                                 ))
                         }
+                        {this.props.item.subject?.length > 0 && (subjectKey || Object.keys(subjectSchemes).length === 0)
+                         && (
+                             <div className="tr">
+                                 <div className="td sd-padding-b--0-5">
+                                     <span className="form-label">
+                                         {gettext(subjectKey ? subjectSchemes[subjectKey] || 'Subjects' : 'Subjects')}
```
petrjasek: I think we have some helper in core to get the proper label

client/components/fields/editor/EventRelatedArticles/PreviewArticle.tsx
```diff
@@ -61,6 +61,13 @@ export class PreviewArticle extends React.PureComponent<IProps, IState> {
         );
         const headlineField = extractedFields['headline'];
         const bodyField = extractedFields['body_html'];
+        const subjectSchemes = appConfig?.SUBJECT_SCHEME_LABELS || {};
```
petrjasek: I would rather avoid some extra config for this and rather show every CV we have stored on the instance, like we do in monitoring preview

petrjasek:
would be good to group this by vocabulary and only show those which have label defined in a vocabulary

petrjasek (approved):

PR #2269, opened 2025-04-09
Stt fixes

thecalcc:
STT-189

## Front-end checklist

### DIFF
--- a/client/api/editor/item_planning.ts
+++ b/client/api/editor/item_planning.ts
@@ -102,13 +102,13 @@ export function getPlanningInstance(type: EDITOR_TYPE): IEditorAPI['item']['plan
         const index = events.findIndex(
             (event) => event._id === original._id
         );
-        const isFirstRelatedEvent = events.length < 1;
+        const isEmpty = events.length < 1;
 
-        if (!isFirstRelatedEvent && index < 0) {
+        if (!isEmpty && index < 0) {
             return;
         }
 
-        events[isFirstRelatedEvent ? 0 : index] = {
+        events[isEmpty ? 0 : index] = {
             _id: updates._id,
             link_type: original.link_type,
             recurrence_id: original.recurrence_id
--- a/client/api/editor/item_planning.ts
+++ b/client/api/editor/item_planning.ts
@@ -102,19 +102,20 @@ export function getPlanningInstance(type: EDITOR_TYPE): IEditorAPI['item']['plan
         const index = events.findIndex(
             (event) => event._id === original._id
         );
+        const isFirstRelatedEvent = events.length < 1;
 
-        if (index < 0) {
+        if (!isFirstRelatedEvent && index < 0) {
             return;
         }
 
-        events[index] = {
+        events[isFirstRelatedEvent ? 0 : index] = {
             _id: updates._id,
             link_type: original.link_type,
             recurrence_id: original.recurrence_id
         };
 
         const updateMainField = () => {
-            editor.form.changeField('related_events', events)
+            return editor.form.changeField('related_events', events, true, true)
                 .then(() => {
                     if (scrollOnChange) {
                         getRelatedEventsDomRef(original._id).current?.scrollIntoView();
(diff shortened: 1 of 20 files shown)


client/api/editor/item_planning.ts
```diff
@@ -102,19 +102,20 @@ export function getPlanningInstance(type: EDITOR_TYPE): IEditorAPI['item']['plan
         const index = events.findIndex(
             (event) => event._id === original._id
         );
+        const isFirstRelatedEvent = events.length < 1;
```
tomaskikutis: what does this mean?

client/components/editor-standalone/authoring-storage-in-memory.ts
```diff
@@ -67,6 +67,9 @@ export function getAuthoringStorageInMemory<T extends IEventOrPlanningItem>(
                     } else {
                         return {...x} as unknown as T;
                     }
+                })
+                .catch(() => {
```
tomaskikutis, line 71: is this for a case when nothing is autosaved yet? add a comment if so

client/components/editor-standalone/save-handling.ts
```diff
@@ -46,23 +46,28 @@ export const handleEmbeddedItems = async<T extends IEventItem | IPlanningItem>(
     action: IEmbeddedPlanningsActionType,
     itemType: ItemType,
 ): Promise<Array<T>> => {
-    const updatedItems: Array<T> = [];
+    const updatedItems: Array<Promise<T>> = [];
```
tomaskikutis: I'm not seeing what's changed here?

client/components/fields/editor/AssociatedEvent.tsx
```diff
@@ -102,20 +123,32 @@ export class EditorFieldAssociatedEventComponent extends React.PureComponent<IAs
...
-                                isTemporary={isTemporaryId(event._id)}
                                 index={i}
                                 key={event._id}
                                 event={event}
-                                updateEventItem={this.props.updateEventItem}
-                                unlinkEvent={this.props.unlinkEvent}
+                                planningItem={this.props.item}
+                                updateEventItem={(item, updates, scrollOnChange) =>
+                                    this.props.updateEventItem(item as IEventItem, updates, scrollOnChange).then(() => {
+                                        setTimeout(() => {
```
tomaskikutis: what's the timeout for?

client/components/fields/editor/AssociatedEventItem.tsx
```diff
@@ -107,11 +117,22 @@ export class AssociatedEventItem extends React.PureComponent<IProps> {
                                         ['_endTime', '_startTime', '_created', '_etag', '_links', '_updated']
                                     );
 
-                                    return eventsApi.events.create(itemClean).then(([createdEvent]) => {
-                                        this.update(createdEvent);
+                                    // Sometimes UI is clicked too fast, and since we don't have a perfect flow
```
tomaskikutis: You mean plus button? Can you disable the button after the first click until promise resolves? It's not clean to handle it here

client/components/Main/ItemEditor/ItemManager.ts
```diff
@@ -715,42 +699,20 @@ export class ItemManager {
...
-                    (updates.related_events ?? []).length > 0
-                    || ((updates as IPlanningItem)._unsaved_related_events ?? []).length > 0
-                )
-            ) {
-                const [_withNewEvents, withExistingEvents] = partition(
-                    ((updates as IPlanningItem).related_events ?? []) as Array<IPlanningRelatedEventLink>,
-                    (x) => !isTemporaryId(x._id),
-                );
-
-                const updatedLinks = eventUtils.addRelatedEvents(
```
tomaskikutis, line 738: How come this is dropped? Won't it break other things?

client/components/editor-standalone/save-handling.ts
thecalcc: Await isn't working in the for loop so I reworked it, so that all items are resolved at once.

client/components/editor-standalone/authoring-storage-in-memory.ts
thecalcc: Exactly

client/api/editor/item_planning.ts
thecalcc: A check if there's any related_events in the array already.

If there is non -> ignore the calculated `index` (since it will always be -1), and put the element in the first place.

client/components/fields/editor/AssociatedEvent.tsx
thecalcc: Reloading the locks on the frontend for related items, and the call of `this.props.updateEventItem` is used for updating the field.

client/components/editor-standalone/save-handling.ts
```diff
@@ -45,18 +45,23 @@ export const handleEmbeddedItems = async<T extends IEventItem | IPlanningItem>(
     editorType: EDITOR_TYPE,
     action: IEmbeddedPlanningsActionType,
     itemType: ItemType,
-): Promise<Array<T>> => {
+) => {
```
tomaskikutis: keep return type

tomaskikutis (approved):

PR #2272, opened 2025-04-10
[SDESK-7567] - Migrate usage of the Agenda resource service to the Pydantic based async resource

BrianMwangi21:
### Purpose
This PR migrate usage of the Agenda resource service to the Pydantic based async resource

- Migrated use of Agenda resource service into using the new async pydantic ones
- Set old services as internal resources as they are still used for some other functions

Solves: [SDESK-7567](https://sofab.atlassian.net/browse/SDESK-7567)

[SDESK-7567]: https://sofab.atlassian.net/browse/SDESK-7567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/agendas_async/agendas_async_service.py
+++ b/server/planning/agendas_async/agendas_async_service.py
@@ -8,7 +8,7 @@ from superdesk.errors import SuperdeskApiError
 from superdesk.notification import push_notification
 
 
-async def generate_planning_info(docs):
+async def generate_planning_info(docs: list[dict[str, Any]]):
     planning_service = PlanningAsyncService()
     for doc in docs:
         cursor = await planning_service.get_planning_by_agenda_id(doc.get(ID_FIELD))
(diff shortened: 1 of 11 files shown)


server/planning/agendas_async/agendas_async_service.py
```diff
@@ -1,26 +1,22 @@
...
-    #     planning_service = get_resource_service("planning")
-    #     for doc in docs:
-    #         doc["plannings"] = planning_service.get_planning_by_agenda_id(doc.get(ID_FIELD)).docs
-    #
-    # async def on_fetched(self, docs):
-    #     await self._generate_planning_info(docs.get(ITEMS))
-    #
-    # async def on_fetched_item(self, doc):
-    #     await self._generate_planning_info([doc])
+async def generate_planning_info(docs):
```
MarkLark86: It would be good to have types here, in this case it should be `(docs: list[dict]) -> None:`

MarkLark86 (approved):
Looks good, just 1 small improvement

PR #2274, opened 2025-04-11
Fix: keywords are not showing in the preview[SDBELGA-932]

devketanpro:
Fixes issue where keywords were not displaying in the article preview. 
Previously, the logic relied only on custom vocabularies, but belga_keywords are not present in that list. Now, using getAll() to fetch all vocabularies and matching schemes accordingly to ensure keywords render correctly.


petrjasek:
we don't need it in 2.8, this should go to develop now

PR #2275, opened 2025-04-11
[SDESK-7568] - Migrate usage of the Autosave resource service to the Pydantic based async resource

BrianMwangi21:
### Purpose
This PR migrate usage of the Autosave resource service to the Pydantic based async resource

- Migrated use of Autosave resource service into using the new async pydantic ones
- Removed old services and resources as they are no longer used
- Refactored file names to simply be `autosave_service` and not `autosave_async_service` since old implementation has been removed.

Solves: [SDESK-7568](https://sofab.atlassian.net/browse/SDESK-7568)

[SDESK-7568]: https://sofab.atlassian.net/browse/SDESK-7568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/commands/purge_expired_locks.py
```diff
@@ -124,7 +127,7 @@ async def purge_item_locks(resource: str, expiry_datetime: datetime):
 
             try:
                 # Delete any autosave items associated with this item
-                autosave_service.delete_action(lookup={"_id": item_id})
+                autosave_service.delete_many(lookup={"_id": item_id})
```
MarkLark86: I believe this is an async function, so we should await it

MarkLark86 (commented):
Just 1 issue found, rest looks good

MarkLark86 (approved):

PR #2277, opened 2025-04-11
Standalone saving for embedded plannings

thecalcc:
STT-189

Not a direct ticket link.

Purpose: 

Harmonize saving behavior between embedded events and embedded plannings. 

With this change embedded plannings can now actually be saved by themselves first, without being linked right away, if user chooses so. And then linked through main editor save button, or if user chooses not to - they can be discarded through "Cancel" and "Discard", while still being in the system as standalone items. 

## Front-end checklist

### DIFF
--- a/client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
@@ -133,6 +133,7 @@ export class RelatedPlanningItem extends React.PureComponent<IProps> {
                                     (item) => {
                                         // Remove fields responsible for creating a link, so that item can be
                                         // standalone until user decides to link it through main editor "Save"
+
                                         const itemClean = omit(
                                             modifyForServer(item, true),
                                             ['_temporary', 'related_events', '_created', '_etag', '_links', '_updated'],
--- a/client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
@@ -133,10 +133,12 @@ export class RelatedPlanningItem extends React.PureComponent<IProps> {
                                     (item) => {
                                         // Remove fields responsible for creating a link, so that item can be
                                         // standalone until user decides to link it through main editor "Save"
-
+                                        const fieldsToOmit = [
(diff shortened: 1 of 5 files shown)


tomaskikutis:
Could describe what's being done? Feature added, bug fixed?

thecalcc:
> Could describe what's being done? Feature added, bug fixed?

You were too quick to write the comment :D I was writing it already

tomaskikutis:
There's a feature to write a PR description before opening a PR ;)

tomaskikutis:
Was this a request? To me it seems nicer that if I add related items, but change my mind and discard - that they wouldn't be created to begin with

thecalcc:
> Was this a request? To me it seems nicer that if I add related items, but change my mind and discard - that they wouldn't be created to begin with

Request was to harmonize the behavior to be the same for both embedded events and planning items. After checking it with Gideon, he saw how both work and we decided to have it like how I made it for embedded events. 

Also if you wish to discard - you still can, just don't save. This is giving more options to the user also, because if you change your mind about linking, you can still have your items separate.

client/components/fields/editor/EventRelatedPlannings/RelatedPlanningItem.tsx
```diff
@@ -133,12 +131,25 @@ export class RelatedPlanningItem extends React.PureComponent<IProps> {
...
                                     (item) => {
-                                        this.update(item);
-
-                                        return Promise.resolve(item);
+                                        // Remove fields responsible for creating a link, so that item can be
+                                        // standalone until user decides to link it through main editor "Save"
+
+                                        const itemClean = omit(
+                                            modifyForServer(item, true),
+                                            ['_temporary', 'related_events', '_created', '_etag', '_links', '_updated'],
```
tomaskikutis: Ensure field names are typed.  It'd be best to move [this](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/helpers/utils.tsx#L39) to `@sourcefabric/common` and use here

tomaskikutis (approved):

thecalcc:
E2E Fix Explained: 

Waiting for the close button to be enabled, doesn't specifically mean that save would be done. Until this point it looks like this was working by accident. Since we now have a more complicated saving process this broke, and I actually had to make things proper by checking for the "Save" button to be disabled - that's when saving has been done and item has been successfully saved.

PR #2279, opened 2025-04-14
[SDESK-7569] - Migrate usage of PlanningTypes to the Pydantic based async resource

BrianMwangi21:
### Purpose
This PR migrate usage of the PlanningTypes resource service to the Pydantic based async resource

- Migrated use of PlanningTypes resource service into using the new async pydantic ones

Solves: [SDESK-7569](https://sofab.atlassian.net/browse/SDESK-7569)

[SDESK-7569]: https://sofab.atlassian.net/browse/SDESK-7569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/commands/populate_planning_types_test.py
+++ b/server/planning/commands/populate_planning_types_test.py
@@ -11,8 +11,8 @@
 import os
 import json
 
+from planning.content_profiles.planning_types_async_service import PlanningTypesAsyncService
 from planning.tests import TestCase
-from superdesk import get_resource_service
 from apps.prepopulate.app_populate import AppPopulateCommand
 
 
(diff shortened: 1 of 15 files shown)


BrianMwangi21:
@MarkLark86 I have a question on PlanningTypes. 

We have a function like this : 

```
def get_planning_schema(resource: str) -> ContentProfile:
    # TODO-ASYNC[PlanningTypesAsyncService] - Update to use PlanningTypesAsyncService when function is updated to async and everywhere else it is also called
    return get_resource_service("planning_types").find_one(req=None, name=resource)
```

The return type is ContentProfile and the `find_one` for the async service will return a PlanningTypesResourceModel. Should I refactor this function and where it's used to return the planning types resource model or should I cast what I get from `find_one` to a ContentProfile ? 

Also, what's the relationship/difference between ContentProfile and PlanningTypes ? 

Thanks

MarkLark86:
> @MarkLark86 I have a question on PlanningTypes.
> 
> We have a function like this :
> 
> ```
> def get_planning_schema(resource: str) -> ContentProfile:
>     # TODO-ASYNC[PlanningTypesAsyncService] - Update to use PlanningTypesAsyncService when function is updated to async and everywhere else it is also called
>     return get_resource_service("planning_types").find_one(req=None, name=resource)
> ```
> 
> The return type is ContentProfile and the `find_one` for the async service will return a PlanningTypesResourceModel. Should I refactor this function and where it's used to return the planning types resource model or should I cast what I get from `find_one` to a ContentProfile ?
> 
> Also, what's the relationship/difference between ContentProfile and PlanningTypes ?
> 
> Thanks

The ContentProfile is the dict version of a PlanningTypesResourceModel. This resource is used in the same way that ContentProfiles are used in superdesk-core (to define the fields/validation for a data type).

As to reduce complexity of this task, and reduce risk, I would say it's fine to use `to_dict()` to convert a PlanningTypesResourceModel to a dict, and use `cast` function to tell mypy that it's an instance of ContentProfile

BrianMwangi21:
@MarkLark86 I've made changes where possible. 

If I was to change the signature of `get_planning_schema` it affects more functions in the utils.py file which are called by more files that haven't been converted to async yet like the events_ml.py file.

MarkLark86 (approved):
Looks good

PR #2283, opened 2025-04-16
Saving fixes

thecalcc:
## Front-end checklist

### DIFF
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -9,7 +9,6 @@ import {
     POST_STATE,
     MAIN,
     TO_BE_CONFIRMED_FIELD,
-    TEMP_ID_PREFIX,
 } from '../../constants';
 import * as selectors from '../../selectors';
 import {
@@ -18,9 +17,7 @@ import {
     isExistingItem,
     isValidFileInput,
     isPublishedItemId,
-    isTemporaryId,
     gettext,
-    planningUtils,
 } from '../../utils';
 
 import planningApis from '../planning/api';
@@ -29,7 +26,7 @@ import main from '../main';
 import {eventParamsToSearchParams} from '../../utils/search';
 import {getRelatedEventIdsForPlanning} from '../../utils/planning';
 import {planning} from '../../api/planning';
-import * as actions from '../../actions';
+import {areEmbeddedItemsDirty} from '../../components/editor-standalone/utils';
 
 /**
  * Action dispatcher to load a series of recurring events into the local store.
@@ -646,8 +643,12 @@ const save = (original, updates) => (
                 planningApi.events.create(eventUpdates);
 
             return createOrUpdatePromise.then(([updatedEvent]: Array<IEventItem>) => {
-                if (updatedEvent?._id != null) {
-                    updateLinkedPlanningsForEvent(
+                const haveLinksChanged = updatedEvent?._id ? areEmbeddedItemsDirty(original, updatedEvent) : false;
+
+                if (!haveLinksChanged) {
+                    return [updatedEvent];
+                } else {
+                    return updateLinkedPlanningsForEvent(
                         updatedEvent._id,
                         updates.associated_plannings,
                     ).then((updatedPlannings) => {
@@ -657,8 +658,6 @@ const save = (original, updates) => (
 
                         return [updatedEvent];
                     });
-                } else {
-                    return [updatedEvent];
                 }
             });
         });
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -333,10 +333,13 @@ const fetchById = (eventId, {force = false, saveToStore = true, loadPlanning = t
         return promise.then((event) => {
             if (loadPlanning) {
                 return dispatch(self.loadAssociatedPlannings(event))
-                    .then(
-                        () => Promise.resolve(event),
-                        (error) => Promise.reject(error)
-                    );
+                    .then((plannings) => {
+                        return Promise.resolve({
+                            ...event,
+                            associated_plannings: plannings,
+                        });
+                    })
+                    .catch((error) => Promise.reject(error));
             }
 
             return Promise.resolve(event);
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -645,18 +645,22 @@ const save = (original, updates) => (
                 planningApi.events.update(originalItem, eventUpdates) :
                 planningApi.events.create(eventUpdates);
 
-            return createOrUpdatePromise.then(([updatedEvent]: Array<IEventItem>) =>
-                updateLinkedPlanningsForEvent(
-                    updatedEvent._id,
-                    updates.associated_plannings,
-                ).then((updatedPlannings) => {
-                    // Update associated events, so if a link has been added/removed etags are updated
-                    // after the change in planning items
-                    updatedEvent.associated_plannings = updatedPlannings;
+            return createOrUpdatePromise.then(([updatedEvent]: Array<IEventItem>) => {
+                if (updatedEvent?._id != null) {
+                    updateLinkedPlanningsForEvent(
+                        updatedEvent._id,
+                        updates.associated_plannings,
+                    ).then((updatedPlannings) => {
+                        // Update associated events, so if a link has been added/removed etags are updated
+                        // after the change in planning items
+                        updatedEvent.associated_plannings = updatedPlannings;
 
+                        return [updatedEvent];
+                    });
+                } else {
                     return [updatedEvent];
-                }),
-            );
+                }
+            });
         });
     }
 );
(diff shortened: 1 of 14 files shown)


client/actions/events/api.ts
```diff
@@ -628,74 +641,25 @@ const save = (original, updates) => (
...
+                    ).then((updatedPlannings) => {
+                        // Update associated events, so if a link has been added/removed etags are updated
+                        // after the change in planning items
+                        updatedEvent.associated_plannings = updatedPlannings;
+
+                        return [updatedEvent];
+                    });
+                } else {
+                    return [updatedEvent];
                 }
```
eos87: Minor suggestion. You could invert the if statement and have an early return and eliminate the need for an else `block`. What do you think?

```suggestion
                if (updatedEvent?._id == null)
                    return [updatedEvent];
                
                return updateLinkedPlanningsForEvent(
                    updatedEvent._id,
                    updates.associated_plannings,
                ).then((updatedPlannings) => {
                    // Update associated events, so if a link has been added/removed etags are updated
                    // after the change in planning items
                    updatedEvent.associated_plannings = updatedPlannings;
                    return [updatedEvent];
                });
```

client/api/editor/item_planning.ts
```diff
@@ -99,21 +99,35 @@ export function getPlanningInstance(type: EDITOR_TYPE): IEditorAPI['item']['plan
...
+                    link_type: original.link_type,
+                    recurrence_id: original.recurrence_id,
+                });
+            } else {
+                // This should never happen, but make sure to notify the user in some corruption case
+                superdeskApi.ui.notify.error(
+                    superdeskApi.localization.gettext(
+                        'Could not link event - A corruption occurred, please close the item and try linking again'
+                    ),
+                );
```
eos87: should this have a `return` statement to stop the execution too?

client/components/Main/ItemEditor/Editor.tsx
```diff
@@ -158,8 +165,19 @@ export class EditorComponent extends React.Component<IEditorProps, IEditorState>
...
-        return !itemsEqual(diff, initialValues);
+    isDirty(
+        initialValues: Partial<IEventOrPlanningItem>,
+        diff: Partial<IEventOrPlanningItem>,
+        checkRelatedItemsChanges = true,
+    ) {
+        const embeddedItemsAreDirty = checkRelatedItemsChanges ? areEmbeddedItemsDirty(initialValues, diff) : false;
+        const fieldsToOmit = ['associated_plannings', '_unsaved_related_events', 'embedded_planning'];
+
+        if (checkRelatedItemsChanges === false) {
```
eos87: nit:
```suggestion
        if (!checkRelatedItemsChanges) {
```

client/components/Main/ItemEditor/ItemManager.ts
```diff
@@ -415,7 +415,22 @@ export class ItemManager {
...
+                                initialValues,
+                                autosaveItem,
+                                false,
+                            );
+                        } else {
+                            return this.editor.isDirty(
+                                initialValues,
+                                autosaveItem,
+                            );
+                        }
```
eos87: Minor improvement/simplification. Thoughts?
```suggestion
                        const ignoreAssociatedPlannings = initialValues.type === 'event' && autosaveItem.associated_plannings === undefined;
                        
                        return this.editor.isDirty(
                            initialValues,
                            autosaveItem,
                            ignoreAssociatedPlannings ? false : undefined,
                        );
```

client/components/editor-standalone/authoring-storage-event-http.ts
```diff
@@ -45,6 +45,11 @@ export const authoringStorageEventItemHttp: IAuthoringStorage<IEventItem> = {
             headers: {
                 'If-Match': original._etag,
             },
+        }).catch((err) => {
+            // Handles date issues like start time should not be after end time
+            superdeskApi.ui.notify.error(getErrorMessage(err, gettext('Could not save event')));
```
eos87, line 50: do you think we should improve the feedback (message) so the user knows better what's the issue about?

client/components/editor-standalone/save-handling.ts
```diff
@@ -46,28 +46,29 @@ export const handleEmbeddedItems = async<T extends IEventItem | IPlanningItem>(
...
-            updatedItems.push(await exposed.getLatestItem());
+            updatedItems.push(Promise.resolve(latestItem));
         } else {
-            updatedItems.push(await exposed.handleUnsavedChanges());
+            updatedItems.push(exposed.handleUnsavedChanges());
         }
     }
 
-    return updatedItems;
+    return Promise.all(updatedItems);
```
eos87, line 71: just curious about this change. Why moving away from `await` to `Promises.all` usage?

client/components/editor-standalone/utils.ts
```diff
@@ -24,3 +25,42 @@ export function omitFields<T extends IBaseRestApiResponse>(
...
+        }).sort((a, b) => a._id.localeCompare(b._id));
+        const plansDiff = diffPlans.map((x) => {
+            const itemCleaned = omit(x, FIELDS_TO_OMIT_FOR_COMPARISON);
+
+            if (isMoment(itemCleaned.planning_date)) {
+                itemCleaned.planning_date = itemCleaned.planning_date.toISOString();
+            }
+
+            return itemCleaned;
+        }).sort((a, b) => a._id.localeCompare(b._id));
```
eos87: This looks very similar. I'd suggest to extract the `map-clean-sort` logic into a separate function to avoid duplication. Something like this

```typescript
const cleanAndSortPlans = (plans: any[]) => {
    return plans
        .map((item) => {
            const cleanedItem = omit(item, FIELDS_TO_OMIT_FOR_COMPARISON);
            if (isMoment(cleanedItem.planning_date)) {
                cleanedItem.planning_date = cleanedItem.planning_date.toISOString();
            }
            return cleanedItem;
        })
        .sort((a, b) => a._id.localeCompare(b._id));
}
```

and then reuse it. What do you think?

eos87 (commented):
Code-wise looks good. I just left a few minor comments.

client/components/editor-standalone/save-handling.ts
thecalcc: Good question, for some reason using `await` doesn't work, because the code that runs inside `expose.save()` is causing component remounting for all the embedded items, and then on the next iteration when trying to get `expose.save()` the error for can't access property of an unmounted component runs, which also breaks the saving of all embedded items. That's why I tried this, which seems to be working always

client/components/editor-standalone/authoring-storage-event-http.ts
thecalcc: Yep, that's a bit tough to do, I only encountered an error if event `start` time is after `end` time. And we do have a specific error message returned from the backend in `err`. That `err` is processes with `getErrorMessage` which extracts the string saying something like "Start time can't be after end time"

client/components/Main/ItemEditor/ItemManager.ts
thecalcc: way cleaner

client/components/Main/ItemEditor/Editor.tsx
thecalcc: I just prefer a more explicit style, the `!` takes into account all falsy values. In this case that could never happen, just a habit I developed because using `!` has caused issues before haha. I can change it here though

client/components/editor-standalone/utils.ts
thecalcc: yep, will do that

client/api/editor/item_planning.ts
MarkLark86: I would probably refrain from using the word "corruption", doesn't sound good to an end user

thecalcc: network issue might be better for the case, I put this because if you sometimes reload the page before autosave finishes you might get an item in a bad state

client/components/Main/ItemEditor/Editor.tsx
eos87: fair enough 😅

client/components/editor-standalone/authoring-storage-event-http.ts
eos87: then I'd suggest something like "Could not save event. Please adjust the event times and try again."

thecalcc: Agreed, but what about other cases, if some other data is wrong? I can't know that in advance what is wrong and I can't parse all the potential error messages we might receive from the backend

client/api/editor/item_planning.ts
eos87: the corruption one sounds very worrying, and if it comes from Nicaragua like myself then it's all lost 🤣 

suggestion: ```'Could not link event - Please close the item and try linking again'```

thecalcc:
I changed the comparison logic for `areEmbeddedItemsDirty` because I realized we only care if the links changed, not the changes inside the items data itself. Because of that I also made changes in `client/actions/events/api.ts` -> `save`.

eos87 (approved):
LGTM

PR #2285, opened 2025-04-22
Embedded planning locking

thecalcc:
## Front-end checklist


tomaskikutis:
Could you briefly describe what was the problem and how you solved it?

client/components/fields/editor/AssociatedEventItem.tsx
```diff
@@ -118,15 +118,15 @@ export class AssociatedEventItem extends React.PureComponent<IProps> {
                                     );
 
                                     return eventsApi.events.create(itemClean).then(([created]) =>
-                                        planningApi.locks.unlockEmbeddedEvent(created._id)
-                                            .then((unlocked) => {
-                                                this.update(unlocked);
+                                        planningApi.locks.unlockEmbeddedItem(created, false).then((unlockedItem) => {
```
tomaskikutis, line 121: doesn't `softOnly` default to `false`? no reason to pass it here then

thecalcc:
Problems: 
1. Locks on related plannings were not getting updated properly - once you open an event not all embedded plannings were getting updated
2. On add/remove/add by dropping an embedded planning item wasn't getting locked
3. Other planning items were getting locked once you open some planning item that had events which had linked plannings.

Solutions:

To problem 1 and 2: Lock all associated plannings on mount/change by calling `softLockPlannings` in `client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx`

To problem 3: Remove the last case of `getPlanningLock` in `client/utils/locks.ts`

Note: I also had to change unlocking of items in `removeLock` in order to properly unlock related plannings on close of the main event item.

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
```diff
@@ -20,6 +21,51 @@ export class EditorFieldEventRelatedPlanningsComponent extends React.PureCompone
         super(props);
 
         this.relatedItemRefs = {};
+        this.softLockPlannings = this.softLockPlannings.bind(this);
+        this.unlockPlannings = this.unlockPlannings.bind(this);
+    }
+
+    private softLockPlannings() {
+        planningApi.locks.setItemAsLocked({
```
tomaskikutis: what about using `unlockEmbeddedItem`?

client/components/fields/editor/AssociatedEventItem.tsx
thecalcc: It defaults to `true`, you can see in `client/api/locks.ts:332`

I need to also unlock it on the backend, because `create` locks the created item on the backend. And in the case that the page is reloaded an error is displayed `Failed to unlock item`, because our current implementation of locks doesn't support unlocking related items on the backend.

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
thecalcc: `unlockEmbeddedItem` is responsible for unlocking a single item, here I am locking, not unlocking, but even if I was unlocking using `unlockEmbeddedItem` would only work if I do something like:

```
this.props.item.associated_plannings.forEach((x) => unlockEmbeddedItem(x)
```

client/components/fields/editor/AssociatedEventItem.tsx
tomaskikutis: It's confusing to have it default to true.

Imagine a function `publishArticles(withMediaOnly: boolean)` where `withMediaOnly` would default to true.

Could you either default to `false` or make it mandatory?

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
tomaskikutis: could you also add a function `lockEmbeddedItem` and use it here then? Your example looks good. It doesn't look good to repeat all fields in multiple places.

client/components/fields/editor/AssociatedEventItem.tsx
thecalcc: I can do it yep

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
thecalcc: It's not the correct way in my opinion. Redux dispatch would then get fired `associated_plannings.length - 1` number of times. By using `setItemAsLocked` I send all `associated_plannings` ids in one go using `plan_ids`, and process them at once in `setItemAsLocked` and redux gets called once.

tomaskikutis (approved):

PR #2286, opened 2025-04-23
Content API for Planning: Create new Mongo/Elastic resources for Events & Planning [STT-104]

devketanpro:


server/planning/content_api/events/event.py
```diff
@@ -0,0 +1,33 @@
...
+from superdesk.core.resources import ModelWithVersions
+from planning.types import EventResourceModel
+from superdesk.core.resources import fields
+from superdesk.types.base import CVItem
+from pydantic import Field
+from typing import Any
+
+
+class ContentAPIEventResourceModel(EventResourceModel, ModelWithVersions):
+    headline: fields.HTML | None = None
```
MarkLark86: These fields look like content related fields, where did they come from?

server/planning/content_api/events/event.py
```diff
@@ -0,0 +1,33 @@
...
+    description_text: str | None = None
```
MarkLark86: Missing re-defining fields that will be expanded. Please look at [event formatter](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/output_formatters/json_event.py) and [planning formatter](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/output_formatters/json_planning.py) for an example of the fields we expand on.

At the moment some things are stored with just it's ID. We should instead be expanding on them, just like we do when we publish them to Newshub.

So please go over those formatters thoroughly, and find out what fields are expanded and update this and planning schema to reflect

server/planning/content_api/events/event.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Any
+
+
+class ContentAPIEventResourceModel(EventResourceModel, ModelWithVersions):
+    event_contact_info: List[Dict[str, Any]] = Field(default_factory=list)
+    products: List[Dict[str, str]] = Field(default_factory=list)
+    files: List[Dict[str, Any]] = Field(default_factory=list)
+
+    # Validators to expand fields
+    @validator("event_contact_info", pre=True)
```
MarkLark86: This does not expand on the schema but rather the data. We need the schema to have the appropriate fields defined, so that the Elasticsearch mapping contains these fields so they can be searchable through the ContentAPI.

i.e.
If we add the appropriate fields under `event_contact_info`, then we can search for Event's based on the name of a Contact.

I have defined the valid fields in a [gist](https://gist.github.com/MarkLark86/b565d4de7bfc4b770f143cbfd7d819e8) for you, for the Event resource.

Could you please do the same level of expanding the fields for the Planning item (for the `agendas`, `products`, `events`, `coverages` and `event_item` fields).

Also it would be good to move these models to a `planning.content_api.types` module, similar to how we have it in planning/superdesk. Something like:
image

server/planning/content_api/events/event.py
```diff
@@ -0,0 +1,58 @@
...
+    def expand_contacts(cls, v):
```
MarkLark86: Once the above is done, we don't want these validator/expansion code here, it will belong somewhere else

server/planning/content_api/events/module.py
```diff
@@ -0,0 +1,49 @@
...
+    ResourceConfig,
+    MongoResourceConfig,
+    MongoIndexOptions,
+    ElasticResourceConfig,
+    RestEndpointConfig,
+)
+from content_api import MONGO_PREFIX, ELASTIC_PREFIX
+
+from .event import EventResourceModel
+from planning.events import EventsService
```
MarkLark86: We won't require the full EventsService from the base planning module. Instead we want a generic one without any app specifics (a simple data layer). Something as easy this should do:

```python
from superdesk.core.resources.service import AsyncResourceService

class ContentAPIEventService(AsyncResourceService[ContentAPIEventsResource]):
    pass
```

server/planning/content_api/events/event.py
```diff
@@ -0,0 +1,53 @@
...
+class ContentAPIEventService(AsyncResourceService[ContentAPIEventResource]):
+    pass
+
+
+content_api_event_resource_config: ResourceConfig = ResourceConfig(
+    name="events_capi",
+    data_class=ContentAPIEventResource,
+    service=ContentAPIEventService,
+    default_sort=[("versioncreated", -1)],
+    versioning=True,
```
MarkLark86: This should be `False`, we won't need versions for Events resources

server/planning/content_api/types/planning.py
```diff
@@ -0,0 +1,59 @@
...
+from datetime import datetime
+from planning.types import PlanningResourceModel, PLANNING_EVENT_LINK_METHOD
+
+
+class AgendaItem(Dataclass):
+    qcode: fields.Keyword
+    name: str
+
+
+class MatchingProduct(Dataclass):
```
MarkLark86: This one is already declared in Events type file. It would be good to then place this is a common file to be imported by both Events and Planning type files

server/planning/content_api/types/planning.py
```diff
@@ -0,0 +1,59 @@
+from superdesk.core.resources import fields, Dataclass
+from pydantic import Field
+from superdesk.utc import utcnow
+from datetime import datetime
+from planning.types import PlanningResourceModel, PLANNING_EVENT_LINK_METHOD
+
+
+class AgendaItem(Dataclass):
+    qcode: fields.Keyword
```
MarkLark86: Agenda does't use `qcode` but rather `_id`

server/planning/content_api/types/planning.py
```diff
@@ -0,0 +1,59 @@
...
+    news_coverage_status: dict | None = None
+    original_creator: fields.Keyword | None = None
+
+
+class ContentAPIPlanningResourceModel(PlanningResourceModel):
+    agendas: list[AgendaItem] = Field(default_factory=list)
+    products: list[MatchingProduct] | None = None
+    events: list[RelatedEvent] | None = None
+    coverages: list[PlanningCoverageItem] = Field(default_factory=list)
+    event_item: list[str] | None = None
```
MarkLark86: I believe this is just a string, not a list of strings

server/planning/content_api/types/planning.py
```diff
@@ -0,0 +1,59 @@
...
+    name: str
+
+
+class MatchingProduct(Dataclass):
+    code: fields.Keyword
+    name: str
+
+
+class RelatedEvent(Dataclass):
+    _id: fields.Keyword
```
MarkLark86: The fields here are `rel`, `uri`, `literal` and `name` [see code](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/output_formatters/json_planning.py#L118).

server/planning/content_api/types/planning.py
```diff
@@ -0,0 +1,59 @@
...
+    recurrence_id: str | None = None
+    link_type: PLANNING_EVENT_LINK_METHOD | None = None
+
+
+class CoverageAssignedTo(Dataclass):
```
MarkLark86: This is missing quite a few fields. see [expand_delivery](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/output_formatters/json_planning.py#L162) and [expand_coverage_details](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/output_formatters/json_planning.py#L223) for the extra fields that must be added to the PlanningCoverageItem model

devketanpro: done

MarkLark86 (approved):

PR #2287, opened 2025-04-23
[SDESK-7571] Update the Planning based resources used for publishing, and use the new async publishing system

eos87:
### Purpose
Update the Planning based resources used for publishing, and use the new async publishing system.

### What has changed
- Updated services to use the new `AsyncBaseService` and also updated the methods to be async
- Updated `enqueue_planning_item` to use the new async publish system without the need of celery

Depends on PR: https://github.com/superdesk/superdesk-core/pull/2892

Resolves: [SDESK-7571]

[SDESK-7571]: https://sofab.atlassian.net/browse/SDESK-7571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/common.py
+++ b/server/planning/common.py
@@ -427,7 +427,6 @@ def get_version_item_for_post(item):
     return version, item
 
 
-@celery.task(soft_time_limit=600)
 async def enqueue_planning_item(id):
     """
     Get the version of the item to be published from the planning versions collection and enqueue it.
@@ -443,7 +442,7 @@ async def enqueue_planning_item(id):
             PublishRequest(
                 item=item,
                 item_id=item[ID_FIELD],
-                item_type="event",
+                item_type=item[ITEM_TYPE],
                 operation=ITEM_PUBLISH,
                 published_state=item[ITEM_STATE],
             )
--- a/server/planning/common.py
+++ b/server/planning/common.py
@@ -425,14 +425,14 @@ def get_version_item_for_post(item):
 
 
 @celery.task(soft_time_limit=600)
-def enqueue_planning_item(id):
+async def enqueue_planning_item(id):
     """
     Get the version of the item to be published from the planning versions collection and enqueue it.
 
     :param id:
     :return:
     """
-    planning_version = get_resource_service("published_planning").find_one(req=None, _id=id)
+    planning_version = await get_resource_service("published_planning").find_one_async(req=None, _id=id)
     if planning_version:
         try:
             get_enqueue_service("publish").enqueue_item(planning_version.get("published_item"), "event")
--- a/server/planning/common.py
+++ b/server/planning/common.py
@@ -12,9 +12,15 @@ from typing import NamedTuple, Dict, Any, Set, Optional, Union
 
 import re
 import time
+import json
+from bson import ObjectId
 from collections import namedtuple
+from eve.utils import ParsedRequest
 from datetime import timedelta, datetime
+from werkzeug.datastructures import MultiDict
 
+from superdesk.types import PublishRequest
+from superdesk.publish_async.commands import publish_item
 from superdesk.core import get_app_config, get_current_app
 from superdesk.resource_fields import ID_FIELD, VERSION
 from superdesk.resource import not_analyzed, build_custom_hateoas
@@ -24,13 +30,8 @@ from superdesk.utc import utcnow
 from superdesk.celery_app import celery
 from superdesk.errors import SuperdeskApiError
 from apps.archive.common import get_user, get_auth
-from apps.publish.enqueue import get_enqueue_service
-from .item_lock import LOCK_SESSION, LOCK_ACTION, LOCK_TIME, LOCK_USER
-from eve.utils import ParsedRequest
-from werkzeug.datastructures import MultiDict
+from apps.publish.content.common import ITEM_PUBLISH
 from superdesk.etree import parse_html
-import json
-from bson import ObjectId
 
 from planning.types import (
     Planning,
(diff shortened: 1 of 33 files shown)


server/planning/published_planning.py
```diff
@@ -47,8 +50,8 @@ def _resolve_embedded_item(self, doc, req):
                 fields,
             )
 
-    def get(self, req, lookup):
-        cursor = super().get(req, lookup)
+    async def get_async(self, req, lookup):
+        cursor = await super().get_async(req, lookup)
         if req and req.embedded:
             documents = []
             for doc in cursor:
```
MarkLark86: I think this will need to be `async for doc in cursor` as `get_async` returns an [async cursor](https://github.com/superdesk/superdesk-core/blob/async/superdesk/eve_async/cursors.py)

MarkLark86: Also, would be good to return an async cursor, so usage of this service can expect an async cursor on response of `get_async`.

I have made these changes with my current task, but have placed the cursor file changes in a gist so you can use it now.
https://gist.github.com/MarkLark86/e95b0f6ebc0eb4bd13f18fd61818c453

So instead of returning a `ListCursor` you would return an instance of `AsyncListCursor`

server/planning/common.py
```diff
@@ -425,19 +428,29 @@ def get_version_item_for_post(item):
...
-        try:
-            get_enqueue_service("publish").enqueue_item(planning_version.get("published_item"), "event")
-        except Exception:
-            logger.exception("Failed to queue {} item {}".format(planning_version.get("type"), id))
+        item = planning_version.get("published_item")
+        publish_response = await publish_item(
+            PublishRequest(
+                item=item,
+                item_id=item[ID_FIELD],
+                item_type="event",
```
MarkLark86: I know the old code always sent `event` as the type. It might be good if we send the actual item type with this publish request.

server/planning/events/events_post.py
```diff
@@ -208,20 +208,20 @@ async def post_event(self, event, new_post_state, repost):
...
 
-    def publish_event(self, event, version):
+    async def publish_event(self, event, version):
         # check and remove private contacts while posting event, only public contact will be visible
         event["event_contact_info"] = [try_cast_object_id(contact["_id"]) for contact in get_contacts_from_item(event)]
 
         """Enqueue the items for publish"""
-        # TODO-ASNYC - Change to async when `published_planning` is converted to async
-        version_id = get_resource_service("published_planning").post(
+        # TODO-ASYNC - Change to async when `published_planning` is converted to async
```
MarkLark86: Is this comment needed anymore?

server/planning/events/events_post.py
```diff
@@ -233,14 +233,14 @@ def publish_event(self, event, version):
...
-            enqueue_planning_item.apply_async(kwargs={"id": version_id[0]}, serializer="eve/json")
+            await enqueue_planning_item.apply_async(kwargs={"id": version_id[0]})
         else:
             logger.error("Failed to save planning version for event item id {}".format(event["_id"]))
 
     async def post_related_plannings(self, plannings, new_post_state):
         from planning.planning.planning_spike import process_spike_planning_item
 
-        # TODO-ASNYC - Change to async when `planning_post` is converted to async
+        # TODO-ASYNC - Change to async when `planning_post` is converted to async
```
MarkLark86: Is this comment needed anymore?

server/planning/common.py
```diff
@@ -425,19 +428,29 @@ def get_version_item_for_post(item):
 
 
 @celery.task(soft_time_limit=600)
```
MarkLark86: Now that I think about it, does this need to use celery anymore?
With the old superdesk-core code/content, we changed from using celery task for enqueue to an async function. We might be able to do the same thing here

MarkLark86: If you're able to convert this to async, don't forget the calling convention, currently using `apply_async` because it's a celery task

MarkLark86 (commented):
Looks good. 1 improvement suggested, 1 interface conformance & bug found (in the `PublishedPlanningService` service)

server/planning/events/events_post.py
eos87: I left it because I thought we're still going to migrate these to full async services (with pydantic and all the toys) in the future but we can remove it if that's not the case.

server/planning/events/events_post.py
eos87: same as above

server/planning/published_planning.py
eos87: Gotcha. I've opened a PR just to include the changes in the cursors.py file https://github.com/superdesk/superdesk-core/pull/2892

server/planning/common.py
eos87: done

BrianMwangi21 (commented):
Looks good. Just the pending queries on the TODO comments from Mark.

MarkLark86 (approved):

MarkLark86:
@eos87 Is this PR still in draft, or can we move it to review (if that's the case, it can now be merged).

MarkLark86:
@eos87 I updated this PR, based on the changes in superdesk-core for publishing (the removal of the old code), and some small fixes.
I tested this locally and I am able to publish both Event & Planning items into Newshub (checked both Superdesk/Workqueue and Newshub/Agenda).

I'll merge this. If there are further fixes you found, we can open a new PR

PR #2292, opened 2025-04-29
Custom vocabularies as separate fields

thecalcc:
STT-133

## Front-end checklist


thecalcc:
hi @petrjasek can you please take a look at backend changes

client/components/ContentProfiles/FieldTab/AddFieldsMenu.tsx
```diff
@@ -20,7 +22,7 @@ export default class AddFieldsMenu extends React.PureComponent<IProps, any> {
                 getId={(field) => field.name}
                 optionTemplate={(item) => item.schema?.type === 'custom_vocabulary' ? (
                     <>
-                        {getFieldNameTranslated(item.name)}
+                        {this.props.allCVs.find((x) => x._id === item.name).display_name}
```
tomaskikutis: Don't iterate instead `optionTemplate`/`getLabel` functions. Make a dictionary/map at the top of the render method and access items by ID later.

client/components/ContentProfiles/FieldTab/FieldEditor.tsx
```diff
@@ -80,6 +80,7 @@ export class FieldEditor extends React.Component<IProps, IState> {
     }
 
     render() {
+        const allCVs = planningApi.vocabularies.getCustomVocabularies();
```
tomaskikutis: In superdesk `CV` stands for controlled vocabulary. Don't use that abbreviation for anything else, especially also having to do with vocabularies.

client/planning-extension/src/extension.ts
```diff
@@ -269,6 +272,28 @@ const extension: IExtension = {
...
         };
 
+        // Do not register coverage related CVs as fields, because it overrides current implementation for these fields.
+        // Do not register CVs that aren't considered "custom CVs" also,
+        // because if some are already registered as specific fields they will be overridden.
+        // Definition of a custom CV: client/services/PlanningStoreService.ts:255
+        const customCvs = superdesk.entities.vocabulary
+            .getAll()
+            .toArray()
+            .filter((cv) => !COVERAGE_CVS.includes(cv._id) && !isEmpty(cv.service) && isEmpty(cv.field_type));
```
tomaskikutis: is there a function somewhere `isCustomVocabulary`? If not add it

server/planning/content_profiles/profiles/event.py
```diff
@@ -43,7 +43,6 @@ class EventSchema(BaseSchema):
     reference = StringField()
     slugline = StringField()
     subject = subjectField
-    custom_vocabularies = schema.ListField()
```
petrjasek: I would keep it in the schema for now so we can later do the migration, without it this data won't be returned from db

petrjasek, line 46: I would keep it in the schema for now so we can later do the migration, without it this data won't be returned from db

client/utils/contentProfiles.ts
```diff
@@ -81,10 +82,57 @@ export function getUnusedProfileFields(
...
     includeGroupCheck: boolean = true
 ): Array<IProfileFieldEntry> {
+    // Temporary FIXME: Will be removed once we add support for CVs as fields in coverage profiles
+    if (profile.name === 'coverage') {
+        const fieldsFromProfile = getProfileFields(profile);
+
+        return orderBy(
+            fieldsFromProfile
+                .filter((field) => (includeGroupCheck && field.field.group == null) || !field.field.enabled),
+            'name'
```
tomaskikutis: type as interface property

client/utils/contentProfiles.ts
```diff
@@ -81,10 +82,57 @@ export function getUnusedProfileFields(
...
+                enabled: false,
+                group: undefined,
+                index: i,
+            },
+            name: x._id,
+            schema: {
+                type: 'custom_vocabulary',
+                required: false,
+            }
+        }) as IProfileFieldEntry);
```
tomaskikutis: It looks like you're casting when you don't need to. Either use `const result: IProfileFieldEntry = ...` or `satisfies` keyword.

tomaskikutis: It looks like you're casting when you don't need to. Either use `const result: IProfileFieldEntry = ...` or `satisfies` keyword.

client/utils/contentProfiles.ts
```diff
@@ -81,10 +82,57 @@ export function getUnusedProfileFields(
...
+    });
+
     return orderBy(
-        getProfileFields(profile).filter(
-            (item) => (includeGroupCheck && item.field.group == null) || !item.field.enabled
-        ),
+        fieldsFromProfile
+            .filter((field) => (includeGroupCheck && field.field.group == null) || !field.field.enabled)
+            .filter((field) => {
+                const fieldSchema = field.schema;
```
tomaskikutis: Variable is only used once and naming is not any better than value. Why would you need it?

It would be good to also use a single `filter`

client/api/editor/form.ts
```diff
@@ -19,7 +19,7 @@ export function getFormInstance(type: EDITOR_TYPE): IEditorAPI['form'] {
     }
 
     function getState() {
-        return planningApi.editor(type).manager.getState();
+        return planningApi.editor(type).manager?.getState?.();
```
tomaskikutis: can optional chaining be avoided here?

server/planning/content_profiles/profiles/event.py
thecalcc: ok will do, thought we'd migrate them before the release

client/api/editor/form.ts
thecalcc: no, there's no way for me to access the `manager` object beforehand through the editorAPI

client/planning-extension/src/extension.ts
thecalcc: there isn't one, and here it won't be of much benefit to reuse since I also exclude `COVERAGE_CVS` not just do `!isEmpty(cv.service) && isEmpty(cv.field_type))`, I left a comment to clarify everything

tomaskikutis: I'd still be useful to have it  `isCustomVocabulary() && !COVERAGE_CVS.includes(cv._id)`. Otherwise reader is left guessing the purpose of cv.service and cv.field_type check

client/utils/contentProfiles.ts
```diff
@@ -81,10 +82,52 @@ export function getUnusedProfileFields(
...
+        && !usedVocabularies.some((usedFieldEntry) => usedFieldEntry.name === field.name)
+    );
+
     return orderBy(
-        getProfileFields(profile).filter(
-            (item) => (includeGroupCheck && item.field.group == null) || !item.field.enabled
-        ),
+        fieldsFromProfile
+            .filter((field) => (
+                ((includeGroupCheck && field.field.group == null) || !field.field.enabled) &&
```
tomaskikutis: use functions/variables/comments to make that filter more understandable to read

server/planning/content_profiles/profiles/event.py
petrjasek: yep we might, but it might be needed for the migration script

client/api/editor/form.ts
tomaskikutis: expose a method instead of check what you need to check in order to avoid optional chaining

tomaskikutis (approved):

PR #2299, opened 2025-05-06
[SDESK-7610] - Articles based on the scheduled exports are not being created in the desks at all

BrianMwangi21:
### Purpose
This PR fixes a bug that cause scheduled export articles to not be created in the assigned desks. 

## Changes made
1. Converted search results from EventsPlanningFiltersAsyncService into a list of dicts to remove `TypeError: 'EventPlanningFilter' object is not subscriptable` errors
2. Added awaits to `render_template` and `render_template_string` as they were returning a coroutine in the `body_html` attribute which couldn't be deep copied at a later step.

## Steps to reproduce
1. Go to Planning > Manage events and planning filters
2. Create a few planning filters
3. Create scheduled exports for the planning filters

Observe: Articles are now created in the desks according to the schedules.

Solves: [SDESK-7610](https://sofab.atlassian.net/browse/SDESK-7610)

[SDESK-7610]: https://sofab.atlassian.net/browse/SDESK-7610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/planning_article_export.py
```diff
@@ -68,7 +68,7 @@ async def get_items(ids, resource_type):
     items = await get_resource_service("events_planning_search").search_repos(
         resource_type, {"item_ids": ",".join(ids_string), "only_future": False}
     )
-    items = sorted([item async for item in items], key=lambda i: ids_string.index(str(i.get("_id"))))
+    items = sorted([item for item in items], key=lambda i: ids_string.index(str(i.get("_id"))))
```
MarkLark86: Doesn't search_repos return a [get_async](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/search/eventsplanning_search.py#L252) which returns an instance of AsyncEveCursor, which is an async iterable.

BrianMwangi21: Hi @MarkLark86 , 

It actually returns an EveCursor if you follow the chain of calls. Here's a breakdown. 

`search_repos` returns `self.get_async` which is inherited and has several returns. 

The returns include await `self._search_events`, `self._search_planning` and `self._get_events_and_planning`. 

For all the 3, they call `get_async` on PlanningSearch service. 
This is the one that eventually returns an EveCursor cause it returns the docs as is.

MarkLark86: Ah, that is where the issue stems from. The [get_async](https://github.com/superdesk/superdesk-core/blob/async/superdesk/eve_async/service.py#L106) function is suppose to return an instance of AsyncEveCursor, therefor the [implementation in planning_search](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/search/planning_search.py#L116) service is invalid.

Whenever we use `get_async`, we should always assume we're receiving an async iterator, makes the usage consistent that way.
Have a look at the [content search](https://github.com/superdesk/superdesk-core/blob/async/apps/search/__init__.py#L252) service that implements the same but for content.

This should fix that async cursor issue there and be consistent with the interface of the base service

MarkLark86 (approved):

PR #2301, opened 2025-05-08
Fix: Convert locks to async, EventPlanningFilter and other async fixes

MarkLark86:
### Purpose
To get the E2E tests running, and begin the process of fixing those tests

### What has changed:
* Convert Event, Planning & Assignment locks to async
* Create E2E fixtures to setup publishing of Planning based resources
* Update planning_prepopulate to use test utils from core (so it uses both Eve & Pydantic based services)
* Fix setting XMP and getting archive items in the assignment service
* Fix EventPlanningFilter resource model - Defaulting to empty list doesn't always work, sometimes the front-end send `null` (we had the same issue in Newshub). Ideally this would be fixed in the front-end, but we're limiting the async changes to the back-end

Depends on https://github.com/superdesk/superdesk-core/pull/2916

### DIFF
--- a/e2e/cypress/e2e/planning/edit_planning.cy.ts
+++ b/e2e/cypress/e2e/planning/edit_planning.cy.ts
@@ -46,8 +46,7 @@ describe('Planning.Planning: edit metadata', () => {
 
         list.expectEmpty();
         editor.expectItemType();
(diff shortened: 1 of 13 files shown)


eos87 (approved):
Looks good

PR #2302, opened 2025-05-08
Fix planning autosave returning not temp IDs after temp ID has been sent

thecalcc:
SDESK-7628

## Front-end checklist

### DIFF
--- a/server/planning/types/autosave.py
+++ b/server/planning/types/autosave.py
@@ -1,5 +1,7 @@
 from .event import EventResourceModel
-from .planning import PlanningResourceModel
+from .planning import PlanningResourceModel, PlanningCoverage
+from pydantic import model_validator
+from typing import Any
 
 
 class EventAutosaveResourceModel(EventResourceModel):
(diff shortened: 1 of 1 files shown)


thecalcc:
@MarkLark86 what happened is that once we send a POST to planning_autosave with payload - an item with a `tempId` prefix, we get back an item with full `guid` ID. 

Then the check to get newly autosaved and locked items fails on frontend, causing the item not to show up in the workqueue - https://github.com/superdesk/superdesk-planning/blob/ce46a127f773f14c4acdb48771d535a83da02d90/client/selectors/forms.ts#L105

I checked the backend and saw that this validator function causes the issue, and thought it's ok to override it on the autosave model. Let me know if there is a better fix. 

PS: Also saw that because tempId is gone after the POST, the logic to remove items from the autosave also fails, thus we could end up getting stuck with quite a bunch of items on an instance. 

https://github.com/user-attachments/assets/bd51f58c-b4ec-441d-b58c-20d4212dcaa8

MarkLark86:
Thanks @thecalcc Good find. I updated my [PR](https://github.com/superdesk/superdesk-planning/pull/2301) with the [change](https://github.com/superdesk/superdesk-planning/pull/2301/commits/e0cb78f46c155349ab4166a98f42a27b56153621) to remove most of the validation on the autosave models, as I don't see the need for validating every field on autosave (we reserve that for the actual save action).

I'll close this one now.

PR #2303, opened 2025-05-09
Content API for Planning: Add API endpoints for Events & Planning (applying subscriber permission filters) [STT-106]

devketanpro:
Depends on: https://github.com/superdesk/superdesk-core/pull/2947


MarkLark86:
It would be good to have some tests for both get one and get many

server/planning/content_api/events/__init__.py
```diff
@@ -8,7 +8,47 @@
...
 
-module = Module("planning.content_api.events", resources=[content_api_event_resource_config])
+event_endpoints = EndpointGroup("events_capi", __name__)
+
+
+@event_endpoints.endpoint("event", methods=["GET"])
+async def get_event_list(args, params, request: Request) -> Response:
+    service = ContentAPIEventService()
+    req = ParsedRequest()
+    req.args = MultiDict()
```
MarkLark86: It would be good if we could add support here for the allowed params, at the moment none are passed through from this endpoint to the underlying service

server/planning/content_api/events/__init__.py
```diff
@@ -8,23 +8,39 @@
...
 from superdesk.core.web import EndpointGroup
 from werkzeug.datastructures import MultiDict
 from superdesk.core.types import Request, Response
 from .event import content_api_event_resource_config, ContentAPIEventService
+from planning.content_api.utils import ALLOWED_PARAMS
 
 event_endpoints = EndpointGroup("events_capi", __name__)
 
 
+class EventListParams(BaseModel):
```
MarkLark86: As these are common fields across both Events and Planning endpoints, it would be good to define these in common types (or at least a common base class with fields that are used with both endpoints).

Also, no need for `Optional[str] | None`, Python now supports this.
You can instead use `str | None = None`

server/planning/content_api/events/__init__.py
```diff
@@ -8,23 +8,39 @@
...
+
 @event_endpoints.endpoint("event", methods=["GET"])
-async def get_event_list(args, params, request: Request) -> Response:
+async def get_event_list(args: None, params: EventListParams, request: Request) -> Response:
     service = ContentAPIEventService()
     req = ParsedRequest()
-    req.args = MultiDict()
-
+    req.args = MultiDict(
+        {param: getattr(params, param) for param in ALLOWED_PARAMS if getattr(params, param) is not None}
```
MarkLark86: You need to provide a default for `getattr` otherwise it will raise an AttributeError

server/planning/module.py
```diff
@@ -79,7 +81,14 @@ def init_planning(app: SuperdeskAsyncApp):
 module = Module(
     "planning",
     init=init_planning,
-    endpoints=[planning_locks_endpoint, planning_endpoint_group, events_endpoints_group, planning_download_endpoint],
+    endpoints=[
```
MarkLark86: I think these are playing up with the Web API for the app itself. The ContentAPI is meant to be run as a separate App to the Web API.
E2E tests are failing (which they're currently passing in async branch), and various behave tests are failing as well. Please have a look, and maybe need to run a separate test environment for the Content API

MarkLark86: Only the resources themselves are needed in the App config (so the publish into it works). The endpoints aren't needed

server/planning/content_api/events/__init__.py
```diff
@@ -8,7 +8,46 @@
...
+from superdesk.flask import g
+from pydantic import BaseModel
+from eve.utils import ParsedRequest
 from superdesk.core.module import Module
-from .event import content_api_event_resource_config
+from superdesk.core.web import EndpointGroup
+from werkzeug.datastructures import MultiDict
+from superdesk.core.types import Request, Response
+from .event import content_api_event_resource_config, ContentAPIEventService
+from planning.content_api.utils import ALLOWED_PARAMS, APIListParams
```
MarkLark86: APIListParams is not that descriptive. How about `PlanningCAPIParams` or something like that

server/planning/content_api/events/__init__.py
```diff
@@ -8,7 +8,46 @@
...
 
-module = Module("planning.content_api.events", resources=[content_api_event_resource_config])
+event_endpoints = EndpointGroup("events_capi", __name__)
+
+
+@event_endpoints.endpoint("event", methods=["GET"])
+async def get_event_list(args: None, params: APIListParams, request: Request) -> Response:
+    service = ContentAPIEventService()
+    req = ParsedRequest()
```
MarkLark86: Why use Eve? We have [SearchRequest](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/types/search.py#L137) type in the new code, as well as using the function [params directly](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/resources/service.py#L699).

This way you won't need Eve code (try to avoid Eve at all cost, we will eventually remove it), and you won't need custom code in the ResourceService to handle these queries

server/planning/module.py
devketanpro: but if we remove the endpoints config from here, we can't used it in the content_api to get the data

devketanpro: afaik, we need to define the endpoint group in the module config, right?

MarkLark86: We need to define these endpoint config in the ContenteAPI planning config, not the one for the main Web API app

devketanpro: Got it, so I need to create a separate module for the planning`content_api` right ?

server/planning/content_api/events/__init__.py
```diff
@@ -8,46 +8,57 @@
...
+
+    lookup = {}
+    user = g.get("user")
+    if user and "_id" in user:
+        lookup["subscribers"] = user["_id"]
+
+    where = dict(params.where) if isinstance(params.where, dict) else {}
+    where.update(lookup)
+
+    search_request = SearchRequest(
```
MarkLark86: Please have a look at the available params in [SearchRequest](https://github.com/superdesk/superdesk-core/blob/async/superdesk/core/types/search.py#L137), the instance here is invalid.

server/settings.py
```diff
@@ -69,7 +69,7 @@ def env(variable, fallback_value=None):
 
 INSTALLED_APPS = ["planning"]
 
-MODULES.append("planning")
+MODULES.extend(["planning", "planning.content_api"])
```
MarkLark86: I believe this is used by the Web API app, that is why you're getting `already registered error` in `server.planning.module.resources`

MarkLark86: This is where you'll need a separate config for the ContentAPI (this new config can be for tests only)

server/planning/content_api/events/event.py
```diff
@@ -63,10 +56,5 @@ async def publish_async(self, item, subscribers=None) -> None:
             MongoIndexOptions(name="template", keys=[("template", 1)], unique=False),
         ],
     ),
-    elastic=ElasticResourceConfig(prefix=ELASTIC_PREFIX),
-    rest_endpoints=RestEndpointConfig(
-        resource_methods=["GET"],
-        item_methods=["GET"],
-        enable_cors=True,
-    ),
+    elastic=ElasticResourceConfig(),
```
MarkLark86: `ELASTIC_PREFIX` is required, so that the data is stored separately to the rest of Superdesk

MarkLark86:
@devketanpro There are flake8 issues found, also please add `aioresponses` to the server/requirements.txt (required from superdesk-core side).

Once these issues are fixed I think this once is good to merge. Let's see how the behave tests go (do they pass locally for the ContentAPI?)

MarkLark86:
@devketanpro You'll need to exclude the `features/planning_content_api` folder from the first behave run (as it's being picked up there as well). Add `--exclude features/planning_content_api` to the [behave cmd line](https://github.com/devketanpro/superdesk-planning/blob/stt-106/.github/workflows/ci-server.yml#L49).

Also I tried running the behave test for the ContentAPI and they're failing

devketanpro:
> --exclude features/planning_content_api

I'll update the behave command with `--exclude features/planning_content_api.`

as for the Content API tests — I’ve been looking into it, and honestly, not fully sure what the issue is yet. I’ve implemented the config and setup as expected, so it should be working. still digging into it and will update once I figure out what’s going wrong.

MarkLark86:
@devketanpro I have updated the PR to get the behave tests working. I have fixed a few issues, as the API and publishing to it was not working.

I focused mostly on the Planning endpoint and tests for it.
Can you please take over and update the Event tests, and also add tests for search params for both Event & Planning endpoints (as I fear the filters and projections may not work)

MarkLark86 (approved):

PR #2304, opened 2025-05-09
Async fix tests 2

MarkLark86:
### Purpose
Main purpose was to get E2E passing (which it does locally). Then continue on fixing various async issues and unit tests. Next stage is to get all unit tests and behave tests passing.

### DIFF
--- a/server/planning/agendas_async/module.py
+++ b/server/planning/agendas_async/module.py
@@ -10,17 +10,17 @@ agendas_resource_config: ResourceConfig = ResourceConfig(
     name="agenda",
     data_class=AgendasResourceModel,
     service=AgendasAsyncService,
-    rest_endpoints=RestEndpointConfig(
-        resource_methods=["GET", "POST"],
-        item_methods=["GET", "PATCH", "DELETE"],
-        enable_cors=True,
-        endpoints_class=AgendasRestEndpoints,
-        auth=http_method_privilege_based_rules(
-            {
-                "POST": "planning_agenda_management",
-                "PATCH": "planning_agenda_management",
-                "DELETE": "planning_agenda_management",
-            }
-        ),
-    ),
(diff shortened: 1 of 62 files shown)


server/planning/agendas_async/module.py
```diff
@@ -10,17 +10,17 @@
...
+    #     enable_cors=True,
+    #     endpoints_class=AgendasRestEndpoints,
+    #     # auth=http_method_privilege_based_rules(
+    #     #     {
+    #     #         "POST": "planning_agenda_management",
+    #     #         "PATCH": "planning_agenda_management",
+    #     #         "DELETE": "planning_agenda_management",
+    #     #     }
+    #     # ),
+    # ),
```
BrianMwangi21, line 25: Just a question for my understanding, why is AgendasAsyncService rest api being disabled and returning back to the old Service. Is it for fixing the tests ?

BrianMwangi21 (approved):
LGTM.

Just curious why like in the commands you reverted back to the old services and not the new pydantic async ones. Test related ?

server/planning/agendas_async/module.py
eos87: Yeah, I think it's for fixing the reason.

eos87 (approved):

server/planning/agendas_async/module.py
MarkLark86: Correct. I don't know why yet, but it was quicker to revert back to eve async service.
@BrianMwangi21 Same goes with your previous comment, as front-end is using eve services, I am preferring these eve services over Pydantic to get not just the tests passing but also the app working.

With getting everything async, it will be easier in the future to upgrade to Pydantic service. Maybe we could start looking at this in a v3.1 version

PR #2305, opened 2025-05-09
Fix coverage fields sorting

thecalcc:
STT-1183

## Front-end checklist

## Fix:

After moving a field to a new position its new index was being overwritten

### DIFF
--- a/client/components/ContentProfiles/FieldTab/index.tsx
+++ b/client/components/ContentProfiles/FieldTab/index.tsx
@@ -154,32 +154,33 @@ export class FieldTab extends React.Component<IProps, IState> {
     }
 
     updateFieldOrder(fields: Array<IProfileFieldEntry>) {
-        fields.forEach((item, index) => {
-            item.field.index = index;
-        });
-
         this.props.updateFields(fields);
     }
 
-    insertField(itemToAdd: IProfileFieldEntry, groupId: IEditorProfileGroup['_id'] | undefined, index: number) {
+    insertField(
+        fieldToAdd: IProfileFieldEntry,
+        groupId: IEditorProfileGroup['_id'] | undefined,
+        index: number,
+    ) {
         const fields = this.props.groupFields ?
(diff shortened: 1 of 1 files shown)


client/components/ContentProfiles/FieldTab/index.tsx
```diff
@@ -154,32 +154,33 @@ export class FieldTab extends React.Component<IProps, IState> {
...
             getEnabledProfileFields(this.props.profile);
 
-        fields.push({
-            ...itemToAdd,
-            field: {
-                ...itemToAdd.field,
-                enabled: true,
-                group: groupId,
-                index: index,
+        fields.splice(
```
tomaskikutis: what are you trying to do, replace first record?

thecalcc: insert the item I get at `index` position

tomaskikutis: isn't that the same as `array[index] = newValue`?

thecalcc: no, this also shifts the rest of the elements in this array. I know it doesn't make any sense, but after this `this.props.updateFields(fields);` code runs, also `index` property is rewritten for each element, making it match it's position in the array. And this `index` property is what orders items when we get them from the backend

this is the code for `this.props.updateFields`:

```
updateFields(fields: Array<IProfileFieldEntry>) {
      const profileCloned = cloneDeep(this.state.profile);

          profileCloned.editor[item.name] = {...item.field};
          profileCloned.schema[item.name] = {...item.schema};
          profileCloned.editor[item.name].index = index;
      });

      this.setState({
          profile: profileCloned,
          dirty: true,
      });
}
 ```

thecalcc: and `array[index] = newValue` would override the existing element at `index` position

tomaskikutis (approved):

PR #2306, opened 2025-05-09
Respect no time config

thecalcc:
STT-1187

## Front-end checklist

### DIFF
--- a/client/components/editor-standalone/field-definitions/planning-date.ts
+++ b/client/components/editor-standalone/field-definitions/planning-date.ts
@@ -17,7 +17,10 @@ export const getPlanningDate = (): IFieldDefinition => {
             const field: IAuthoringFieldV2 = {
                 id: id,
                 name: gettext('Planning Date'),
+
+                // time picker must not be displayed for all day events
                 fieldType: appConfig.planning.all_day ? 'date' : 'datetime-v2',
+
                 fieldConfig: {
                     ...config,
                     required: required,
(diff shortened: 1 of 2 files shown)


client/components/editor-standalone/field-definitions/planning-date.ts
```diff
@@ -16,7 +17,7 @@ export const getPlanningDate = (): IFieldDefinition => {
             const field: IAuthoringFieldV2 = {
                 id: id,
                 name: gettext('Planning Date'),
-                fieldType: 'datetime-v2',
+                fieldType: appConfig.planning.all_day ? 'date' : 'datetime-v2',
```
tomaskikutis, line 22: why would field type matter depending on all day config?

thecalcc: this is just named all day, in reality it means "no time". We want to hide the timepicker, and I don't want to couple the `datetime-v2` field with something like: `dateOnly: boolean;`

client/globals.d.ts
```diff
@@ -186,11 +186,13 @@ declare module 'superdesk-api' {
             autosave_timeout?: number;
             default_create_planning_series_with_event_series?: boolean;
             event_related_item_search_provider_name?: string;
+
+            // Controls whether date only input is rendered in the editor
```
tomaskikutis: which editor?

thecalcc: all should be synced, which they are now, so for all editor views - embedded and main one, maybe tomorrow we add a new one

thecalcc: I realize this flag can control anything, depending on how it's used, so I made it agnostic of editors or anything else

tomaskikutis (approved):

PR #2307, opened 2025-05-11
[STT-105] Implement publish of planning and event items into content api

eos87:
Depends on https://github.com/superdesk/superdesk-core/pull/2920

[STT-105]

[STT-105]: https://sofab.atlassian.net/browse/STT-105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/content_api/events/event.py
+++ b/server/planning/content_api/events/event.py
@@ -15,21 +15,32 @@ from superdesk.core.resources import (
     ElasticResourceConfig,
     RestEndpointConfig,
 )
-from content_api import MONGO_PREFIX, ELASTIC_PREFIX
-from planning.content_api.types.events import ContentAPIEventResource
 from superdesk.core.resources.service import AsyncResourceService
+
+from content_api import MONGO_PREFIX, ELASTIC_PREFIX
+from planning.output_formatters import JsonEventFormatter
+from planning.content_api.types.events import ContentAPIEventResource
 
 
 class ContentAPIEventService(AsyncResourceService[ContentAPIEventResource]):
+    """Service for publishing events to the content API"""
+
+    formatter = JsonEventFormatter()
+
     async def publish_async(self, item, subscribers=None) -> None:
+        """
+        Uses the `JsonEventFormatter` to format the event and publish it to the content API.
+        If the event already exists, it will be updated, otherwise it will be created.
+        """
+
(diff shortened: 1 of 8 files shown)


server/planning/content_api/events/event.py
```diff
@@ -21,7 +21,15 @@
 
 
 class ContentAPIEventService(AsyncResourceService[ContentAPIEventResource]):
-    pass
+    async def publish_async(self, item, subscribers=None) -> None:
```
MarkLark86, line 30: Won't the item passed in would still be in the original internal Event & Planning format?
I believe we may have to format the item based on [json_event](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/output_formatters/json_event.py) and [json_planning](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/output_formatters/json_planning.py) to convert the item into the correct ContentAPI format, similar to what we do with [content formatting](https://github.com/superdesk/superdesk-core/blob/async/content_api/publish/service.py#L81)

eos87: yeah, they are but in the end both content_api planning and event models inherit from both "internal" event & planning models therefore, they share all the same fields so I was not sure which fields should be preserved and which aren't.

eos87: @MarkLark86 I've implemented the changes to format the items before creating/updating them. Please take another look.

MarkLark86 (approved):
Looks good

server/planning/content_api/events/event.py
```diff
@@ -15,13 +15,32 @@
...
+
+    formatter = JsonEventFormatter()
+
+    async def publish_async(self, item, subscribers=None) -> None:
+        """
+        Uses the `JsonEventFormatter` to format the event and publish it to the content API.
+        If the event already exists, it will be updated, otherwise it will be created.
+        """
+
+        formatted_item = await self.formatter._format_item(item)
```
devketanpro, line 36: It would be beneficial to add error handling and logging to this section to ensure that any unhandled exceptions are properly caught. thoughts ?

eos87: exceptions are being handled from outside in the consumer so it should be fine.

PR #2309, opened 2025-05-13
Assignment Id in preview

thecalcc:
STT-122

## Front-end checklist

### DIFF
--- a/client/components/AbsoluteDate/index.tsx
+++ b/client/components/AbsoluteDate/index.tsx
@@ -1,8 +1,15 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import moment from 'moment';
 import {DATE_FORMATS} from '../../constants';
 import {TextInput} from '../UI/Form';
+
+interface AbsoluteDateProps extends React.HTMLAttributes<HTMLElement> {
+    date?: string;
+    noDateString?: string;
+    className?: string;
+    asTextInput?: boolean;
+    toBeConfirmed?: boolean;
+}
 
 /**
  * Display absolute date in <time> element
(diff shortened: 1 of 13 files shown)


client/components/PriorityLabel/index.tsx
```diff
@@ -16,9 +17,9 @@ interface IProps {
             name?: {[key: string]: string};
         };
     }>;
-    tooltipFlow: 'up' | 'right' | 'down' | 'left';
+    tooltipFlow: 'top' | 'right' | 'down' | 'left';
```
thecalcc, line 20: I have checked usages here

thecalcc:
This is how the UI looks, major addition is the new row with assignment Id
image

client/components/Assignments/AssignmentPreviewContainer/AssignmentPreviewHeader.tsx
```diff
@@ -70,105 +71,109 @@ export const AssignmentPreviewHeader = ({
...
-                                            {gettext('Assigned by {{name}}', {name: deskAssignorName})},
+                            <UserAvatar
+                                user={assignedUser}
+                                size="large"
+                            />
+                        </Column>
+                    )}
+                    <Spacer v gap="4" noWrap>
+                        <Spacer gap="4" h justifyContent="start" noWrap>
+                            {gettext('Id: {{ assignmentId}}', {assignmentId: assignment._id})}
```
tomaskikutis: 1. there's an extra space before `assignmentId`
2. use a generic placeholder like `{{id}}` or `{{value}}` so you don't generate a new translation string for every field.

client/components/AbsoluteDate/index.tsx
```diff
@@ -1,9 +1,16 @@
...
-import PropTypes from 'prop-types';
 import moment from 'moment';
 import {DATE_FORMATS} from '../../constants';
 import {TextInput} from '../UI/Form';
 
+interface AbsoluteDateProps extends React.HTMLAttributes<HTMLElement> {
+    date?: string;
+    noDateString?: string;
+    className?: string;
+    asTextInput?: boolean;
```
tomaskikutis, line 17: would be good to comment what is this doing

tomaskikutis (approved):

PR #2310, opened 2025-05-13
Async fix tests 3

MarkLark86:
### Purpose
Almost have all the unit tests passing, fix a lot of feature tests. Migrated from using Pydantic Event/Planning resource service to Eve ones, as there were numerous bugs caused by both the model itself and the service code. It was easier to convert back to eve code (as it is now async as well).

At a later date we can revisit these models/services and fix them, but figured it's best to get the code working and tests passing, which would make it easier later to migrate to Pydantic if we have all tests passing now.

Depends on https://github.com/superdesk/superdesk-core/pull/2924

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -69,7 +69,7 @@ LDAP_USER_ATTRIBUTES = json.loads(
 
 INSTALLED_APPS = ["planning"]
 
-MODULES.extend(["planning", "planning.content_api.events", "planning.content_api.planning"])
+MODULES.extend(["planning"])
 
 RENDITIONS = {
     "picture": {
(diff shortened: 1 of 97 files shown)


BrianMwangi21 (approved):

server/settings.py
```diff
@@ -69,7 +69,7 @@ def env(variable, fallback_value=None):
 
 INSTALLED_APPS = ["planning"]
 
-MODULES.extend(["planning", "planning.content_api.events", "planning.content_api.planning"])
+MODULES.extend(["planning"])
```
eos87, line 72: Is this settings.py file really used anywhere? I couldn't find a reference to it the other day.

eos87 (approved):
Looks good. Definitely there was no way we could meet the deadline by going full pydantic + async services, but that's clearer now ;) Good call.

PR #2318, opened 2025-05-19
scheme not required

thecalcc:
SDESK-7640

### Purpose
With the introduction of custom vocabularies, subject schema had to be changed, since entries for all cv fields are stored in subject field. When getting them from the backend we diff based on `scheme` - this is where vocabulary ID is stored for each cv field. Except for subject itself - `scheme` for it is null, we just store vocabulary items without it there. 

### What has changed
Make `scheme` not required, due to `subject` field failing to save.

### Steps to test
Add `subject` field to any coverage type profile. Make a planning/event item, add a coverage, add a couple items in subject, save the item. Saving should work, and UI should not be getting stuck with endless loading. 

### Front-end checklist

Resolves: SDESK-7640


petrjasek (approved):
it should be also possible to set it `null` and keep it required, not sure what's easier to work with on client side

PR #2320, opened 2025-05-22
Accreditation should not be automatically filled in

dzonidoo:
SDBELGA-859

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/components/fields/editor/base/dateTimeUIFramework.tsx
+++ b/client/components/fields/editor/base/dateTimeUIFramework.tsx
@@ -4,6 +4,7 @@ import {IEditorFieldProps} from '../../../../interfaces';
 import {get} from 'lodash';
 import {DateTimePicker} from 'superdesk-ui-framework/react';
 import {appConfig} from 'appConfig';
+import {format} from 'date-fns';
 
 interface IProps extends IEditorFieldProps {
     canClear?: boolean;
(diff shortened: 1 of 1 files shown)


client/components/fields/editor/base/dateTimeUIFramework.tsx
```diff
@@ -55,10 +57,17 @@ export class EditorFieldDateTimeUIFramework extends React.PureComponent<IProps>
                     error={error}
                     disabled={this.props.disabled}
                     dateFormat={appConfig.planning.dateformat}
-                    value={value == undefined ? null : new Date(value)}
+                    value={
+                        value == undefined
```
tomaskikutis: should be `value != null` or `typeof === 'undefined'` if it's important if it's `null` or `undefined`

client/components/fields/editor/base/dateTimeUIFramework.tsx
```diff
@@ -55,10 +57,17 @@ export class EditorFieldDateTimeUIFramework extends React.PureComponent<IProps>
...
+                            ? null
+                            : {
+                                date: format(new Date(value), 'yyyy-MM-dd'),
+                                time: format(new Date(value), 'HH:mm'),
+                            }
+                    }
                     required={this.props.schema?.required}
                     onChange={(value) => {
-                        this.onChange(this.props.field, value);
+                        this.onChange(this.props.field, new Date(`${value.date} ${value.time}`));
```
tomaskikutis: Time is optional, so if it's not set - you'll get invalid date. If I'm not wrong, `onChange` accepts ISO string of a date, not date itself.

tomaskikutis (approved):

PR #2321, opened 2025-05-23
[SDESK-7497] - Fix planning behave test

BrianMwangi21:
### Purpose
This PR fixes some of the breaking behave tests.

NOTE: Not all tests have been resolved.

Resolves: #[SDESK-7497]

[SDESK-7497]: https://sofab.atlassian.net/browse/SDESK-7497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/assignments/assignments.py
```diff
@@ -117,7 +117,7 @@ async def on_fetched_item_async(self, doc):
 
     async def _enhance_assignments(self, docs):
         """Populate `item_ids` with ids for all linked Archive items for an Assignment"""
-        items = await (await self.get_archive_items_for_assignments([str(doc.get(ID_FIELD)) for doc in docs])).to_list()
+        items = await self.get_archive_items_for_assignments([str(doc.get(ID_FIELD)) for doc in docs])
```
MarkLark86: Unfortunately there are Assignment E2E tests failing now, I wonder if it's due to these changes.
image

[get_archive_items_for_assignments](https://github.com/superdesk/superdesk-planning/blob/0e20b3547aba5eb65ce4e248ff3b34f1e4c06722/server/planning/assignments/assignments.py#L139) returns an async cursor, so we would need to get the list of items from it (as we're iterating them)

server/features/events_spike.feature
```diff
@@ -96,6 +96,10 @@ Feature: Events Spike
         }
         """
         Then we get OK response
+        When we post to "/events/#events._id#/lock"
```
MarkLark86: Spike action's shouldn't require a lock (originally they didn't, check the service [here](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/events/events_spike.py#L45) and [here](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/events/events_spike.py#L251)).

So I'm guessing [here](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/events/events_spike.py#L247) and [here](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/events/events_spike.py#L297) we would send `require_lock = False`

MarkLark86: I guess the same goes for any other item service that had the `REQUIRE_LOCK = False` on the old eve service, we would need to update the new endpoint/utils to not require lock either

server/features/search_autocomplete.feature
```diff
@@ -99,15 +99,23 @@ Feature: Planning autocomplete
...
+        {"_items": [
+            {"value": "coverage-en-slugline", "count": 1}
+            {"value": "event-en-test", "count": 1},
+            {"value": "planning-1", "count": 1},
+        ]}
         """
         # Suggests base field if language translation not populated
         When we get "/archive_autocomplete?field=slugline&language=fr"
-        Then we get list with 1 items
+        Then we get list with 3 items
```
MarkLark86: We shouldn't be changing the details of the original test. In this scenario there should only be 1 result.
I'm guessing it's because of an issue of the DBs not clearing between Scenario's? If so leave these ones and I will have a look

server/features/search_combined.feature
```diff
@@ -3,7 +3,7 @@ Feature: Search Events and Planning
         Given "agenda"
         """
             [
-                {"name": "sports", "_id": "sports", "is_enabled": true},
+                {"name": "sports", "_id": "68345becf63aa4e3993ff94d", "is_enabled": true},
```
MarkLark86: Rather than specifying the IDs manually, we can now use then with expansion. In this case
`"#agenda_0._id#"` Would be the ID for Sports Agenda
`"#agenda_1._id#"` Would be the ID for Finance Agenda
`"#agenda_2._id#"` Would be the ID for Entertainment Agenda

MarkLark86: See https://github.com/superdesk/superdesk-core/blob/async/superdesk/tests/steps.py#L524 on how it's added

server/features/steps/steps.py
```diff
@@ -175,7 +175,8 @@ async def step_imp_when_action_resource(context, action, resource, item_id):
 @then('we get text in "{field}"')
 @async_run_until_complete
 async def then_we_get_text_in_response_field(context, field):
-    response = await get_json_data(context.response)[field]
+    data = await get_json_data(context.response)
```
MarkLark86, line 178: Good catch, these tend to always slip past us

MarkLark86:
@BrianMwangi21 Looks good, just a couple small things

server/planning/events/events_cancel.py
```diff
@@ -152,7 +152,7 @@ async def process_cancel_event(updates: dict[str, Any], original: dict[str, Any]
     ACTION = "cancel"
 
     # Perform pre update event actions
-    await pre_update_event_actions(updates, original, ACTION)
+    await pre_update_event_actions(updates, original, ACTION, False)
```
MarkLark86: I believe `cancel` and `postpone` actions require lock. Please have another look

MarkLark86:
Behave tests are failing to run, could you please look into it?
```
  File "features/planning_content_api/environment.py", line 16, in <module>
    from content_api.app.settings import MODULES
ImportError: cannot import name 'MODULES' from 'content_api.app.settings'
```

I believe the new name is [CONTENT_API_MODULES](https://github.com/superdesk/superdesk-core/blob/async/content_api/app/settings.py#L61).

In fact I believe due to recent changes in core the following could be changed:
https://github.com/superdesk/superdesk-planning/blob/7b241defc6a9c5f82a6ffe91252513821a8361b5/server/features/planning_content_api/environment.py#L33
to:
```python
config = update_config({}, auto_add_apps=False) 
```

BrianMwangi21:
@MarkLark86 the Behave Content API is now running

MarkLark86 (approved):
:+1: Looks good to merge, progress

PR #2324, opened 2025-05-27
Fix serving of swagger.yaml in content API docs [stt-107]

devketanpro:
- Renamed static route from `/planning-static` to `/api-planning-static` to match Nginx routing
- Added 'static/*.yaml' to package_data in setup.py to ensure swagger.yaml is included during installation
- Added Security check: ensure the file stays within base_path.

### DIFF
--- a/server/planning/content_api/content_api_docs.py
+++ b/server/planning/content_api/content_api_docs.py
@@ -8,7 +8,7 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 import os
-from quart import send_file, abort
+from quart import abort, send_from_directory
 from superdesk.core.web import EndpointGroup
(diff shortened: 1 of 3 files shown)


MarkLark86:
@devketanpro Please use the [send_from_directory](https://quart.palletsprojects.com/en/latest/reference/source/quart.helpers.html#quart.helpers.send_from_directory), as it's more secure and does these checks for us.

> This is a secure way to serve files from a folder, such as static files or uploads. Uses [safe_join()](https://werkzeug.palletsprojects.com/en/stable/utils/#werkzeug.security.safe_join) to ensure the path coming from the client is not maliciously crafted to point outside the specified directory.

It also sends the file back using async

MarkLark86 (approved):

eos87 (approved):
looks good!

PR #2339, opened 2025-06-02
[SDESK-7497] Fix Event Cancel & Reschedule endpoints

MarkLark86:
### Purpose
To continue fixing behave tests and any bugs found. This focus was on Event Cancel & Reschedule actions.

### What has changed
* Don't raise an exception on publish `routed == False` (keeping to existing functionality of < v3.0, we can revisit this later with front-end improvements to publishing)
* Fix dev requirements.txt, we need to install superdesk-core in editable state to include feature fixtures. We need to revisit the superdesk-core setup.py and use newer standards (and support installs like `superdesk-core[tests]`)
* Fix `file_providers` output formatter unit tests
* Fix Event Cancel & Reschedule endpoints (various issues, from invalid date/time instances to using `create_async` and `update_async` instead of `post_async` and `patch_async` which have side-affects, such as creating more Events or automatically posting the Event)
* Clear AsyncSignals before running behave tests (as old service was keeping a reference to previous MongoDB connection which was closed).

### DIFF
--- a/server/features/environment.py
+++ b/server/features/environment.py
@@ -15,9 +15,10 @@ from os import path
 from copy import copy
 
 from apps.prepopulate.app_populate import AppPopulateCommand
+from superdesk.core import AsyncSignal
 from superdesk.tests.environment import (
     setup_before_all,
-    before_feature,
+    before_feature_async as setup_before_feature_async,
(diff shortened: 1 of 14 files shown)


eos87 (approved):
Looks good

PR #2343, opened 2025-06-04
[SDESK-7497] Fix remaining behave tests

MarkLark86:
### Purpose
To get all the behave tests passing

### What has changed
* Fix Assignment locks (missing `await` prefix)
* Fix Assignment publish Planning (missing `await` prefix)
* Fix Event publish validation (validation statement was commented out from early on in project)
* Fix remaining Planning actions (use Eve services, use `update_async` instead of `patch_async`, make sure to call signals)
* Fix sending email notifications (missing `await` prefix)
* Fix FeaturedPlanning ID generation (was being done on the service, I moved it onto the Model itself)
* Fix base PlanningSearch (migrated to use our async eve elastic service as previous version of newer elastic client was failing on projections).

### DIFF
--- a/server/features/assignments_link.feature
+++ b/server/features/assignments_link.feature
@@ -1134,7 +1134,7 @@ Feature: Assignment link
         """
         {
         "_items": [
-          {"state": "pending", "content_type": "text",
+          {"state": "success", "content_type": "text",
           "subscriber_id": "#wire#", "item_id": "123", "item_version": 2,
           "ingest_provider": "__none__",
           "destination": {
--- a/server/features/assignments_revert.feature
+++ b/server/features/assignments_revert.feature
@@ -7,6 +7,16 @@ Feature: Assignment Revert
     """
     [{"_id": "desk123", "name": "Politic Desk"}]
     """
+    When we post to "planning" with success
+    """
+    [{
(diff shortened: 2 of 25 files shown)


server/features/assignments_link.feature
```diff
@@ -1134,7 +1134,7 @@ Feature: Assignment link
         """
         {
         "_items": [
-          {"state": "pending", "content_type": "text",
+          {"state": "success", "content_type": "text",
```
eos87, line 1137: what is the reason for this state difference?

eos87 (approved):
It looks good. Just wondering about the change in the state plus why those tests were removed. Why aren't they needed any longer?

server/features/assignments_link.feature
MarkLark86: With the new publish async system, items are processed straight away which means their state in the publish queue is `success`. Previously they were marked `pending` and picked up later in a celery beat task

MarkLark86:
> It looks good. Just wondering about the change in the state plus why those tests were removed. Why aren't they needed any longer?

The steps for vocabularies were removed because those documents are already added before the feature is run, so the extra step is not required to re-add the same docs.

PR #2358, opened 2025-06-12
[SDESK-7681] - Rescheduling of a recurring event yields errors

BrianMwangi21:
### Purpose
This PR fixes an issue where a normal update to a recurring event (e.g. changing the slugline) fails due to `moment` objects lingering in the `dates.recurring_rule.until` field. This field remains unnormalized after a previous rescheduling action, and is incorrectly treated as a new update—even though no change occurred—because it's still a `moment` object.

### What has changed
- Updated `modifyForServer()` to ensure `dates.recurring_rule.until` is always serialized to an ISO string, regardless of timezone context. This prevents the `getEventDiff()` function from falsely detecting `until` as an update 

### Steps to test
1. Create a recurring event with no time specified.
2. Go to Reschedule and provide a valid date/time. Confirm the event is successfully rescheduled.
3. Then open the same event and perform a minor unrelated update like changing the slugline.
4. Click Save and ensure the update succeeds.
5. Observe that no `dates.recurring_rule.until` field is being unnecessarily sent unless actually changed.

Resolves: [SDESK-7681](https://sofab.atlassian.net/browse/SDESK-7681)

[SDESK-7681]: https://sofab.atlassian.net/browse/SDESK-7681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/utils/events.tsx
+++ b/client/utils/events.tsx
@@ -1271,6 +1271,10 @@ function modifyForServer(event: IEventItem, removeNullLinks: boolean = false) {
         event.dates.end = event.dates.end.toISOString();
     }
 
+    if (event.dates?.recurring_rule?.until != null && moment.isMoment(event.dates.recurring_rule.until)) {
+        event.dates.recurring_rule.until = event.dates.recurring_rule.until.toISOString();
+    }
+
     return event;
 }
 


BrianMwangi21:
@MarkLark86 in the ticket it mentions 

`Go to the event rescheduling and notice the date of the event is already in the past, also the time is pre-selected although it was not chosen initially.`

I'm guessing this is because if you create an event with no time it puts midnight by default so that's expected behavior right ?

MarkLark86 (approved):

MarkLark86:
> @MarkLark86 in the ticket it mentions
> 
> `Go to the event rescheduling and notice the date of the event is already in the past, also the time is pre-selected although it was not chosen initially.`
> 
> I'm guessing this is because if you create an event with no time it puts midnight by default so that's expected behavior right ?

I'm guessing so, there has been some changes around the ability to make time optional.

PR #2359, opened 2025-06-13
Add sorting config

thecalcc:
STT-90

### DIFF
--- a/client/apps/PageContent.tsx
+++ b/client/apps/PageContent.tsx
@@ -5,6 +5,10 @@ import {WorkqueueContainer, ModalsContainer} from '../components';
 import {PopupEditorPortal} from '../components/Main/ItemEditorModal';
 
 import './style.scss';
+import {memoize} from 'lodash';
+import {IPlanningExtensionConfigurationOptions} from 'planning-extension/src/extension_configuration_options';
+import {superdeskApi} from '../superdeskApi';
+import {IEventOrPlanningItem} from 'interfaces';
 
 export interface IBasePanelProps {
     toggleFilterPanel(): void;
@@ -119,6 +123,13 @@ export class PageContent<T> extends React.Component<IProps<T>, IState> {
             'sd-page-content__content-block--30-slide'
         );
 
+        const extensionConfig: IPlanningExtensionConfigurationOptions = superdeskApi.getExtensionConfig();
+        const memoizedSort = memoize((items: Array<IEventOrPlanningItem>) =>
+                    extensionConfig?.comparePlanningItems != null
+                        ? items.sort(extensionConfig.comparePlanningItems)
+                        : items
+        );
+
         return (
             <div className={sectionClassName}>
                 <div className={mainClassName} aria-labelledby="planning-heading">
@@ -157,6 +168,7 @@ export class PageContent<T> extends React.Component<IProps<T>, IState> {
                                 <ListPanel
                                     previewOpen={previewOpen}
                                     toggleFilterPanel={this.toggleFilterPanel}
+                                    sortItems={memoizedSort}
                                     {...listProps}
                                 />
                             </div>
--- a/client/apps/PageContent.tsx
+++ b/client/apps/PageContent.tsx
@@ -5,10 +5,6 @@ import {WorkqueueContainer, ModalsContainer} from '../components';
 import {PopupEditorPortal} from '../components/Main/ItemEditorModal';
 
 import './style.scss';
-import {memoize} from 'lodash';
-import {IPlanningExtensionConfigurationOptions} from 'planning-extension/src/extension_configuration_options';
-import {superdeskApi} from '../superdeskApi';
-import {IEventOrPlanningItem} from 'interfaces';
 
 export interface IBasePanelProps {
     toggleFilterPanel(): void;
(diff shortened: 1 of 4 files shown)


client/components/Main/ListPanel.tsx
```diff
@@ -375,7 +377,14 @@ export class ListPanel extends React.Component<IProps, IState> {
...
                             let listGroupProps: {[key: string]: any} = {
                                 name: group.date,
-                                items: group.events,
+
+                                /**
+                                 * lodash `get` is used, so we can access the sort method from lodash
+                                 * and pass a scoring function to it. JS Array.prototype.sort is considered unreliable
+                                 */
+                                items: extensionConfig?.comparePlanningItems != null
+                                    ? group.events.sort(extensionConfig.comparePlanningItems)
```
petrjasek: I wonder if we should use some memo here, or is it fast enough we shouldn't bother?

thecalcc: seemed fast enough to me when I tested on stt uat, but it's a good idea if user goes back and forth would be useless to re-run

client/components/Main/ListPanel.tsx
```diff
@@ -373,9 +375,18 @@ export class ListPanel extends React.Component<IProps, IState> {
                                 getChildId: getChildId,
                             };
 
+                            const memoizedSort = memoize(group.events.sort);
```
petrjasek: I wonder what this does actually, that extension sort func you pass later won't ever change, so it will never reorder those items right? but isn't that `sort` a function on prototype so it will be also the same for all? or it handles somehow the array instance?

thecalcc: you're right, docs says that a cache key is created based on the supplied args, I changed it so now it should be working based on provided items

client/components/Main/ListPanel.tsx
```diff
@@ -326,6 +327,12 @@ export class ListPanel extends React.Component<IProps, IState> {
         } = this.props;
 
         let indexFrom = 0;
+        const extensionConfig: IPlanningExtensionConfigurationOptions = superdeskApi.getExtensionConfig();
+        const memoizedSort = memoize((items) =>
```
petrjasek: I wonder if that will actually memoize it if it's created in render, wouldn't it be new instance each time? maybe it should be created on `componentDidMount` or something similar

client/components/Main/ListPanel.tsx
```diff
@@ -19,7 +19,8 @@ import {onEventCapture} from '../../utils';
 
 import {ListGroup} from '.';
 import {PanelInfo} from '../UI';
-import {Item, Column, Group} from '../UI/List';
+import {Column, Group} from '../UI/List';
+import {IPlanningExtensionConfigurationOptions} from '../../planning-extension/src/extension_configuration_options';
```
tomaskikutis, line 23: You shouldn't be importing from an extension directly. See if it works to use the bridge.

client/components/Main/ListPanel.tsx
tomaskikutis: Yep, memoization result has to be stored on the component

client/components/Main/ListPanel.tsx
thecalcc: I'm importing a type so it's not breaking the build

client/components/Main/ListPanel.tsx
thecalcc: doing it on `componentDidMount` in the same component wouldn't help since props will change each time user changes filters and that would remount the component. Moving it up seems better, because that component loads once and then it doesn't keep remounting on filter change

client/components/Main/ListPanel.tsx
tomaskikutis: ok then

client/apps/PageContent.tsx
```diff
@@ -119,6 +123,13 @@ export class PageContent<T> extends React.Component<IProps<T>, IState> {
             'sd-page-content__content-block--30-slide'
         );
 
+        const extensionConfig: IPlanningExtensionConfigurationOptions = superdeskApi.getExtensionConfig();
+        const memoizedSort = memoize((items: Array<IEventOrPlanningItem>) =>
```
petrjasek: imo that has the same issue still, just moved a level higher. so when you eg open item for editing it will rerender everything. should be like:
```
componentDidMount() {
    self.sort = _.memoize(...);
}
...
render () {
    const items={self.sort(group.events);
}
```

client/components/Main/ListPanel.tsx
tomaskikutis: > doing it on componentDidMount in the same component wouldn't help since props will change each time user changes filters and that would remount the component.

If prop change remounts a component - that's another issue that should be handled. Prop change shouldn't unmount the component in most cases.

client/apps/PageContent.tsx
tomaskikutis: yep, `memoize` call should never be inside `render` method because cache will get dropped on every render. And we want to cache it until component unmounts. It might be that some component doesn't re-render often, but there's no reason to rely on that when there is an alternative that performs well even if component re-renders often.

client/components/Main/ListPanel.tsx
petrjasek: what about setting that memoization in constructor then? I mean that remounting should be avoided anyhow, but this might be also a bit cleaner, there is no need to wait for mount

petrjasek (approved):

PR #2365, opened 2025-06-16
update client 20250616

tomaskikutis:
fixing build


petrjasek:
updated it in https://github.com/superdesk/superdesk-planning/pull/2366

PR #2375, opened 2025-06-24
Handle pagination of planning items

thecalcc:
STT-1225


client/actions/planning/api.ts
```diff
@@ -115,36 +116,101 @@ function query(
...
+
+    const itemsGrouped = planningUtils.getPlanningByDate(
+        items,
+        {},
+        params.advancedSearch.dates.start,
+        params.advancedSearch.dates.end,
+        params.timezoneOffset,
+    );
+
+    if (itemsGrouped.length === 1) {
```
petrjasek: I'm just wondering if there should be a branch when lenght == 1 but you already fetched all items, but I suppose that promises list will be empty and it just returns the list right?

client/actions/planning/ui.ts
```diff
@@ -125,11 +125,23 @@ const loadMore = () => (
...
                 }
-                dispatch(self.addToList(items.map((p) => p._id)));
-                return Promise.resolve(items);
+
+                const result: Array<IPlanningItem> = items;
+
+                if ((lastDayGroupItems?.events ?? []).length > 0) {
+                    // add items from group to stored planning items, so grouping works later
+                    dispatch(planningApis.receivePlannings(lastDayGroupItems.events));
+                    result.push(...lastDayGroupItems.events);
```
petrjasek: I wonder if it should rather do `lastDayGroupItems + items`, the new items in theory go after, but it's probably re-sorted later anyhow right?

thecalcc: yep, they do get re-sorted, since these are ungrouped items, later there is a call to prepare them in groups for the list UI

client/actions/planning/api.ts
thecalcc: yes, I adjusted that branch so it returns explicitly all items if there's no pages to fetch for

thecalcc:
@petrjasek could you check my latest changes?

e2e/cypress/e2e/search/search_planning.cy.ts
```diff
@@ -135,10 +135,9 @@ describe('Search.Planning: searching planning items', () => {
                 'start_date.date': '02/02/2045',
                 'end_date.date': '03/02/2045',
             },
-            expectedCount: 2,
+            expectedCount: 1,
```
petrjasek: why is it less items now?

thecalcc: the last group won't show up since I cut it

petrjasek: are there that many items that it should do it?

client/actions/planning/api.ts
```diff
@@ -115,36 +117,114 @@ function query(
...
+            ...itemsForDate,
+            ...results.flatMap((x) => x._items),
+        ]);
+    } else if (Object.keys(itemsGrouped).length > 1) {
+        const lastGroup = itemsGrouped[itemsGrouped.length - 1];
+
+        dispatch(storeLastDayGroup(lastGroup));
+
+        return Promise.resolve(
+            itemsGrouped.slice(0, itemsGrouped.length - 1).flatMap((x) => x.events),
```
petrjasek, line 183: I suppose the issue is here where it wouldn't check if it's at the end of the results, in that case it shouldn't slice anything

thecalcc: yep, I think I can keep track of that by `itemsGrouped.flatMap((x) => x.events).length === params.total`, I'll have to do some testing though

thecalcc: tested and I think this check I added before the function call is ok and there's no need to process items further

petrjasek: I just wonder now what would happen if you are on page 2, you get 3 days back, and it's the end of result set. would you see the last day?

thecalcc: if it's then end then with this condition I added we won't have any difference in the UI, we'd just stick to the old flow since items will be passed to the next `.then`. 

The `total` of items is based on the `query` we send to the endpoint right?

the real shortcoming of how I'm doing it now is that we don't know how many items there are for a given date, we're only aware of total pages for all items.

thecalcc: in this code here we don't yet have a concept of the items as being grouped by day, this is done later in the UI -> `PlanningList.tsx` line 80, `groups: selectors.main.itemGroups(state),`, for planning only view this itemGroups runs `orderedPlanningList` which does the grouping and ordering of items

petrjasek: yep I'm just wondering if it can detect that's it's the very end of results and it won't skip the last day.
total is for all days

thecalcc: oh wait so total isn't calculated based on the range like `start: today; end: today+5 days` and total returned is `300` for this range, while all items in DB are e.g. `1000`?

petrjasek: it should be calculated based on the range

thecalcc:
I tested a case where the items in group 1 are `1,` and group two is `50`. With the old implementation this wasn't covered and if you were on page one, you saw just 1 group and also you weren't able to scroll and load more since the container had a single item thus overflow scroll didn't activate. I did this by passing items returned from `query` to `handleItemsForLastFetchedDay`, which in turn passes it to `loadMore` on page `1`, which goes into a recursion with `fetch`, etc., this way we keep track of total items and eventually get all items and in the end return all items where `response._meta.total === itemsInList.length` is `true` in `fetch`. I also tested it with a bunch of different cases, seems to be working, but ideally we'd do further testing with QA

client/actions/planning/api.ts
```diff
@@ -115,36 +117,132 @@ function query(
...
+        [...items, ...lastGroupItems],
+        {},
+        params.advancedSearch.dates.start,
+        params.advancedSearch.dates.end,
+        params.timezoneOffset,
+    );
+
+    if (itemsGrouped.length === 0) {
+        return Promise.resolve([...items, ...lastGroupItems]);
+    } else if (itemsGrouped.length === 1 || params.page === 1) {
```
petrjasek: why check the params.page here?

client/actions/planning/api.ts
```diff
@@ -115,36 +117,132 @@ function query(
...
+                return response._items;
+            }
+
+            const lastDayGroupItems = selectors.planning.lastDayGroup(getState())?.events ?? [];
+            const itemsInList = selectors.planning.planIdsInList(getState());
+            const storedPlannings = selectors.planning.storedPlannings(getState());
+
+            // response._meta.total - total items for provided range
+            // all items for the query were already fetched
+            if (response._meta.total === itemsInList.length) {
```
petrjasek: imo it should check if it fetched all items, that would be something like:
```
if (response._meta.total === itemsInList.length + lastDayGroupItems.length + response._items.length)
```

client/actions/planning/api.ts
thecalcc: because on initial load, we fetch page one, but we fetch it using `fetch` and not `loadMore`, so I need to run this branch regardless if `itemsGrouped` is `1`, so I make sure that I fetch all the items for the groups

thecalcc: otherwise on initial load we may end up with incomplete items, and a sub edge case to that is group 1 having 1 or a couple items not covering the whole page, the second group being sliced, and user ends up in a state where `loadMore` is also impossible because number of items isn't more than the view height

client/actions/planning/api.ts
thecalcc: this checks if all items were already loaded, thus they'd all be in `itemsInList`

client/actions/planning/api.ts
petrjasek: I see, can use some comment imo

client/actions/planning/api.ts
petrjasek: but isn't it checking when it has more items in `response._items` but before those were added to `itemsInList` ?

thecalcc: yep, maybe I'll move this check before loadMore is called, that way if all items were already fetched and to `itemsInList` by `loadMore` fetch isn't called for no reason

client/actions/planning/ui.ts
```diff
@@ -123,13 +124,25 @@ const loadMore = () => (
...
                     dispatch(self.requestPlannings(params));
                 }
-                dispatch(self.addToList(items.map((p) => p._id)));
-                return Promise.resolve(items);
+
+                const result: Array<IPlanningItem> = items;
+
+                if ((lastDayGroupItems?.events ?? []).length > 0) {
+                    // add items from group to stored planning items, so grouping works later
+                    dispatch(planningApis.receivePlannings(lastDayGroupItems.events));
```
petrjasek: why is it added here? I thought those items shouldn't be used until they are returned from fetch (when the group is fetched completely)

client/actions/planning/api.ts
```diff
@@ -115,36 +117,136 @@ function query(
...
+        // otherwise if the first group has 1 item, a second has 50 items, user
+        // can end up in an invalid state where he can't trigger loadMore by scrolling
+    } else if (itemsGrouped.length === 1 || params.page === 1) {
+        const pageToFetchUntil = Math.ceil(total / params.maxResults);
+
+        // end is incremented by 1, to include last page, start is
+        // incremented by 1 so we don't fetch for a page, we already have
+        const allPages = range(params.page + 1, pageToFetchUntil + 1);
+
+        if (allPages.length < 1) {
```
petrjasek: this could be simplified now:
```ts
const lastPage = Math.ceil(total / params.maxResults);

if (params.page + 1 > lastPage) {
    ...
```

client/actions/planning/api.ts
```diff
@@ -115,36 +117,136 @@ function query(
...
+            return Promise.resolve(itemsGrouped.flatMap((x) => x.events));
+        }
+
+        const itemsInList = selectors.planning.planIdsInList(getState());
+
+        // all items were already fetched from previous
+        // requests and are already stored in the list items.
+        // Mostly used for when switching range display - Day, Week, Month
+        if (total === itemsInList.length) {
```
petrjasek: I think this should be tested on top, probably combined with:
```ts
if (total === (items ?? []).length + (lastDayGroupItems ?? []).length) {
```
where it should test not only `items + lastDayGroupItems` but also `itemsInList` otherwise it can only be trye on    some initial loading

thecalcc: That might be ok, if I combine those two checks I can then return to even save further fetching since I check for `itemsInList`.

```js
uniqBy([
    ...listItems,
    ...items,
    ...lastDayGroupItems
], (x: IPlanningItem) => x._id)
```

This uniqBy would be to make sure there's no duplicate items stored further down. 

But I wonder, would it then be possible for user on occasion to select some range where item length is the same, and then they'd end up with wrong items. What do you think?

petrjasek: so when you change query params and it won't reset listItems?

thecalcc: no, if there's any that are already stored, whenever there's a loadMore/fetch we add to `storedPlannings` and `planIdsInList`

thecalcc: and we only kind of reset the `planIdsInList` after fetching

client/actions/planning/api.ts
```diff
@@ -115,36 +117,133 @@ function query(
...
+        const listItems: Array<IPlanningItem> = (itemIdsInList ?? []).map((id) => storedPlannings[id]);
+
+        // listItems might contain some or all of items, get unique set of all
+        const allItems = uniqBy([
+            ...listItems,
+            ...items,
+            ...lastDayGroupItems
+        ], (item) => item._id);
+
+        return Promise.resolve(allItems);
```
petrjasek: I wonder why it should return also items which are already in the list? afaik that combining is happening later in reducer. here it should only return `items + lastDayGroupItems`, not any items fetched before

client/actions/planning/api.ts
```diff
@@ -115,36 +117,133 @@ function query(
...
+    getState: () => any,
+): Promise<Array<IPlanningItem>> => {
+    if (items.length < 1) {
+        return Promise.resolve([]);
+    }
+
+    const lastDayGroupItems = selectors.planning.lastDayGroup(getState()) ?? [];
+    const itemIdsInList = selectors.planning.planIdsInList(getState());
+
+    const hasFetchedTotalItems = total === (items ?? []).length + (lastDayGroupItems ?? []).length;
```
petrjasek: imo this can only be true when you fetch all items on the first load (with optional loadMore call when you get single group), because you're ignoring the items fetched previously.
if you would also count items already rendered it should be ok and you wouldn't need that next one, which I wonder when it's actually true, probably when total is 0 or if it's calling the method after fetching 0 items (so you know you're at the end but still call it once more)

petrjasek (approved):

PR #2380, opened 2025-06-26
TBC feature in time picker causes UI issue

dzonidoo:
SDESK-7680

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/components/UI/Form/TimeInput/index.tsx
+++ b/client/components/UI/Form/TimeInput/index.tsx
@@ -36,7 +36,6 @@ export class TimeInput extends React.Component {
         this.dom = {
             inputField: null,
             hiddenSpan: null,
-            inputWrapper: null,
             iconWrapper: null,
         };
 
@@ -244,14 +243,14 @@ export class TimeInput extends React.Component {
     }
 
     updateInputWidth() {
-        if ((this.props.fullWidth === false) && this.dom.hiddenSpan && this.dom.inputField && this.dom.inputWrapper) {
-            const width = this.dom.hiddenSpan.offsetWidth
-                + this.inputPadding
-                + this.dom.iconWrapper.offsetWidth
-                + this.iconPadding;
+        if ((this.props.fullWidth === false) && this.dom.hiddenSpan && this.dom.inputField) {
+            const characterCount = this.state.viewValue.length;
 
-            this.dom.inputField.style.width = `${width}px`;
-            this.dom.inputWrapper.style.width = `${width}px`;
+            const additionalPixelWidth = this.inputPadding
+            + this.dom.iconWrapper.offsetWidth
+            + this.iconPadding;
+
+            this.dom.inputField.style.width = `calc(${characterCount}ch + ${additionalPixelWidth}px)`;
         } else {
             return null;
         }
@@ -312,8 +311,11 @@ export class TimeInput extends React.Component {
             >
                 <Label text={label} />
                 <div
-                    style={{position: 'relative', paddingBlock: '1.8rem'}}
-                    ref={(ref) => this.dom.inputWrapper = ref}
+                    style={{
+                        position: 'relative',
+                        paddingBlock: '1.8rem',
+                        display: 'inline-flex',
+                    }}
                 >
                     <IconButton
                         className="sd-line-input__icon-right"
--- a/client/components/UI/Form/TimeInput/index.tsx
+++ b/client/components/UI/Form/TimeInput/index.tsx
@@ -187,7 +187,7 @@ export class TimeInput extends React.Component {
             this.setState({
                 invalid: false,
                 viewValue: (newValue.length === 4 ? ('0' + newValue) : newValue),
-            }, this.updateInputWidth);
+            });
         }
     }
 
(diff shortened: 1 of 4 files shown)


client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -100,14 +117,14 @@ export class TimeInput extends React.Component {
                 invalid: true,
                 viewValue: val,
                 showLocalValidation: false,
-            });
+            }, this.updateInputWidth);
```
tomaskikutis: Do not add `this.updateInputWidth` call here. Move to `componentDidUpdate` instead.

client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -21,6 +21,8 @@ import './style.scss';
  * @description Component to pick time in hours and minutes
  */
 export class TimeInput extends React.Component {
+    inputPadding: number;
+    iconPadding: number;
```
tomaskikutis: set properties to private

tomaskikutis (commented):
I've searched for simpler approaches to this. I found a `ch` unit which will represents character width. See if you can use it.

`<input type="text" style={{width: '5ch'}} />`

dzonidoo:
i did not manage to do with this approach `<input type="text" style={{width: '5ch'}} />`

tomaskikutis:
Could you elaborate on why did it not work out?

dzonidoo:
Probably because of some CSS styles or wrappers… we pull in quite a lot of that from framework

tomaskikutis:
Which step in the process did you get stuck on?

client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -53,7 +61,7 @@ export class TimeInput extends React.Component {
                 previousValidValue: '',
                 invalid: false,
                 showLocalValidation: false,
-            });
+            }, this.updateInputWidth);
```
tomaskikutis: You're already calling it in componentDidUpdate - the call here shouldn't be needed.

client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -77,7 +85,15 @@ export class TimeInput extends React.Component {
             value.format(appConfig.planning.timeformat) :
             (value || '');
 
-        this.setState({viewValue});
+        this.setState({viewValue}, this.updateInputWidth);
+
+        if (this.dom.inputField) {
+            this.inputPadding = parseFloat(window.getComputedStyle(this.dom.inputField).paddingInline) * 2;
```
tomaskikutis: Why are you multiplying by 2? Can you get the properties that you need directly? maybe paddingInlineStart + paddingInlineEnd ?

client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -77,7 +85,15 @@ export class TimeInput extends React.Component {
             value.format(appConfig.planning.timeformat) :
             (value || '');
 
-        this.setState({viewValue});
+        this.setState({viewValue}, this.updateInputWidth);
+
+        if (this.dom.inputField) {
```
tomaskikutis: can this be moved inside `updateInputWidth`?

client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -211,12 +236,26 @@ export class TimeInput extends React.Component {
...
                     previousValidValue: newValue,
                     showLocalValidation: false,
-                });
+                }, this.updateInputWidth);
             }
             onChange(field, newMoment);
         }
     }
 
+    updateInputWidth() {
```
tomaskikutis: add `private`

client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -266,31 +307,56 @@ export class TimeInput extends React.Component {
                 errors={errors}
                 message={message}
                 boxed={true}
+                className={fullWidth === false ? 'no-padding' : ''}
```
tomaskikutis: if you don't need a value - pass undefined instead of an empty string

client/components/UI/Form/TimeInput/index.tsx
```diff
@@ -266,31 +307,56 @@ export class TimeInput extends React.Component {
...
+                    <IconButton
+                        className="sd-line-input__icon-right"
+                        icon="icon-time"
+                        onFocus={onFocus}
+                        onClick={!readOnly ? this.toggleOpenTimePicker : null}
+                        aria-label={gettext('Time picker')}
+                        refNode={(ref) => this.dom.iconWrapper = ref}
+                    />
+
+                    <span
```
tomaskikutis: add a comment describing why do you need a hidden element

client/components/UI/Form/TimeInput/index.tsx
dzonidoo: I thought it would be better to get the padding value just once instead of on every onChange, but Ill move it into the updateInputWidth function.

client/components/UI/Form/TimeInput/index.tsx
dzonidoo: Thats my misstake... ill remove that hidden element

tomaskikutis (commented):
Remove `updateInputWidth` call from everywhere except component did mount/update

tomaskikutis (approved):

PR #2383, opened 2025-06-27
Respect config for vocabulary as field

thecalcc:
SDESK-7714


client/components/fields/editor/CustomCV.tsx
```diff
@@ -22,23 +21,32 @@ export class EditorFieldCV extends React.PureComponent<IEditorFieldProps> {
         } = this.props;
         const cv = superdeskApi.entities.vocabulary.getVocabulary(this.props.field);
 
+        if (cv.selection_type === 'do not show') {
```
tomaskikutis, line 24: use a constant for `'do not show'` string

thecalcc: this is type checked, since `selection_type` is typed as specific strings

tomaskikutis (approved):

PR #2385, opened 2025-06-30
add tooltips to contacts list icons

petrjasek:
SDESK-7692

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/components/UI/IconButton.tsx
+++ b/client/components/UI/IconButton.tsx
@@ -6,7 +6,6 @@ import {KEYCODES} from './constants';
 import {onEventCapture} from './utils';
 
 import {Icon} from './';
-import {Prop} from 'superdesk-ui-framework';
 
 /**
  * @ngdoc react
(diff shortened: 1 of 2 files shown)


client/components/UI/IconButton.tsx
```diff
@@ -6,6 +6,7 @@ import {KEYCODES} from './constants';
 import {onEventCapture} from './utils';
 
 import {Icon} from './';
+import {Prop} from 'superdesk-ui-framework';
```
thecalcc: is this needed?

thecalcc (approved):

client/components/UI/IconButton.tsx
petrjasek: shouldn't be :smile:

thecalcc: ideally I think also we'd replace usages with a component from UI Framework since we're doing the gradual migration - https://ui-framework.superdesk.org/#/components/icon-buttons and not add to PropTypes of the existing component

thecalcc (approved):

PR #2388, opened 2025-07-01
reuse event time field in convert to recurring modal

petrjasek:
STT-1227

### Purpose

Reusing a component to avoid duplication.

### Front-end checklist

Resolves: STT-1227


e2e/cypress/support/common/inputs/coverageUserSelectInput.ts
```diff
@@ -13,7 +13,8 @@ export class CoverageUserSelectInput extends Input {
         this.element.click();
         popup.waitTillOpen();
 
-        popup.element.find('[data-test-id="filter-input"]')
+        popup.element.find('[data-test-id="filter-input"]', {timeout: 2000})
```
thecalcc, line 16: is it taking 2 seconds now to mount that element or it's some network call that's being awaited?

client/components/Events/EventScheduleInput/EventScheduleInput_test.tsx
```diff
@@ -1,106 +0,0 @@
-import React from 'react';
```
thecalcc, line 1: so now we no longer need a test for this component because it's covered by some other test?

e2e/cypress/support/common/inputs/coverageUserSelectInput.ts
petrjasek: it's using tree select modal and it takes some time to be rendered

client/components/Events/EventScheduleInput/EventScheduleInput_test.tsx
petrjasek: yep the functionality it was testing is not part of that component anymore

thecalcc (approved):

PR #2389, opened 2025-07-01
[SDESK-7687] - Event schedule date gets stuck

BrianMwangi21:
### Purpose

Fixes an issue where all-day events would appear as the previous day in negative timezones (e.g., UTC-4) when selected. 

### What has changed

- Added an offset during rendering for all-day events to center the visual display. This is only a rendering offset and value saved remains correct

### Steps to test

1. Set your browser or OS timezone to a negative offset like `America/Toronto`
2. Create an all-day event and select different dates. Observe that the day selected is the day shown in the picker.
3. Switch to a positive timezone like `Africa/Nairobi` and repeat.

Resolves: [#SDESK-7687](https://sofab.atlassian.net/browse/SDESK-7687)

### DIFF
--- a/client/components/fields/editor/EventSchedule.tsx
+++ b/client/components/fields/editor/EventSchedule.tsx
@@ -67,9 +67,6 @@ export class EditorFieldEventSchedule extends React.PureComponent<IProps> {
         if (this.props.item.dates?.all_day === true || changes['dates.all_day'] === true) {
             value = localDateToUtc(value, this.props.item.dates?.tz);
             changes['dates.start'] = value;
-            console.log('[EditorFieldEventSchedule] after UTC conversion', {
-                value: value?.toISOString?.(),
-            });
         }
 
         if (endDate == null) {
--- a/client/components/fields/editor/EventSchedule.tsx
+++ b/client/components/fields/editor/EventSchedule.tsx
@@ -53,20 +53,20 @@ export class EditorFieldEventSchedule extends React.PureComponent<IProps> {
             changes['dates.all_day'] = true;
         }
 
-        if (startDate != null) {
-            value
-                .hour(startDate.hour())
-                .minute(startDate.minute())
-                .second(startDate.second());
-        } else {
-            value.hour(0)
-                .minute(0)
-                .second(0);
-        }
-
         if (this.props.item.dates?.all_day === true || changes['dates.all_day'] === true) {
-            value = localDateToUtc(value, this.props.item.dates?.tz);
+            value = safeLocalDateToUtc(value, this.props.item.dates?.tz);
             changes['dates.start'] = value;
+        } else {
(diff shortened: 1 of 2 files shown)


tomaskikutis:
Why was the issue happening?

BrianMwangi21:
@tomaskikutis The issue happened because all-day events are saved in utc at midnight, and when rendered in negative timezones like utc-4, it becomes like 8pm the previous day locally — so the date picker shows the wrong day. So you pick 15th, it shows 14th. 

You can see this in Nareg’s videos on the ticket, and I also confirmed it by changing my timezone to a negative one.

tomaskikutis:
The issue isn't specific to all day events. I reproduced it also happening for events with start/end times when browser is in Tokyo and event time zone is set to Toronto. See this [video](https://github.com/user-attachments/assets/a86d4d8c-abb1-4f89-80ec-eb99746b1cb5).

@petrjasek I saw you changed this code block [a few months ago](https://github.com/user-attachments/assets/7d9e41cf-8b0e-4a25-a986-53403dc33815). Do you remember how date pickers are supposed to work regarding timezones? From current UI I'd expect browser time zone not play any role since there's a separate dropdown to pick the time zone in the UI.

Regarding solutions - if the date that's being selected is correct, and it only renders using the wrong timezone - we should rather modify how `viewValue` is computed in the date picker component. How about formatting it using the time zone from the dropdown?

Current patch that adds hours to the value isn't a reliable fix. It will still break for [some](https://github.com/user-attachments/assets/fbe106f7-1e8f-4548-991e-40a335b78128) timezones and it also increases likelihood for other issues since you're passing wrong time to date picker as value and if that date picker was to add a few hours and trigger `onChange` - it would pass a wrong value too.

We could also look into replacing the date picker. I see `EditorFieldDateTime` is deprecated in favor of  `dateTimeUIFramework.tsx`. I'm not sure if the new one would have the same issue.

petrjasek:
yep if it's all day from 00:00-23:59 in specific timezone it has to use the timezone to properly display it, that `moment.utc` call was for cases when we have `dates.all_day` flag set and in that case there is no timezone and it should only use date from the timestamp without any conversion

BrianMwangi21:
@petrjasek @tomaskikutis I've updated the PR. 

I took in what you guys said, dug a little deeper and found that the issue was caused by rendering all-day events using the raw UTC value without applying the event’s timezone. So even though the data was correct, the picker would show the wrong day in some timezones, especially negative ones.

Instead of adding a manual offset, I now format the value as a simple date with no time and reapply the event's timezone. This ensures the correct date always shows up, regardless of browser timezone.

I've tested it manually across a wide range of timezones, from UTC-12 to UTC+14, and it renders correctly in all cases.

petrjasek (approved):

PR #2390, opened 2025-07-02
[SDESK-7697] - Saving prompt intiated even with no new changes

BrianMwangi21:
### Purpose
Fixes a false-positive dirty state in the Planning Editor caused by the `_scheduledTime` field. Despite having the same datetime value in the intialValues and in the diff, moment internals caused `_.isEqual` to falsely detect differences, triggering the "Save changes?" prompt even when no real changes were made on the editor.

This also led to the Save and Save & Post and other buttons being incorrectly enabled — as if the form was dirty — when in reality, no user modifications had occurred yet.

### What has changed
- Remove `_scheduledTime` from the coverages before performing the deep equality check on the cloned copies. Original data remains untouched but now prevents the false positive.

### Steps to test
1. Create a planning item with no coverage and save. Reopen it but make no changes. Click close. No prompt should be triggered.
2. Create a planning item with at least one coverage and save. Reopen it but make no changes. Click close. No prompt should be triggered.
3. In both cases above, also confirm that Save and Save & Post buttons remain disabled unless an actual edit is made.
4. Make real edits (e.g., slugline or description) and confirm that the dirty state triggers as expected.

Resolves: [SDESK-7697](https://sofab.atlassian.net/browse/SDESK-7697)

[SDESK-7697]: https://sofab.atlassian.net/browse/SDESK-7697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/utils/index.ts
+++ b/client/utils/index.ts
@@ -880,12 +880,14 @@ export const itemsEqual = (nextItem, currentItem) => {
 
     get(lhs, 'coverages', []).forEach(
         (coverage) => {
+            delete coverage.planning._scheduledTime;
             formatDate(coverage, 'planning.scheduled');
         }
     );
 
     get(rhs, 'coverages', []).forEach(
         (coverage) => {
+            delete coverage.planning._scheduledTime;
             formatDate(coverage, 'planning.scheduled');
         }
     );


client/utils/index.ts
```diff
@@ -880,12 +880,14 @@ export const itemsEqual = (nextItem, currentItem) => {
 
     get(lhs, 'coverages', []).forEach(
```
eos87, line 881: I think we can update this and take advantage of typescript null coalescing operator :)

client/utils/index.ts
```diff
@@ -880,12 +880,14 @@ export const itemsEqual = (nextItem, currentItem) => {
...
         (coverage) => {
+            delete coverage.planning._scheduledTime;
```
eos87, line 883: do you think deleting it won't impact other areas?

client/utils/index.ts
```diff
@@ -880,12 +880,14 @@ export const itemsEqual = (nextItem, currentItem) => {
...
             formatDate(coverage, 'planning.scheduled');
         }
     );
 
     get(rhs, 'coverages', []).forEach(
```
eos87, line 888: same

client/utils/index.ts
BrianMwangi21: @eos87 no it won't. The object we are deleting from is a deep clone of the actual object which is used to check if there's any new difference that needs to be saved.

client/utils/index.ts
BrianMwangi21: I get you but I feel like this is more cleaner. The function is relied on by other stuff and in some instances lhs or rhs can be null and I don't wanna stack to many `?.` if that makes sense.

eos87: Fair enough

eos87 (approved):

PR #2397, opened 2025-07-07
Add to Planning not working - update

dzonidoo:
STT-1237

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/components/Archive/ArchiveItem.tsx
+++ b/client/components/Archive/ArchiveItem.tsx
@@ -19,15 +19,13 @@ export const ArchiveItemComponent = ({item, priorities, urgencies, urgencyLabel}
         />
         <Column>
             <Row>
-                {urgencies.length > 0 && (
-                    <UrgencyLabel
-                        item={item}
-                        label={urgencyLabel}
-                        urgencies={urgencies}
-                        tooltipFlow="down"
-                        inline={true}
-                    />
-                )}
+                <UrgencyLabel
+                    item={item}
+                    label={urgencyLabel}
+                    urgencies={urgencies}
+                    tooltipFlow="down"
+                    inline={true}
+                />
                 <PriorityLabel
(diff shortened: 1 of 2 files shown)


petrjasek:
can you rebase it @dzonidoo and only push the actual fix?

dzonidoo:
sorry i change the base right now to be on right brach

dzonidoo:
@petrjasek now u can approve i guess

dzonidoo:
@petrjasek now u can approve i guess...

client/components/Archive/ArchiveItem.tsx
```diff
@@ -19,13 +19,15 @@ export const ArchiveItemComponent = ({item, priorities, urgencies, urgencyLabel}
...
             <Row>
-                <UrgencyLabel
-                    item={item}
-                    label={urgencyLabel}
-                    urgencies={urgencies}
-                    tooltipFlow="down"
-                    inline={true}
-                />
+                {urgencies.length > 0 && (
+                    <UrgencyLabel
```
petrjasek: would be good to handle it inside the component like we do in `PriorityLabel` so we don't have to wrap every usage

client/components/UrgencyLabel/index.tsx
```diff
@@ -6,7 +6,7 @@ import {get} from 'lodash';
 export const UrgencyLabel = ({item, urgencies, label, tooltipFlow, className, inline}) => {
     const qcode = get(item, 'urgency', null);
 
-    if (!qcode) {
+    if (!qcode || urgencies.length < 1) {
```
petrjasek, line 9: imo this might still crash if `const urgency = urgencies.find((u) => u.qcode === qcode);` is `null`, so would be good to check that too

petrjasek (approved):

PR #2399, opened 2025-07-11
As an editor using the Planning Details Widget, I want to have certain actions available concerning a related Planning item [STT-128]

devketanpro:


client/planning-extension/src/planning-details-widget.tsx
```diff
@@ -2,21 +2,125 @@ import * as React from 'react';
...
+        this.state = {
+            loading: !!props.article.assignment_id,
+            planningId: null,
+        };
+    }
+
+    componentDidMount() {
+        const {assignment_id} = this.props.article;
+
+        if (assignment_id) {
```
tomaskikutis: Only use booleans inside the `if` condition. If you have a `nullable` string, use `assignment_id != null` so it evaluates to a boolean. Update it for other conditions in this PR too.

client/planning-extension/src/planning-details-widget.tsx
```diff
@@ -2,21 +2,125 @@ import * as React from 'react';
...
+                    text={gettext('Fulfil Assignment')}
+                    icon="bolt"
+                    onClick={this.dispatchFulfilAssignment}
+                    expand
+                />
+            </div>
+        );
+    }
+
+    renderActionsLinked() {
```
tomaskikutis: Component methods should not return JSX. In this case you can inline it since it only has a single usage. If it had more usages you could either define a variable or extract it to a separate component.

tomaskikutis (approved):

tomaskikutis:
@thecalcc would you point Ketan to the CI fix?

thecalcc:
> @thecalcc would you point Ketan to the CI fix?

Hi @devketanpro you can take a look at a PR I made here - https://github.com/superdesk/superdesk-planning/pull/2403 that  has CI fixes for `CI-Client`. You can cherry pick commits: `308d829a9c3806a322928fff14ac32e54afcc284` and `3e77f4cb9a1868b38716fb538d295233b6639662`.

The build is failing for you in `CI-E2E` because you're importing a function from main planning to the `planning-extension` and that's causing circular dependencies. To fix that you can use `client/extension_bridge.ts` and use the function you need through that bridge. Check the comment at the top of that file, and see how other functions (e.g. `isContentLinkToCoverageAllowed`) to see how to bridge the function you need.

From what I see this should solve all CI issues, but if there's any more left after these two fixes I'll take another look.

To be more explicit - [this import](https://github.com/superdesk/superdesk-planning/pull/2399/files#diff-de30ace472acc2a2651dc8b8a60c8ee38256d1f1d0cc89af6879d790391404e4R5) breaks the build.

PR #2402, opened 2025-07-15
Fix anpa category disabled state and option removal

thecalcc:
STT-1252

### DIFF
--- a/client/components/editor-standalone/field-definitions/category-field.ts
+++ b/client/components/editor-standalone/field-definitions/category-field.ts
@@ -33,7 +33,7 @@ export const getCategoriesField = (): IFieldDefinition => ({
 
         const field: IAuthoringFieldV2 = {
             id: id,
-            name: superdeskApi.localization.gettext('Categories'),
+            name: superdeskApi.localization.gettext('ANPA Category'),
             fieldType: 'dropdown',
             fieldConfig: fieldConfig,
         };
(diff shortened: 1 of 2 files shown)


client/components/editor-standalone/field-definitions/category-field.ts
```diff
@@ -33,7 +33,7 @@ export const getCategoriesField = (): IFieldDefinition => ({
 
         const field: IAuthoringFieldV2 = {
             id: id,
-            name: superdeskApi.localization.gettext('Categories'),
+            name: superdeskApi.localization.gettext('ANPA Category'),
```
tomaskikutis, line 36: I think label should stay the same here and `ANPA Category` alias is applied with language overwrites when needed

tomaskikutis (approved):

client/components/editor-standalone/field-definitions/category-field.ts
thecalcc: I saw it in the main planning editor field it's called ANPA Category as fallback, so I decided to change it so it matches the fallback - https://github.com/superdesk/superdesk-planning/blob/38afe8e06b45d523b39eacf93eebd7dfb7490ac2/client/components/fields/editor/Categories.tsx#L25

Let me know what you think

tomaskikutis: I would leave it as is not to mess it up for people that aren't using the overwrite

PR #2405, opened 2025-07-17
Fix vocabulary crash in planning only view filter

thecalcc:
STT-1266

### DIFF
--- a/client/utils/contentProfiles.ts
+++ b/client/utils/contentProfiles.ts
@@ -227,96 +227,96 @@ export function getFieldNameTranslated(field: string): string {
     const {gettext} = superdeskApi.localization;
 
     switch (field) {
-        case 'recurring_rules':
-            return gettext('Recurring Rules');
-        case 'dates':
-            return gettext('Dates');
-        case 'language':
-            return gettext('Language');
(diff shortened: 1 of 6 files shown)


client/utils/contentProfiles.ts
```diff
@@ -78,7 +78,19 @@ export function getGroupFieldsSorted(
...
-export const COVERAGE_VOCABULARIES = new Set([
+/**
+ * Used to determine which custom vocabularies should not be registered as field
+ * of type `custom_vocabulary`.
+ */
+export const VOCABULARIES_TO_BE_EXCLUDED = new Set([
+    /**
+     * General vocabularies to be excluded:
+     */
+    'sttsubj',
```
tomaskikutis: This one looks STT specific. It should be configured from STT repo.

thecalcc: That's right. I chose not to make a config since it's also tied in the backend and this is a one off case rather. Let me know if you still want me to do a config

tomaskikutis: Do the config anyway.

tomaskikutis (approved):

PR #2407, opened 2025-07-17
As an editor using the Planning Details Widget I want to see all changes saved to the Planning item in real time [STT-127]

devketanpro:


client/planning-extension/src/planning-details-widget.tsx
```diff
@@ -29,35 +29,65 @@ export class PlanningDetailsWidget extends React.PureComponent<IArticleSideWidge
...
+                    loading: false,
+                });
+            })
+            .catch(() => {
+                this.setState({loading: false});
+            });
+    };
+
+    subscribeToUpdates = () => {
+        (superdesk as any)?.events?.on?.('planning:updated', this.onPlanningUpdated);
```
tomaskikutis: You shouldn't be casting `(superdesk as any)`. `superdesk.events` doesn't exist in the API, not sure how would it work?

client/planning-extension/src/planning-details-widget.tsx
```diff
@@ -29,35 +29,66 @@ export class PlanningDetailsWidget extends React.PureComponent<IArticleSideWidge
...
+                    planningId: planning._id,
+                    loading: false,
+                });
+            })
+            .catch(() => {
+                this.setState({loading: false});
+            });
+    };
+
+    subscribeToUpdates = () => {
```
tomaskikutis: mark properties as `private`

tomaskikutis (approved):

PR #2411, opened 2025-07-18
merge async

petrjasek:
### Front-end checklist

Resolves: #[issue-number]


petrjasek:
hi @MarkLark86 can you pls check the server changes?

server/features/content_profiles_coverage.feature
```diff
@@ -296,10 +296,11 @@ Feature: Coverage Content Profiles
...
         Then we get list with 1 item
 
-        When we post to "coverage_profiles"
-        """
-        {
-            "content_type": "text"
-        }
-        """
-        Then we get error 409
+        # SKIP (mongo indexes are not set)
```
MarkLark86, line 299: What's the problem here?
I see you've added the `coverage_profiles` async resource, but it does not have a data_class set (aka the model)

server/planning/content_profiles/module.py
```diff
@@ -0,0 +1,31 @@
...
+    # rest_endpoints=RestEndpointConfig(
+    #     resource_methods=["GET", "POST"],
+    #     item_methods=["GET", "PATCH"],
+    #     enable_cors=True,
+    # ),
+)
+
+coverage_profiles_resource_config = ResourceConfig(
+    name="coverage_profiles",
+    data_class=None,
```
MarkLark86, line 21: Using async resource without a data class is not tested, not sure how this will behave.
The type is mandatory ```data_class: type["ResourceModel"]```

MarkLark86 (approved):
Looks fine as far as I can tell, besides the new `coverage_profiles` async resource

server/features/content_profiles_coverage.feature
petrjasek: the mongo indexes are not created for the tests, looked to me like that was the case for all resources

server/planning/content_profiles/module.py
petrjasek: I was only setting it to get mongo index created, but that didn't work out

petrjasek:
made a new PR using upstream branch https://github.com/superdesk/superdesk-planning/pull/2430

PR #2418, opened 2025-07-23
As an editor using the Planning Details Widget I want to see all changes saved to the Planning item in real time [STT-127]

devketanpro:
Issue:
The Planning Details Widget did not reflect real-time updates when a planning item was modified by another user or in a different tab. Updates were only visible after manually refreshing the browser.

Fix:
Added a manual refresh mechanism using setInterval in the componentDidMount lifecycle method. This periodically re-fetches the latest planning item every 10 seconds. The updated planning item is then compared and rendered to ensure users see near real-time updates without needing to reload the page. The interval is cleared on unmount to avoid memory leaks.

### DIFF
--- a/client/components/PlanningDetailsWidget.tsx
+++ b/client/components/PlanningDetailsWidget.tsx
@@ -33,48 +33,47 @@ class PlanningDetailsWidget extends React.Component<IProps, IState> {
     static defaultProps: Partial<IProps>;
     readonly state = {store: null, planning: null};
     private sdPlanningStore: any;
-    private intervalId: NodeJS.Timeout | null = null;
+    private planningId: string | null = null;
+    private unsubscribe: (() => void) | null = null;
 
     constructor(props: IProps) {
         super(props);
-
         this.sdPlanningStore = ng.get('sdPlanningStore');
     }
 
     componentDidMount() {
         const {item} = this.props;
 
-        // Allow the Planning item and store to be loaded concurrently
         getItemPlanningInfo(item).then((planning) => {
+            this.planningId = planning._id;
             this.setState({planning});
         });
 
         this.sdPlanningStore.initWorkspace(WORKSPACE.AUTHORING_WIDGET, (store) => {
-            // Fetch the agendas before saving the store to the state
             store.dispatch(fetchAgendas()).then(() => {
                 this.setState({store});
             });
         });
 
-        this.intervalId = setInterval(() => {
-            this.fetchPlanning(item);
-        }, 10000);
+        const $rootScope = ng.get('$rootScope');
+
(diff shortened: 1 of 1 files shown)


tomaskikutis:
> periodically re-fetches the latest planning item every 10 seconds

can you listen to websocket message instead so refetching every 10 seconds is not needed?

client/components/PlanningDetailsWidget.tsx
```diff
@@ -33,53 +33,48 @@ class PlanningDetailsWidget extends React.Component<IProps, IState> {
...
-            // Fetch the agendas before saving the store to the state
             store.dispatch(fetchAgendas()).then(() => {
                 this.setState({store});
             });
         });
 
-        // Listen to real-time planning updates
-        window.addEventListener('planning:updated', this.onPlanningUpdated as EventListener);
-    }
+        const $rootScope = ng.get('$rootScope');
```
tomaskikutis: use `addWebsocketMessageListener` from superdesk api instead of $rootScope. It should also work to use a generic event `resource:updated` instead of `planning:updated`

client/components/PlanningDetailsWidget.tsx
```diff
@@ -33,53 +33,48 @@ class PlanningDetailsWidget extends React.Component<IProps, IState> {
...
 
-    componentWillUnmount() {
-        window.removeEventListener('planning:updated', this.onPlanningUpdated as EventListener);
-    }
+        this.unsubscribe = $rootScope.$on('planning:updated', (_e: any, updatedPlanning: any) => {
+            const updatedPlanningId = updatedPlanning?.item;
 
-    private loadPlanning = () => {
-        const {item} = this.props;
+            if (!this.planningId || !updatedPlanningId || updatedPlanningId !== this.planningId) {
```
tomaskikutis: do not use `!` operator on non-boolean values

client/components/PlanningDetailsWidget.tsx
```diff
@@ -33,57 +34,53 @@ class PlanningDetailsWidget extends React.Component<IProps, IState> {
...
-            // Fetch the agendas before saving the store to the state
             store.dispatch(fetchAgendas()).then(() => {
                 this.setState({store});
             });
         });
 
-        // Listen to real-time planning updates
-        window.addEventListener('planning:updated', this.onPlanningUpdated as EventListener);
-    }
+        this.unsubscribe = superdeskApi.addWebsocketMessageListener('resource:updated', (event: any) => {
```
tomaskikutis: Specify the type instead of using `any`.

tomaskikutis (approved):

PR #2422, opened 2025-07-27
List item configuration

tomaskikutis:
STT-1191

- Extended existing list item configuration to make all fields configurable. Some were hardcoded.
- Implemented support for configuring a field to render on left or right side of a line
- Made related plannings and featured plannings use new configuration
- Configuration is now done via `superdesk.config.js` instead of a vocabulary as it was before.

Will make another PR to implement more fields that are mentioned in the ticket.

### Front-end checklist


client/components/fields/coverages.tsx
```diff
@@ -1,39 +1,63 @@
...
+import * as selectors from '../../selectors';
+import {CoverageIcons} from '../../components/Coverages/CoverageIcons';
+
+interface IOwnProps extends IFieldsProps {
+    fieldsProps: {
+        coverages?: {
+            /**
+             * Needed when coverages need to be adjusted before being displayed. e.g. filtered
+             */
+            prepare?(coverages: Array<IPlanningCoverageItem>): Array<IPlanningCoverageItem>;
```
petrjasek, line 16: I wonder if `filter` or `beforeRender` would be a better name

petrjasek (approved):

client/components/fields/coverages.tsx
tomaskikutis: Since it's not a predicate function, but is passing and returning the entire array, technically it's not limited to filtering. I could add items or change properties of existing ones for example and it would work fine. Because of this `filter` is not the most accurate.

`beforeRender` I don't like because it gives the impression it'd be related to react component lifecycle.

PR #2427, opened 2025-07-29
Export templates ui

thecalcc:
STT-1232

- Add export templates CRUD UI
- Allow CRUD on backend

### DIFF
--- a/client/components/ExportTemplates/ExportTemplateItem.tsx
+++ b/client/components/ExportTemplates/ExportTemplateItem.tsx
@@ -13,7 +13,7 @@ export class ExportTemplateItem extends React.PureComponent<IPropsGenericFormIte
         return (
             <ListItem
                 onClick={() => page.openPreview(item._id)}
-                data-test-id="internal-destinations-item"
+                data-test-id="export-template-item"
             >
                 <ListItemColumn ellipsisAndGrow noBorder>
                     {getFormFieldPreviewComponent(item, getNameField())}
--- a/client/components/ExportTemplates/ExportTemplateItem.tsx
+++ b/client/components/ExportTemplates/ExportTemplateItem.tsx
@@ -3,7 +3,7 @@ import {IPlanningExportTemplate} from 'interfaces';
 import {IPropsGenericFormItemComponent} from 'superdesk-api';
 import {ListItem, ListItemColumn, ListItemActionsMenu} from 'superdesk-core/scripts/core/components/ListItem';
 import {getFormFieldPreviewComponent} from 'superdesk-core/scripts/core/ui/components/generic-form/form-field';
-import {gettext} from 'utils';
+import {gettext} from '../../utils';
 import {getNameField} from './ManageExportTemplates';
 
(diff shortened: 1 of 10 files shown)


server/planning/planning_export_templates.py
```diff
@@ -28,7 +28,9 @@ class PlanningExportTemplatesResource(superdesk.Resource):
         "label": {"type": "string"},
         "download": {"type": "boolean", "default": False},
     }
-    resource_methods = ["GET"]
+    item_methods = ["PATCH", "DELETE"]
+    resource_methods = ["GET", "POST"]
+    privileges = {"POST": "planning", "PATCH": "planning", "DELETE": "planning"}
```
thecalcc: hi @petrjasek wanted to check with you if that's the way to enable all methods for this resource

petrjasek: you can also add `GET` to `item_methods`. I also wonder what privilege we should use here, probably we should introduce a new one `planning_manage_export_templates`

thecalcc: yep, I was wondering the same about privileges, but then the requirement in the ticket just says only admin should be able to access it. And i made that through the UI. But let me know if I should make it. Is there a way to set the default value for admins to be `true` on the backend? Or on the frontend I can just do `privilege === undefined  && account === 'administrator' ? true : privilege`?

client/components/ExportTemplates/ExportTemplateItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+import {getNameField} from './ManageExportTemplates';
+
+export class ExportTemplateItem extends React.PureComponent<IPropsGenericFormItemComponent<IPlanningExportTemplate>> {
+    render() {
+        const {item, page} = this.props;
+
+        return (
+            <ListItem
+                onClick={() => page.openPreview(item._id)}
+                data-test-id="internal-destinations-item"
```
tomaskikutis: that's incorrect data-test-id

client/components/Main/ActionsSubnavDropdown.tsx
```diff
@@ -12,6 +12,8 @@ import {showModal} from '../../actions/modal';
 import {showModal as showModalSf} from '@sourcefabric/common';
 import {MODALS} from '../../constants/modals';
 import {CoverageProfilesModal} from '../../components/ContentProfiles/CoverageProfileModal';
+import * as selectors from '../../selectors';
+import ManageExportTemplatesModal from '../../components/ExportTemplates/ManageExportTemplates';
```
tomaskikutis: use named exports

tomaskikutis (commented):
I mean to use named export/import for `ManageExportTemplatesModal`. selectors let's keep as is due to amount of usages

tomaskikutis (approved):

server/planning/planning_export_templates.py
petrjasek: we should use privileges here so they can configure it in a way that only admins will be able to use it, no need to make another way to handle that

thecalcc: I added it but not sure if there's anything else to be done related to the DB

petrjasek (approved):

PR #2429, opened 2025-07-30
[SDESK-7723] Fix add to planning item remaining locked

eos87:
### Purpose
To fix the issue of the add to planning item remaining locked after closing the modal or after adding the item to coverage. In this PR we make to unlock the item if it was locked by the "Add to planning" modal in the first place.

### What has changed
- Refactored the `AddToPlanningController` to move unlock logic into a dedicated method and ensure it is called when the modal is closed
- Updated `ModalsContainer` to support an (optional) `onModalHide` callback 
- Adjusted event handling in `index.ts` to delay DOM cleanup to avoid race conditions with certain actions 
  
### Steps to test
Please follow the steps in the ticket comment https://sofab.atlassian.net/browse/SDESK-7723?focusedCommentId=141132

### Front-end checklist

Resolves: [SDESK-7723]

Thanks for checking!

[SDESK-7723]: https://sofab.atlassian.net/browse/SDESK-7723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/controllers/AddToPlanningController.tsx
+++ b/client/controllers/AddToPlanningController.tsx
@@ -107,7 +107,7 @@ export class AddToPlanningController {
 
         ReactDOM.render(
             <Provider store={this.store}>
-                <ModalsContainer />
+                <ModalsContainer onModalHide={this.unlockAddToPlanningItem} />
             </Provider>,
             this.$element
         );
@@ -170,17 +170,23 @@ export class AddToPlanningController {
             }, 1000);
         }
 
-        // Only unlock the item if it was locked when launching this modal
-        if ((this.newsItem?.lock_session ?? null) !== null
-            && (this.newsItem.lock_action ?? 'edit') === 'add_to_planning'
-        ) {
-            this.lock.unlock(this.newsItem);
-        }
+        this.unlockAddToPlanningItem();
 
         if (this.rendered) {
             ReactDOM.unmountComponentAtNode(this.$element.get(0));
         }
     }
+
+    /**
+     * Unlocks the item if it was locked when launching this modal
+     */
+    unlockAddToPlanningItem = () => {
+        const {lock_session, lock_action} = this.newsItem || {};
+
+        if (lock_session && lock_action === 'add_to_planning') {
+            this.lock.unlock(this.newsItem);
+        }
+    };
 
     onItemUnlock(_e, data) {
         if (this.store &&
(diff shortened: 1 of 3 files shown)


index.ts
```diff
@@ -68,16 +68,20 @@ window.addEventListener('planning:fulfilassignment', (event: CustomEvent) => {
...
     const rootScope = ng.get('$rootScope');
+    const timeout = ng.get('$timeout');
 
+    document.body.appendChild(newElement);
     newElement.className = 'modal__dialog ng-scope';
     rootScope.locals = {data: {item: e.detail}};
 
-    // unmount the component and remove the element when the modal is closed
+    // clean up: unmount the React component and remove the DOM element after the modal closes.
+    // delay is added to ensure any pending actions complete before removal.
```
thecalcc: do you use `timeout` instead of `setTimeout` because it also handles UI updates? if `scope.$apply()` isn't needed here, it's better to remove angular wherever possible

Could `resolve` function be called after attempting to unlock locked items? That way we can avoid a timeout all together, which would also fail if action takes longer than 500ms

client/controllers/AddToPlanningController.tsx
```diff
@@ -170,18 +170,24 @@ export class AddToPlanningController {
...
-            this.lock.unlock(this.newsItem);
-        }
+        this.unlockAddToPlanningItem();
 
         if (this.rendered) {
             ReactDOM.unmountComponentAtNode(this.$element.get(0));
         }
     }
 
+    /**
```
thecalcc, line 182: nitpick: comment could be improved by saying something like: "Unlocks all items locked due to `add_to_planning` action or something like that

client/controllers/AddToPlanningController.tsx
```diff
@@ -170,18 +170,24 @@ export class AddToPlanningController {
...
+     * Unlocks the item if it was locked when launching this modal
+     */
+    unlockAddToPlanningItem = () => {
+        const {lock_session, lock_action} = this.newsItem || {};
+
+        if (lock_session && lock_action === 'add_to_planning') {
```
thecalcc: existing constant can be used for `add_to_planning` from `PLANNING.ITEM_ACTIONS.ADD_TO_PLANNING.lock_action`

index.ts
eos87: I found a way to avoid it :) Thanks for the suggestion.

thecalcc (approved):

index.ts
thecalcc: nice one! :-)

PR #2434, opened 2025-08-03
Implement more fields for list item

tomaskikutis:
STT-1191


client/components/fields/anpa_category.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+    const anpa_category = item.anpa_category;
+
+    if ((anpa_category ?? []).length < 1) {
+        return null;
+    }
+
+    return (
+        <Spacer h gap="4" noWrap style={{whiteSpace: 'nowrap'}}>
+            <div className="sd-list-item__text-label">{gettext('ANPA Category')}</div>
```
petrjasek: would be good to use the cv name, I think we do it in authoring, but not yet in planning.

client/globals.d.ts
```diff
@@ -153,11 +153,21 @@ interface ResizeObserverCallback {
...
+export interface ILineConfigStandard {
     fieldId: string;
     position?: 'start' | 'end';
+    fieldOptions: any; // type of options will be different for each field type
 }
 
+export interface ILineConfigVocabulary extends ILineConfigStandard {
+    fieldId: 'vocabulary';
+    fieldOptions: {
+        vocabularyId: string;
```
petrjasek, line 172: might be useful to be able to hide vocabulary name and only show selected values to save some space in the list

petrjasek (approved):

client/components/fields/anpa_category.tsx
tomaskikutis: :+1:

client/globals.d.ts
tomaskikutis: I don't think it's a good idea - it wouldn't be possible to differentiate items from different vocabularies. Maybe it'd be fine if only one vocabulary is configured.

petrjasek (approved):

PR #2441, opened 2025-08-05
[STT-1260] - Fix location details error in the embedded events editor

BrianMwangi21:
### Purpose
Fix location details field integration in planning editor.

### What has changed
- Added logic to ensure location and location.details are saved as needed and don't break the editor

### Steps to test
1. Have Location and Location details fields in the embedded event editor
2. Create a planning item
3. Create an event from the planning item form
4. Fill in the Location in the embedded editor, save the embedded event. Observe that it saves as expected.
5. Fill in the Location details in the embedded event editor, save the embedded event. Observe that it saves as expected.
6. Remove either the location or the location details and save and observe it saves properly.

Resolves: #[STT-1260](https://sofab.atlassian.net/browse/STT-1260)

[STT-1260]: https://sofab.atlassian.net/browse/STT-1260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/planning-extension/src/authoring-react-fields/location/editor.tsx
+++ b/client/planning-extension/src/authoring-react-fields/location/editor.tsx
@@ -6,6 +6,10 @@ import {extensionBridge} from '../../extension_bridge';
 import {cloneDeep, set} from 'lodash';
 
 type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldConfig, ILocationFieldUserPreferences>;
+type FieldValue = ILocation | string | null;
+
+const isLocation = (v: FieldValue): v is ILocation =>
+    typeof v === 'object' && v !== null;
 
 export class Editor extends React.PureComponent<IProps> {
     render() {
@@ -15,17 +19,17 @@ export class Editor extends React.PureComponent<IProps> {
                 ? this.props.value?.[0] ?? null
                 : (this.props.value as any) ?? null;
 
-        const handleChange = (field: string, value: any) => {
+        const handleChange = (field: string, value: FieldValue) => {
             const previousDetails = location?.details;
-            const nextValue: {location: ILocation | null} = {
-                location: cloneDeep(location ?? ({} as ILocation)),
+            const nextValue = {
+                location: location != null ? cloneDeep(location) : null,
             };
 
             set(nextValue, field, value);
 
             if (field === 'location' && previousDetails != null) {
-                const newDetails = (value as ILocation | null)?.details;
-                const hasNewDetails = Array.isArray(newDetails) && newDetails.length > 0;
+                const newDetails = isLocation(value) ? value.details : undefined;
+                const hasNewDetails = (newDetails ?? []).length > 0;
 
                 if (hasNewDetails === false) {
                     nextValue.location = nextValue.location ?? ({} as ILocation);
@@ -33,7 +37,9 @@ export class Editor extends React.PureComponent<IProps> {
                 }
             }
 
-            this.props.onChange([nextValue.location] as ILocationValueOperational);
+            const finalValue = nextValue.location ? [nextValue.location] : [];
+
+            this.props.onChange(finalValue);
         };
 
         return (
--- a/client/planning-extension/src/authoring-react-fields/location/editor.tsx
+++ b/client/planning-extension/src/authoring-react-fields/location/editor.tsx
@@ -25,8 +25,11 @@ export class Editor extends React.PureComponent<IProps> {
                     set(valueCopy, field, value);
 
                     // Preserve location.details if a new location is selected and details are present
-                    if (field === 'location' && currentDetails && !value?.details) {
-                        set(valueCopy, 'location.details', currentDetails);
+                    if (field === 'location' && currentDetails != null && value?.details == null) {
+                        valueCopy.location = {
+                            ...valueCopy.location,
+                            details: currentDetails
+                        };
                     }
                     this.props.onChange([valueCopy.location]);
                 }}
--- a/client/planning-extension/src/authoring-react-fields/location/editor.tsx
+++ b/client/planning-extension/src/authoring-react-fields/location/editor.tsx
@@ -3,7 +3,8 @@ import {IEditorComponentProps} from 'superdesk-api';
 import {ILocationFieldConfig, ILocationFieldUserPreferences, ILocationValueOperational} from './interfaces';
 import {ILocation} from '../../../../interfaces';
 import {extensionBridge} from '../../extension_bridge';
-import {cloneDeep, set} from 'lodash';
+import {set} from 'lodash';
+import {produce} from 'immer';
 
 type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldConfig, ILocationFieldUserPreferences>;
 type LocationOrDetailsValue = ILocation | string | null;
@@ -14,7 +15,7 @@ const isLocationObject = (v: LocationOrDetailsValue): v is ILocation =>
 export class Editor extends React.PureComponent<IProps> {
     render() {
         const {EditorFieldLocation} = extensionBridge.editor.fields;
-        
+
         /*
         * Schema/back-end provide `location` as an array,
         * but the editor supports only a single entry.
@@ -25,23 +26,21 @@ export class Editor extends React.PureComponent<IProps> {
 
         const handleChange = (field: string, value: LocationOrDetailsValue) => {
             const previousDetails = location?.details;
-            const nextValue = {location: cloneDeep(location)};
+            const nextValue = produce({location}, (draft) => {
+                set(draft, field, value);
 
-            set(nextValue, field, value);
+                if (field === 'location' && previousDetails != null) {
+                    const newDetails = isLocationObject(value) ? value.details : undefined;
+                    const hasNewDetails = (newDetails ?? []).length > 0;
 
-            if (field === 'location' && previousDetails != null) {
-                const newDetails = isLocationObject(value) ? value.details : undefined;
-                const hasNewDetails = (newDetails ?? []).length > 0;
-
-                if (hasNewDetails === false) {
-                    nextValue.location = nextValue.location ?? ({} as ILocation);
-                    nextValue.location.details = previousDetails;
+                    if (!hasNewDetails) {
+                        draft.location = draft.location ?? ({} as ILocation);
+                        draft.location.details = previousDetails;
+                    }
                 }
-            }
+            });
 
-            const finalValue = nextValue.location != null ? [nextValue.location] : [];
-
-            this.props.onChange(finalValue);
+            this.props.onChange(nextValue.location ? [nextValue.location] : []);
         };
 
         return (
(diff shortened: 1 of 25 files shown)


client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -2,32 +2,30 @@ import * as React from 'react';
...
-                    this.props.onChange(value);
+                onChange={(field: string, value: any) => {
+                    if (field === 'location') {
+                        this.props.onChange(value);
+                    } else if (field === 'location.details') {
+                        const valueCopy = cloneDeep(this.props.value ?? {});
+
+                        set(valueCopy, 'details', value);
+                        this.props.onChange(valueCopy);
+                    }
```
thecalcc: there is no other case like changing another property e.g. `location.coordinates` where coordinates is an object right?

BrianMwangi21: Yeah, there's no other case. 

The other location properties are handled by the first if since they are handled in the GeoInput.

I put this actually so that it flattens the details cause it was nesting it another layer deep i.e `location.location.details` when it should be location.details as the backend expects and since it it separate from the GeoInput field

thecalcc (approved):

thecalcc (approved):

thecalcc:
Remove package-lock changes, they shouldn't be necessary.

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -9,28 +9,25 @@ type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldCon
...
+                    const currentDetails = location?.details;
+                    const valueCopy = {location: cloneDeep(location ?? {})};
 
                     set(valueCopy, field, value);
 
-                    // Preserve location.details if a new location was selected without details
-                    if (field === 'location' && currentDetails != null) {
-                        valueCopy.location.details = currentDetails;
+                    // Preserve location.details if a new location is selected and details are present
+                    if (field === 'location' && currentDetails && !value?.details) {
```
thecalcc: what's the reason to check `value.details`? If you really should check it, do it explicitly with `!= null`.

Do not convert object to boolean, check it explicitly like this: `currentDetails != null`

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -9,28 +9,25 @@ type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldCon
...
+                        set(valueCopy, 'location.details', currentDetails);
```
thecalcc: Don't use `set` since you already know what will change. Just use `valueCopy.location.details`

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -9,30 +10,40 @@ type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldCon
...
+            const previousDetails = location?.details;
+            const nextValue: {location: ILocation | null} = {
+                location: cloneDeep(location ?? ({} as ILocation)),
+            };
+
+            set(nextValue, field, value);
+
+            if (field === 'location' && previousDetails != null) {
+                const newDetails = (value as ILocation | null)?.details;
+                const hasNewDetails = Array.isArray(newDetails) && newDetails.length > 0;
```
thecalcc: use `(newDetails ?? []).length > 0`

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -9,30 +10,40 @@ type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldCon
...
+        const handleChange = (field: string, value: any) => {
+            const previousDetails = location?.details;
+            const nextValue: {location: ILocation | null} = {
+                location: cloneDeep(location ?? ({} as ILocation)),
+            };
+
+            set(nextValue, field, value);
+
+            if (field === 'location' && previousDetails != null) {
+                const newDetails = (value as ILocation | null)?.details;
```
thecalcc: instead of casting here to `ILocation | null`, add a proper type for `value` prop.

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -9,30 +10,40 @@ type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldCon
...
+                const hasNewDetails = Array.isArray(newDetails) && newDetails.length > 0;
+
+                if (hasNewDetails === false) {
+                    nextValue.location = nextValue.location ?? ({} as ILocation);
```
thecalcc: do not cast

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -9,30 +10,40 @@ type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldCon
...
-        const location = Array.isArray(this.props.value) ? this.props.value[0] : this.props.value;
+        const location: ILocation | null =
+            Array.isArray(this.props.value)
+                ? this.props.value?.[0] ?? null
+                : (this.props.value as any) ?? null;
+
+        const handleChange = (field: string, value: any) => {
+            const previousDetails = location?.details;
+            const nextValue: {location: ILocation | null} = {
+                location: cloneDeep(location ?? ({} as ILocation)),
```
thecalcc: Type of `nextValue` is wrong since you fallback to an object. Instead of casting you can do `location != null ? cloneDeep(location) : null`.

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -9,30 +10,40 @@ type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldCon
...
+                const newDetails = (value as ILocation | null)?.details;
+                const hasNewDetails = Array.isArray(newDetails) && newDetails.length > 0;
+
+                if (hasNewDetails === false) {
+                    nextValue.location = nextValue.location ?? ({} as ILocation);
+                    nextValue.location.details = previousDetails;
+                }
+            }
+
+            this.props.onChange([nextValue.location] as ILocationValueOperational);
```
thecalcc: if value of `nextValue.location` is `ILocation | null` typescript should be able to infer type of `ILocationValueOperational`. If `ILocationValueOperational` does not include `null` you can add it.

thecalcc: also check what happens if nextValue.location is `null` and you send `[null]` to the backend

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -6,6 +6,10 @@ import {extensionBridge} from '../../extension_bridge';
 import {cloneDeep, set} from 'lodash';
 
 type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldConfig, ILocationFieldUserPreferences>;
+type FieldValue = ILocation | string | null;
```
thecalcc: in which case can `FieldValue` be `string`?

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -15,25 +19,27 @@ export class Editor extends React.PureComponent<IProps> {
...
+                const hasNewDetails = (newDetails ?? []).length > 0;
 
                 if (hasNewDetails === false) {
                     nextValue.location = nextValue.location ?? ({} as ILocation);
                     nextValue.location.details = previousDetails;
                 }
             }
 
-            this.props.onChange([nextValue.location] as ILocationValueOperational);
+            const finalValue = nextValue.location ? [nextValue.location] : [];
```
thecalcc: do not implicitly check for `!= null` use `nextValue.location != null`

client/planning-extension/src/authoring-react-fields/location/editor.tsx
thecalcc: Regardless, can you tweak type of `ILocationValueOperational` instead so we don't need another type?

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -6,6 +6,10 @@ import {extensionBridge} from '../../extension_bridge';
 import {cloneDeep, set} from 'lodash';
 
 type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldConfig, ILocationFieldUserPreferences>;
+type FieldValue = ILocation | string | null;
+
+const isLocation = (v: FieldValue): v is ILocation =>
```
thecalcc: rename prop for better readability. 

You can improve the name of the function to be something like `hasLocation` or something along those lines

client/planning-extension/src/authoring-react-fields/location/editor.tsx
BrianMwangi21: It can be string when there's no data. Then location details is put in first. At this point, it first comes as a string before now we insert it in `location.details`

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -8,27 +8,20 @@ import {cloneDeep, set} from 'lodash';
...
-            const nextValue = {
-                location: location != null ? cloneDeep(location) : null,
-            };
+            const nextValue = {location: cloneDeep(location)};
 
             set(nextValue, field, value);
 
             if (field === 'location' && previousDetails != null) {
-                const newDetails = isLocationObject(value) ? value.details : undefined;
+                const newDetails = typeof value === 'object' ? value?.details : undefined;
```
thecalcc: no need to do optional chaining in the first case since you're already checking if `value` is `null` with `typeof value === 'object'`

thecalcc (approved):

client/planning-extension/src/authoring-react-fields/location/editor.tsx
thecalcc: False comment - `typeof null` is `object`, so it wouldn't get caught by just doing `typeof value === 'object'`

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -1,34 +1,50 @@
...
 type IProps = IEditorComponentProps<ILocationValueOperational, ILocationFieldConfig, ILocationFieldUserPreferences>;
+type LocationOrDetailsValue = ILocation | string | null;
+
+const isLocationObject = (v: LocationOrDetailsValue): v is ILocation =>
+    typeof v === 'object' && v !== null;
 
 export class Editor extends React.PureComponent<IProps> {
     render() {
         const {EditorFieldLocation} = extensionBridge.editor.fields;
+        const location: ILocation | null = this.props.value?.[0] ?? this.props.value;
```
tomaskikutis, line 25: when doing something special/unexpected like conditionally getting first item or the entire list, add a comment clarifying what's going on. Maybe default isn't a list at all?

client/planning-extension/src/authoring-react-fields/location/editor.tsx
```diff
@@ -1,34 +1,50 @@
...
+
+        const handleChange = (field: string, value: LocationOrDetailsValue) => {
+            const previousDetails = location?.details;
+            const nextValue = {location: cloneDeep(location)};
```
tomaskikutis: cloning the entire thing will reduce efficiency of react rendering that is comparing by reference - use immer here

tomaskikutis (approved):

PR #2442, opened 2025-08-06
Remove priority from custom vocabulary field registration

thecalcc:
STT-1281

### DIFF
--- a/client/utils/contentProfiles.ts
+++ b/client/utils/contentProfiles.ts
@@ -97,6 +97,11 @@ export const VOCABULARIES_TO_BE_EXCLUDED = new Set([
     'genre',
 
     /**
+     * Fields that are manually registered in `client/components/fields/resources` and use a specific custom vocabulary
+     */
+    'priority',
+
+    /**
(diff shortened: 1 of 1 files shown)


client/utils/contentProfiles.ts
```diff
@@ -96,6 +96,11 @@ export const VOCABULARIES_TO_BE_EXCLUDED = new Set([
     'g2_content_type',
     'genre',
 
+    /**
+     * Fields that are manually registered in `client/components/fields/resources` and use a specific custom vocabulary
```
tomaskikutis, line 100: could you make the comment more clear? I'm not following what was the issue

thecalcc: The comment is not meant to describe the issue, it's meant to say that this section of field ids is meant for fields that are manually registered in the said folder, thus we should prevent them from being registered as a custom vocabulary field also.

But on the issue - it was that we already have a field registered for `priority`, but also `priority` is a custom vocabulary. So I had to remove the custom vocabulary registration.

tomaskikutis (approved):

PR #2444, opened 2025-08-07
Add error message

thecalcc:
STT-1282

### DIFF
--- a/client/planning-extension/src/extension.ts
+++ b/client/planning-extension/src/extension.ts
@@ -114,6 +114,8 @@ function onSendBefore(superdesk: ISuperdesk, items: Array<IArticle>, desk: IDesk
     return Promise.resolve();
 }
 
+const PLACE_FIELD_DEPENDENT_VOCABULARY_ID = 'locators';
+
 const extension: IExtension = {
     activate: (superdesk: ISuperdesk) => {
         const extensionConfig: IPlanningExtensionConfigurationOptions = superdesk.getExtensionConfig();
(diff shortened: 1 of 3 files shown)


client/planning-extension/src/extension.ts
```diff
@@ -269,17 +271,24 @@ const extension: IExtension = {
             },
         };
 
+        const allVocabularies = superdesk.entities.vocabulary.getAll();
+
+        if (allVocabularies.has(PLACE_FIELD_DEPENDENT_VOCABULARY_ID) === false) {
+            result.contributions.getInstanceConfigurationIssues = () => (
```
tomaskikutis, line 277: another approach is to define `getInstanceConfigurationIssues` unconditionally in `result.contributions` and return an empty array if there are no issues.

tomaskikutis (approved):

client/planning-extension/src/extension.ts
thecalcc: yep, can be done, also I thought if we want to add more checks for other fields in the future I can do 

```js
const ERRORS = [];

....contributions: {
...
getInstanceConfigurationIssues: () => Promise.resolve(ERRORS),
}

fieldsDependentOnVocabularies.forEach(({vocabularyId}) => {
ERRORS.push({message: gettext('Vocabulary `{{ vocabularyId }}` is not configured!, {vocabularyId})});
});
```

Though I'm not sure now if this approach would cause a closure, and the function will always return an empty array. If that's the case, I can first fill the `ERRORS` array then define `getInstanceConfigurationIssues` after the for loop.

tomaskikutis: I'm not following why in this example you're declaring `ERRORS` outside `getInstanceConfigurationIssues` fn. You could do all inside the body of the function.

thecalcc: that's also true actually

PR #2452, opened 2025-08-11
[SDESK-7734] - Removed planning items from the event are reverted back when the event is updated

BrianMwangi21:
### Purpose
This PR fixes the incorrect unlinking and syncing behavior between Events and their related Planning items.

### What has changed
  - Unlink only when `embedded_planning` is present in the payload.
  - Preserve unrelated `related_events` links when unlinking.

### Steps to test
1. Create an Event with no Planning items.
2. Add at least 2 Planning items to an Event. Observe that they are linked successfully.
3. Remove one or more Planning items. Observe that only the removed items should be unlinked from the event and the remaining items remain linked.
4. Update Event fields unrelated to related planning e.g event name and observe the planning items remain linked

Resolves: #[SDESK-7734](https://sofab.atlassian.net/browse/SDESK-7734)

[SDESK-7734]: https://sofab.atlassian.net/browse/SDESK-7734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/events/events_sync/__init__.py
+++ b/server/planning/events/events_sync/__init__.py
@@ -19,7 +19,7 @@ from planning.utils import parse_date
 from planning.utils import get_related_planning_for_events
 from planning.content_profiles.utils import AllContentProfileData
 from planning.common import get_config_event_fields_to_sync_with_planning
-from planning.types import Event, EmbeddedPlanningDict, StringFieldTranslation
+from planning.types import Event, EmbeddedPlanningDict, StringFieldTranslation, Planning
 from planning.types.event import EmbeddedPlanning as EmbeddedPlanningModel
 
 from .common import VocabsSyncData, SyncItemData, SyncData
@@ -45,6 +45,7 @@ async def sync_event_metadata_with_planning_items(
     original: Optional[Event],
     updates: Event,
     embedded_planning: list[EmbeddedPlanningDict] | list[EmbeddedPlanningModel],
+    embedded_planning_present: bool,
 ):
     embedded_planning = [
         cast(EmbeddedPlanningDict, obj.to_dict()) if isinstance(obj, EmbeddedPlanningModel) else obj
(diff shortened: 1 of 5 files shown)


server/planning/events/events_sync/__init__.py
```diff
@@ -100,11 +101,42 @@ async def sync_event_metadata_with_planning_items(
...
+
         for planning_original, planning_updates, update_required in get_existing_plannings_from_embedded_planning(
             event_updated, event_translations, embedded_planning, profiles, vocabs
         ):
+            embedded_existing_ids.add(planning_original["_id"])
             if update_required:
-                await planning_service.patch_async(planning_original["_id"], planning_updates)
+                pending_planning_updates.append((planning_original["_id"], planning_updates))
+
+        for planning_id, payload in pending_planning_updates:
```
MarkLark86: I don't see `pending_planning_updates` being used elsewhere.
What was the reason to create a separate list and loop through those, as opposed to directly updating while in the loop

server/planning/events/events_sync/__init__.py
```diff
@@ -100,11 +101,42 @@ async def sync_event_metadata_with_planning_items(
...
+            await planning_service.patch_async(planning_id, payload)
+
+        # 2. Unlink removed planning items
+        if embedded_planning_present:
+            existing_linked_planning_items = get_related_planning_for_events([event_updated["_id"]], "primary")
```
MarkLark86: Should we not send anything for the `link_type` arg to this function, so that we get all linked Planning items?
By sending `primary`, this would remove any Planning items with a `secondary` link

MarkLark86 (commented):
It would be good to have some tests so we don't introduce this regression in the future.

MarkLark86:
@BrianMwangi21 Unfortunately there are tests failing. For the client/e2e, you might try re-creating the package-lock to see if that fixes it.
As for behave, looks like recurring event issue, is this related somehow?

server/planning/events/events_sync/events_sync_tests.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        CASE 1: When original is empty function should call create_new_plannings_from_embedded_planning once
+        and do an early return
+        """
+        original_event = {}  # new event ( will go down the creation path)
+        updated_event = deepcopy(self.get_base_event())
+        embedded_planning = self.build_embedded_planning(updated_event)
+
+        # Mocks for dependencies inside the module under test
+        with patch(
```
MarkLark86: This is a bit hard to read/follow, could you improve on this maybe?

BrianMwangi21:
@MarkLark86 I've simplified the tests and fixed the behave test. It was related. 

I've tried re-creating the package-lock for the e2e tests but they are still failing. Not sure if those are related to my changes though.

MarkLark86 (approved):

PR #2456, opened 2025-08-12
Add a new toggle field 'Multiple Coverages' to the Coverages [STT-1294]

devketanpro:
STT-1294

image

image

### Front-end checklist

Resolves: #[STT-1294]

[STT-1294]: https://sofab.atlassian.net/browse/STT-1294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/content_profiles/profiles/fields.py
+++ b/server/planning/content_profiles/profiles/fields.py
@@ -137,3 +137,16 @@ subjectField = schema.ListField(
         },
     },
 )
+
+
+class MultipleCoveragesField(StringField):
+    def __init__(
+        self,
+        required: bool = False,
+        default_value: bool = False,
+        read_only: bool = False,
+    ):
+        super().__init__(required=required)
+        self.schema["default_value"] = default_value
+        self.schema["read_only"] = read_only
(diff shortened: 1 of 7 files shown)


server/planning/content_profiles/profiles/fields.py
```diff
@@ -137,3 +137,16 @@ def __init__(
         },
     },
 )
+
+
+class MultipleCoveragesField(StringField):
```
MarkLark86: Why does this inherit from a `StringField` instead of the `BooleanField`?

MarkLark86 (commented):
Based on the screenshots provided, I don't see a way to add a default value, this will be required by the customer

thecalcc:
Also support in the embedded editor should be added.

devketanpro:
image

thecalcc:
Hi @MarkLark86 @devketanpro did we test if this field works when it's inside an embedded planning item's coverage?

PR #2458, opened 2025-08-14
[STT-1296] Support linking content to Assignments with multiple_content

MarkLark86:
### Purpose
Now that we have the ability to perform 'Start Working' multiple times on the same Assignment, we need the ability to link multiple content to the 1 Assignment.

### What has changed
* Rename the 'Fulfil Assignment' action to 'Link to Assignment'
* Modify the query for the modal, so we get all Assignment that can have content linked to it
* Fixed icons in list item actions in monitoring (missing `icon-` prefix)

Resolves: [STT-1296](https://sofab.atlassian.net/browse/STT-1296)

[STT-1296]: https://sofab.atlassian.net/browse/STT-1296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/actions/assignments/api.ts
+++ b/client/actions/assignments/api.ts
@@ -164,6 +164,7 @@ const query = ({
     size = null,
     ignoreScheduledUpdates = false,
     max_results = null,
+    baseQuery = null,
 }) => (
     (dispatch, getState, {api}) => {
         const filterByValues = {
@@ -176,10 +177,20 @@ const query = ({
         let sort = '[("' + (get(filterByValues, orderByField, 'planning.scheduled')) + '", '
             + (orderDirection === SORT_DIRECTION.ASCENDING ? 1 : -1) + ')]';
 
-        const baseQuery = selectors.getBaseAssignmentQuery(getState());
+        const baseElasticQuery = selectors.getBaseAssignmentQuery(getState());
+
+        if (baseQuery) {
+            // Combine the elastic queries from the provided baseQuery and the one from the redux store
+            for (const field of ['must', 'must_not', 'should']) {
+                if (baseQuery[field]) {
+                    baseElasticQuery[field] = (baseElasticQuery[field] || []).concat(baseQuery[field]);
+                }
+            }
+        }
+
         const query = constructQuery({
             systemTimezone: appConfig.default_timezone,
-            baseQuery: baseQuery,
+            baseQuery: baseElasticQuery,
             searchQuery: searchQuery,
             deskId: deskId,
             userId: userId,
--- a/client/actions/assignments/api.ts
+++ b/client/actions/assignments/api.ts
@@ -2,7 +2,7 @@ import moment from 'moment-timezone';
 import {get, cloneDeep, has, pick} from 'lodash';
 
 import {appConfig} from 'appConfig';
-import {IAssignmentItem} from '../../interfaces';
+import {IAssignmentItem, ISearchQueryOperator} from '../../interfaces';
 import {planningApi} from '../../superdeskApi';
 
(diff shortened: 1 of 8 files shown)


client/actions/assignments/api.ts
```diff
@@ -176,10 +177,20 @@ const query = ({
         let sort = '[("' + (get(filterByValues, orderByField, 'planning.scheduled')) + '", '
             + (orderDirection === SORT_DIRECTION.ASCENDING ? 1 : -1) + ')]';
 
-        const baseQuery = selectors.getBaseAssignmentQuery(getState());
+        const baseElasticQuery = selectors.getBaseAssignmentQuery(getState());
+
+        if (baseQuery) {
+            // Combine the elastic queries from the provided baseQuery and the one from the redux store
+            for (const field of ['must', 'must_not', 'should']) {
```
thecalcc: can this array of fields be typed or maybe extract to a const?

client/actions/assignments/ui.ts
```diff
@@ -202,13 +202,19 @@ const queryAndSetAssignmentListGroups = (groupKey, page = 1) => (
...
+            querySearchSettings.states = group.states;
+        }
+
         querySearchSettings.page = page;
         querySearchSettings.dateFilter = group.dateFilter;
         querySearchSettings.orderDirection = assignmentListSelectors.sortOrder(getState());
         if (group.max_results) {
             querySearchSettings.max_results = group.max_results;
         }
+        if (group.baseQuery) {
```
thecalcc: can this not be an implicit `null` check?

client/constants/assignments.ts
```diff
@@ -2,6 +2,27 @@ import {superdeskApi} from '../superdeskApi';
...
+    must: [
+        {
+            bool: {
+                should: [
+                    {terms: {'assigned_to.state': ['assigned', 'submitted']}},
+                    {
+                        bool: {
+                            must: [
+                                {term: {'assigned_to.state': 'in_progress'}},
+                                {term: {'planning.multiple_content': true}},
```
thecalcc, line 15: can you type property paths using `nameof` from `superdeskApi` or `Paths` interface from `@sourcefabric/common` and allowed values?

MarkLark86: I have absolutely no idea what you're talking about, sorry!
This is a raw elasticsearch query, and these field names are not dynamic but hard coded into the resource schema in Elasticsearch mappings

client/actions/assignments/api.ts
MarkLark86: Doesn't ts interpret this as an array of strings already?

client/constants/assignments.ts
thecalcc: No worries, it's great if they're already hard coded on the backend, I'm talking about making e.g. `assigned_to.state` typed on the front end, so if a property of `IPlanningCoverageItem` interface changes we have type checks to tell us we should also update the query we send to the server

client/actions/assignments/api.ts
thecalcc: It does, I'm just wondering if it can be more type safe, so you can't add to it e.g. `must_not_should`, since we probably accept only a specific set of strings on the backend

client/constants/assignments.ts
MarkLark86: `assigned_to.state` has not changed since pre v1.0 of Newshub.

thecalcc (approved):

PR #2462, opened 2025-08-19
[STT-1299] Assignment page improvements for multi content linked to Assignments

MarkLark86:
### Purpose
Improve the Assignments page for showing/interacting with multiple content that is linked to an Assignment.

### What has changed
* Update Redux store to store multiple content items per Assignment
* Add expandable list item in Assignment Lists to show all linked content items
* Use toggle boxes to show all linked content items in the preview panel
* Update visuals based on mock ups provided in the ticket

image
image
image
image

### Front-end checklist

Resolves: [STT-1299](https://sofab.atlassian.net/browse/STT-1299)

[STT-1299]: https://sofab.atlassian.net/browse/STT-1299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Archive/ArchivePreview/index.tsx
+++ b/client/components/Archive/ArchivePreview/index.tsx
@@ -1,312 +1,177 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
-import {get, map} from 'lodash';
-import classNames from 'classnames';
 
-import {AuditInformation, HtmlPreview, StateLabel, ItemRendition, PriorityLabel, UrgencyLabel} from '../../';
-import {getCreator} from '../../../utils';
-import * as actions from '../../../actions';
+import {
+    ToggleBox,
+    SimpleList,
+    SimpleListItem,
+    ContentDivider,
+    Heading,
+    PanelContentBlock,
+} from 'superdesk-ui-framework/react';
+
+import {IArticle} from 'superdesk-api';
+import {superdeskApi} from '../../../superdeskApi';
+import {IAssignmentItem} from '../../../interfaces';
 import * as selectors from '../../../selectors';
 
-import './style.scss';
+import {HtmlPreview, ItemRendition} from '../../';
+import {ArchiveItem} from '../ArchiveItem';
+import {AuditInformation} from '../../';
+import {getCreator} from '../../../utils';
 
-class ArchivePreviewComponent extends React.Component {
-    constructor(props) {
-        super(props);
-        this.state = {headerOpen: true};
-        this.toggleHeader = this.toggleHeader.bind(this);
-    }
 
-    componentDidMount() {
-        // When mounting this component, if the Assignment item is set
-        // then load the associated Archive item now
-        const {assignment, loadArchiveItem} = this.props;
+interface IProps {
+    assignment: IAssignmentItem;
+    archiveItems: {[itemId: string]: IArticle};
+    selectedArchiveItemId: IArticle['_id'];
+}
 
-        if (get(assignment, '_id', null) !== null) {
-            loadArchiveItem(assignment);
-        }
-    }
-
-    componentWillReceiveProps(nextProps) {
-        // If the Assignment item has changed, then load the associated Archive item
-        const nextId = get(nextProps, 'assignment._id', null);
-        const currentId = get(this.props, 'assignment._id', null);
-
-        if (nextId !== currentId) {
-            this.props.loadArchiveItem(nextProps.assignment);
-        }
-    }
-
-    toggleHeader() {
-        this.setState({headerOpen: !this.state.headerOpen});
-    }
-
+class ArchivePreviewComponent extends React.PureComponent<IProps> {
     render() {
-        const {archive, users, priorities, urgencies, urgencyLabel} = this.props;
+        const {gettext} = superdeskApi.localization;
+        const {UserAvatar} = superdeskApi.components;
+        const users = Object.values(superdeskApi.entities.users.getAllUsers());
 
-        if (archive === null) {
-            return null;
-        }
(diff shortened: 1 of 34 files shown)


client/components/Archive/ArchivePreview/index.tsx
```diff
@@ -1,312 +1,177 @@
...
+        const users = Object.values(superdeskApi.entities.users.getAllUsers());
+
+        const relatedItems = this.props.assignment.linked_items
+            .map((itemLink) => (this.props.archiveItems[itemLink._id]))
+            .filter((item) => (item != null));
+
+        return relatedItems.map((archive) => (
+            <PanelContentBlock key={archive._id} className="ArchivePreview content-item-preview">
+                <ToggleBox
+                    key={archive._id + '-' + (this.props.selectedArchiveItemId == archive._id).toString()}
```
thecalcc: why do we need a `boolean`.toString() for `key`?

thecalcc: And in general wouldn't this key you put on `PanelContentBlock` be enough?

client/components/Archive/ArchivePreview/index.tsx
```diff
@@ -1,312 +1,177 @@
...
-                    {this.state.headerOpen && (
-                        <div className="ArchivePreview__header-left side-panel__content-block-inner">
-                            <div>
-                                <span
-                                    data-sd-tooltip={`Article Type: ${archive.type}`}
-                                    data-flow="right"
-                                >
-                                    <i className={`coverage-icon icon-${archive.type}`} />
-                                </span>
+                    {(() => {
```
thecalcc: might be better to extract this into a separate small component for better readability, can also be done for the code below this one that's also in a self executing function

client/components/Archive/ArchivePreview/index.tsx
```diff
@@ -1,312 +1,177 @@
...
+                        const createdBy = getCreator(archive, 'original_creator', users);
+                        const updatedBy = getCreator(archive, 'version_creator', users);
+                        const creationDate = archive._created;
+                        const updatedDate = archive._updated;
+                        const versionCreator = updatedBy?.display_name ?
+                            updatedBy :
+                            users.find((user) => user._id === updatedBy);
+
+                        return (
+                            <div className="flex-row p-1 gap-1">
```
thecalcc: Using `Spacer` might be better, so we can continue removing custom class usages and do a better job at UI consistency

client/components/Assignments/AssignmentItem/fields/Content.tsx
```diff
@@ -1,17 +1,29 @@
...
+    assignment: IAssignmentItem;
 }
 
 export const ContentComponent = ({assignment}: IProps) => {
-    const hasContent = assignmentUtils.assignmentHasContent(assignment);
+    const itemEventIds = (assignment.linked_items ?? []).map((item) => item.event_id);
+    const numberOfContent = (new Set(itemEventIds)).size;
 
-    if (!hasContent) {
+    if (!numberOfContent) {
```
thecalcc: could be an explicit check instead of converting number to a boolean

client/components/Archive/ArchivePreview/index.tsx
MarkLark86: Spacer does not strictly follow the [layout principles](https://ui-framework.superdesk.org/#/design/layout-principles) outlines by Fritz, and if a theme changes the grid size from 8px to something else, then all these Spacer usages will not be updated.

thecalcc: I don't think we'll implement such functionality but if we do it would be better to implement it in one place. Why does it not follow the principles -  we have this `gap` property that has `0` | `4` | `8` | `16`, etc. spacing, we could improve it to be based on an 8px number system though which would turn gap options to `0.5` | `1` | `2`, etc.

MarkLark86: These aren't custom classes though, they're ones defined in the UI Framework. These CSS classes are the ones that define the style.

On another point, using classes instead of inline styles are more performant for the browser to implement the style to DOM nodes.

thecalcc: Ok, I did check if we use `flex-row` in other projects, and it looks like it's used about 10 times in legacy angular code and 2 times in UI framework. I think you can use `Spacer` so at least we have a consistent and unified UI, and if we decide to do performance optimisations, design improvements, changing themes, etc. we can do it from a single place - the `Spacer` component. And by custom classes I meant classes defined in UI framework also. 

I guess the least that we need to do is agree to use the spacer, where spacing is needed. Or decide if we'll remove `Spacer` and start using inline classes. Until now we've followed the convention to use the `Spacer` wherever possible. While it's not been all ideal, I don't see a reason not to reuse it here

client/components/Archive/ArchivePreview/index.tsx
MarkLark86: Done, I have moved the key to the PanelContentBlock (I forgot to move it when I added the panel block).
I used `selectedArchiveItemId` in the key for an easy solution for clicking on a list item and controlling the show/hide of these toggle boxes.

client/components/Archive/ArchivePreview/index.tsx
MarkLark86: I reluctantly did used the Spacer element. I believe we need a greater conversation about UI development/styling though.

thecalcc: I agree on the UI conversation. I notice that we're lacking a proper agreed understanding of conventions we use

thecalcc (approved):

PR #2463, opened 2025-08-19
[SDESK-7734] - Planning items not unlinked when removed from Events ContentAPI output

BrianMwangi21:
### Purpose
This PR ensure planning items in the Content API are correctly synced with their related events when added and when removed.
Previously, removed planning items still linked to the event in the database causing stale connections when event is fetched by `add_plan_ids_to_event` function.

### What has changed
- Unlinks stale planning items no longer related to the event by checking the related events from the core and those in the content_api database and doing a comparison to figure out which to unlink.

### Steps to test
1. Create an event with two linked planning items and post it. Check the events contentapi output to see both planning items.
2. Remove one planning item.
3. Check events contentapi output again to see that one planning item is left.
4. Verify in MongoDB too that only the correct planning item remains linked to the events in the `events` key.
5. Confirm the unlinked planning item has no link to the event in the `events` key.

Resolves: Part 2 of #[SDESK-7734](https://sofab.atlassian.net/browse/SDESK-7734)

[SDESK-7734]: https://sofab.atlassian.net/browse/SDESK-7734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/content_api/resources/event.py
+++ b/server/planning/content_api/resources/event.py
@@ -17,9 +17,7 @@ from superdesk.core.resources import (
 from superdesk.core.resources.service import AsyncResourceService
 
 from content_api import MONGO_PREFIX, ELASTIC_PREFIX
-from planning.utils import get_related_planning_for_events
 
-from .planning import ContentAPIPlanningService
 from ..types import ContentAPIEventResource
 from ..output_formatters import ContentApiEventFormatter
 
@@ -43,12 +41,6 @@ class ContentAPIEventService(AsyncResourceService[ContentAPIEventResource]):
             await self.update(event_id, formatted_item)
         else:
             await self.create([formatted_item])
-
-        # Check for linked planning items to the event and update them as well
-        related_planning_item = get_related_planning_for_events([event_id]) or []
-        for planning_item in related_planning_item:
-            await ContentAPIPlanningService().publish_async(planning_item, subscribers)
-
 
 content_api_event_resource_config: ResourceConfig = ResourceConfig(
     name="events_capi",
--- a/server/planning/content_api/resources/event.py
+++ b/server/planning/content_api/resources/event.py
@@ -20,7 +20,7 @@ from content_api import MONGO_PREFIX, ELASTIC_PREFIX
 from planning.utils import get_related_planning_for_events
 
 from .planning import ContentAPIPlanningService
-from ..types import ContentAPIEventResource
+from ..types import ContentAPIEventResource, ContentAPIPlanningResource
 from ..output_formatters import ContentApiEventFormatter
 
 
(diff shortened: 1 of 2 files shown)


server/planning/content_api/resources/event.py
```diff
@@ -44,10 +44,38 @@ async def publish_async(self, item, subscribers=None) -> None:
         else:
             await self.create([formatted_item])
 
-        # Check for linked planning items to the event and update them as well
-        related_planning_items = get_related_planning_for_events([event_id])
-        for planning_item in related_planning_items:
-            await ContentAPIPlanningService().publish_async(planning_item, subscribers)
+        # Get set of planning items linked to this event from core
+        planning_service = ContentAPIPlanningResource.get_service()
```
MarkLark86: This might be causing a race condition. When a Planning item is updated through the Web API, that Planning item should be automatically re-published (to ContentAPI, Newshub and any other subscriber configured).

This happens when [planning_service.patch_async](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/events/events_sync/__init__.py#L220) is called, which then the [planning_service.on_updated_async](https://github.com/superdesk/superdesk-planning/blob/async/server/planning/planning/planning.py#L547) is called where the item should be automatically re-published.

I'm wondering if there's another solution here. In the ContentAPI Event publish service, I'm not sure it's an ideal place to fix this. Might be good to figure out why when updating the Planning it's not getting into the ContentAPI's Planning publish service

server/planning/content_api/resources/event.py
```diff
@@ -45,8 +45,8 @@ async def publish_async(self, item, subscribers=None) -> None:
             await self.create([formatted_item])
 
         # Check for linked planning items to the event and update them as well
-        related_planning_items = get_related_planning_for_events([event_id])
-        for planning_item in related_planning_items:
+        related_planning_item = get_related_planning_for_events([event_id]) or []
```
MarkLark86: I'm not sure we should be touching the ContentAPI Planning items when publishing an Event to ContentAPI.
This should be handled when the Planning item itself is published. This is where I suggested a race condition can happen.

For example:
Core: Used by Web App
CAPI: Used by Content API

Dataflow:
1. Browser: Update published Event, that includes embedded coverages
2. Core[event.update_async]: Update Event in Mongo/Elastic Core databases
3. Core[event.update_async]: run sync_event_metadata_with_planning_items
4. Core[sync_event_metadata_with_planning_items]: [planning.patch_async]
4a. Core[planning.on_updated_async]: Publish Planning item
4b. CAPI[planning.publish_async]: Update Planning item in Mongo/Elastic CAPI databases
5. Core[event.on_updated_async]: Publish Event item
5a. CAPI[event.publish_async]: Update Event item in Mongo/Elastic CAPI databases
5b. CAPI[event.publish_async]: Update Planning item in Mongo/Elastic CAPI databases

As you can see the Planning item is getting updated in CAPI databases twice. This won't be helping any possible bugs, and if there are current bugs they should get fixed in steps 2, 3 or 4.

MarkLark86 (approved):

PR #2481, opened 2025-09-03
add custom fields to coverage schema

petrjasek:
STT-1308 STT-1243

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/server/features/planning.feature
+++ b/server/features/planning.feature
@@ -287,7 +287,10 @@ Feature: Planning
                     "planning": {
                         "ednote": "test coverage, I want 250 words",
                         "headline": "test headline",
-                        "slugline": "test slugline"
+                        "slugline": "test slugline",
+                        "fields": [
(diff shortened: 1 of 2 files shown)


MarkLark86 (commented):
What about the Coverage Profile, where these custom field configs will be stored.
It would be good to have tests, and config storage, for that too

petrjasek:
not sure if we need to modify profiles, we could store those using field id in both editor and schema and that's already allowed, or you had something else in mind?

MarkLark86:
> not sure if we need to modify profiles, we could store those using field id in both editor and schema and that's already allowed, or you had something else in mind?

Would the input be validated, so the correct attributes for the field configs will be provided?

petrjasek:
we're not validating the configs in the profiles https://github.com/superdesk/superdesk-planning/blob/feature/embedded-events-editor-async/server/planning/content_profiles/resource.py#L31-L44

PR #2485, opened 2025-09-04
Handle urgency and priority fields

dzonidoo:
SDESK-7736

### Purpose

improve urgency and priority fields
### What has changed

rendering priority and urgency fileds in embeded form
### Front-end checklist

Resolves: #[issue-number]


tomaskikutis (approved):
fix tests before merging

tomaskikutis:
@petrjasek is this branch still used or should it go straight to async one?

petrjasek:
it should go to async one

e2e/cypress/e2e/planning/edit_planning.cy.ts
```diff
@@ -28,7 +28,7 @@ describe('Planning.Planning: edit metadata', () => {
             ednote: 'Ed. Note',
             anpa_category: ['Domestic Sport'],
             subject: ['sports awards'],
-            urgency: '2',
+            urgency: ['2'],
```
petrjasek: those values shouldn't be lists

tomaskikutis:
image

I looked more carefully at tree select planning wrapper and it looks like it already had the feature to store only the first item of the array when `allowMultiple` is false. It'd be best if you could use that and revert your new stuff (getValue, onSelectionChange).

petrjasek:
I think you still need custom onChange to store only the qcode value and not the whole CV item

tomaskikutis:
You're right that `allowMultiple` wouldn't work since it's not only one vs many, but that we need number only and not an object. Let me look at implementation once more, I think it's not great to have multiple `onChange` functions when we could use only one `onChnage` and have additional converter.

tomaskikutis (approved):
Interesting how it worked for me without `valueAsString` conditional.

petrjasek:
it works in editor, but we have the field also in other place where it actually should set the whole object

PR #2490, opened 2025-09-09
Support custom text fields in coverages

thecalcc:
STT-1309

### DIFF
--- a/client/components/ContentProfiles/FieldTab/FieldEditor.tsx
+++ b/client/components/ContentProfiles/FieldTab/FieldEditor.tsx
@@ -9,7 +9,6 @@ import {
 } from '../../../interfaces';
 
 import {superdeskApi, planningApi} from '../../../superdeskApi';
-import {getFieldNameTranslated} from '../../../utils/contentProfiles';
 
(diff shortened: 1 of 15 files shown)


client/components/ContentProfiles/FieldTab/FieldEditor.tsx
```diff
@@ -148,24 +147,10 @@ export class FieldEditor extends React.Component<IProps, IState> {
...
-                                        {customVocabularies.find((x) => x._id === this.props.item.name).display_name}
-                                        {' '}
-                                        <span className="sd-text--italic sd-text--light">
-                                            {gettext('(custom vocabulary)')}
-                                        </span>
-                                    </>
-                                )
-                                : getFieldNameTranslated(this.props.item.name)
-                            }
-                        </span>
```
petrjasek, line 168: shouldn't this call `getFieldName()` instead?

petrjasek (approved):

client/components/ContentProfiles/FieldTab/FieldEditor.tsx
thecalcc: I talked to fritz, and we decided to remove it, so it's consistent with other such views

PR #2492, opened 2025-09-11
Location for coverages

tomaskikutis:
STT-1317


server/planning/content_profiles/profiles/coverage.py
```diff
@@ -80,6 +80,12 @@ class CoverageSchema(BaseSchema):
             "enabled": False,
             "index": 10,
         },
+
+        # PR-TODO: is this a good approach? How do we make it available for existing profiles?
```
tomaskikutis: @petrjasek how should we handle this?

petrjasek: not sure actually how this works, @thecalcc is this needed here?

thecalcc: i put e.g. `"anpa_category": {"enabled": False}` to make it available in the coverage fields picker. I don't know why you need the index for disabled fields, since if the field is disabled we won't render it at any position. And once you add it, only then an index is assigned to it, based on where you added it.

petrjasek:
I think that failing e2e with filters is actually due those changes:
```
Possibly unhandled rejection: {"data":{"_error":{"code":400,"message":"Insertion failure: 1 document(s) contain(s) error(s)"},"_issues":{"params":{"location":{"dataclass_type":"Input should be a dictionary or an instance of CVItem"}}},"_status":"ERR"},"status":400,"config":{"method":"POST","transformRequest":[null],"transformResponse":[null],"jsonpCallbackParam":"callback","url":"http://localhost:5000/api/events_planning_filters","data":{"name":"Test Events","item_type":"events","params":{"full_text":"searching","name":"Test","slugline":"Original","anpa_category":[{"name":"Overseas Sport","qcode":"s","subject":"15000000","scheme":null}],"subject":[{"qcode":"01001000","name":"archaeology","parent":"01000000","scheme":null},{"qcode":"01011000","name":"music","parent":"01000000","scheme":null ...
```

server/planning/content_profiles/profiles/coverage.py
tomaskikutis: True, index isn't needed. The question is how to make the field possible to configure in coverage profile. I was looking into which fields are allowed to be added to profile are coming from schema. When creating a new profile, default schema would be used - but for profiles that are already created we'd either need an upgrade script or update the list of fields that are available, but I haven't seen such list, it appears it's being taken from schema of a profile.

petrjasek (approved):

PR #2495, opened 2025-09-12
Add custom text field validation fix field changes

thecalcc:
STT-1309

### Purpose
Add validation for custom text fields in coverages. Fix onChange handling after noticing it previously worked, but wasn't accessing and storing field info how it should. 

### What has changed
Introduce validation for custom text fields. Change textArea component to support an additional property that is used later in the change handler for custom text fields.

### DIFF
--- a/client/components/Coverages/CoverageEditor/CoverageForm.tsx
+++ b/client/components/Coverages/CoverageEditor/CoverageForm.tsx
@@ -29,6 +29,7 @@ import {coverageProfiles} from '../../../selectors/coverageProfiles';
 import '../style.scss';
 import {VOCABULARIES_TO_BE_EXCLUDED} from '../../../utils/contentProfiles';
 import {isCustomVocabulary} from '../../../helpers';
+import {notNullOrUndefined} from '@sourcefabric/common';
 
 interface IOwnProps {
     field: string;
@@ -109,6 +110,7 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
         this.onRemoveXmpFile = this.onRemoveXmpFile.bind(this);
         this.onContentTypeChange = this.onContentTypeChange.bind(this);
         this.toggleAddToWorkflow = this.toggleAddToWorkflow.bind(this);
+        this.onCustomTextFieldChange = this.onCustomTextFieldChange.bind(this);
         this.onAnpaCategoryChange = this.onAnpaCategoryChange.bind(this);
 
         this.dom = {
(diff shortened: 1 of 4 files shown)


client/components/UI/Form/TextArea.tsx
```diff
@@ -20,22 +42,22 @@ export class TextArea extends React.Component {
...
         if (this.props.initialFocus) {
-            this.dom.input.focus();
+            this.dom.input?.focus();
         }
     }
 
-    componentWillReceiveProps(nextProps) {
-        if (this.props.autoHeight && nextProps.value !== this.props.value) {
-            this.delayedResize(nextProps.value);
+    componentWillReceiveProps(nextProps: ITextAreaProps) {
```
petrjasek, line 54: wouldn't this be better in `componentDidUpdate` ?

client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -133,36 +135,27 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
...
+                `coverages.${index}.planning.fields`,
+                [{field: fieldId, value: value}]
             );
         } else {
             onChange(
-                `coverages.${index}.${field}`,
-                value
+                // remove value from change path, because even though fields has items,
+                // we may not yet have this field in the array
+                `coverages.${index}.${fieldPath.replace('.value', '')}`,
```
petrjasek, line 157: what's in the fieldPath? I wonder how that works

client/components/UI/Form/TextArea.tsx
thecalcc: I gotta check it, didn't look at what it does, just added types

client/components/Coverages/CoverageEditor/CoverageForm.tsx
thecalcc: yeah that fieldPath is used to access the value inside the component. So for `fields` that would be something like `planning.fields[0].value` so full change path would be e.g. `coverages.[0].planning.fields[0].value`. And so to keep it consistent in here I just strip `.value` from the fieldPath and replace the whole object for the given field

thecalcc: So this fieldPath is essentially a string and is sent along with the change, and then applied with lodash's `set`

petrjasek (approved):

thecalcc:
closing in favour of https://github.com/superdesk/superdesk-planning/pull/2498/files

PR #2505, opened 2025-09-17
Improve coverage add to workflow handling

thecalcc:
SDESK-7772

### Purpose
Improve the interaction for adding a coverage to workflow and canceling it. 

### What has changed
Since the add to workflow field always comes enabled from the backend and is always needed, it was removed from the coverage profile config. The order of fields also got changed, so that add to workflow toggle is always the first profile field in the editor, whenever visible. 

The toggle is now visible only if adding coverages to workflow is manual on the instance. The toggle now allows users to cancel a coverage, if it was previously added to workflow also only on manual instances. 

If adding coverages to workflow is configured to be automatic, the toggle is hidden from the label, and the "Cancel" coverage action is shown in the three dots menu of a coverage. 

### Steps to test
Set `planning_auto_assign_to_workflow` to `true`. You should now see only a "Cancel" action in the three dots menu of a coverage. 

Set `planning_auto_assign_to_workflow` to `false`. You should now see the "Add to workflow" toggle as the first field in the coverage editor, regardless if you made changes to the coverage profile or not. After you assign the coverage, you should be able to add it to workflow. After saving the whole planning item, you should be able to "Cancel" the coverage by using the toggle off. Once you save the planning item, you should not be able to add it to workflow again. 

### Screenshots

Field, no longer in the profile config:
image

---

Adding to workflow manually:
image

---

Cancelling a coverage in workflow:

image
--- 
image
---
image

### DIFF
--- a/client/components/ContentProfiles/CoverageProfileModal.tsx
+++ b/client/components/ContentProfiles/CoverageProfileModal.tsx
@@ -170,10 +170,14 @@ export class CoverageProfilesModal extends React.Component<IProps, IState> {
     updateFields(fields: Array<IProfileFieldEntry>) {
         const profileCloned = cloneDeep(this.state.profile);
 
+        // Make sure `add_to_workflow is always first so
+        // it renders first in the editor
+        profileCloned.editor.add_coverage_to_workflow.index = 0;
+
         fields.forEach((item, index) => {
             profileCloned.editor[item.name] = {...item.field};
             profileCloned.schema[item.name] = {...item.schema};
-            profileCloned.editor[item.name].index = index;
+            profileCloned.editor[item.name].index = index + 1;
         });
 
         this.setState({
--- a/client/components/ContentProfiles/FieldTab/index.tsx
+++ b/client/components/ContentProfiles/FieldTab/index.tsx
@@ -269,7 +269,9 @@ export class FieldTab extends React.Component<IProps, IState> {
                             <FieldList
                                 profile={this.props.profile}
                                 group={undefined}
-                                fields={getGroupFieldsSorted(this.props.profile)}
+                                fields={getGroupFieldsSorted(this.props.profile)
+                                    .filter((field) => field.name !== 'add_coverage_to_workflow')
+                                }
                                 unusedFields={unusedFields}
                                 systemRequiredFields={this.props.systemRequiredFields}
                                 onSortChange={this.updateFieldOrder}
@@ -289,7 +291,9 @@ export class FieldTab extends React.Component<IProps, IState> {
                                     key={group._id}
                                     profile={this.props.profile}
                                     group={group}
-                                    fields={getGroupFieldsSorted(this.props.profile, group._id)}
+                                    fields={getGroupFieldsSorted(this.props.profile, group._id)
+                                        .filter((field) => field.name !== 'add_coverage_to_workflow')
+                                    }
                                     unusedFields={unusedFields}
                                     onSortChange={this.updateFieldOrder}
                                     insertField={this.insertField}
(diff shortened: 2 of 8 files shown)


client/components/ContentProfiles/CoverageProfileModal.tsx
```diff
@@ -170,10 +170,14 @@ export class CoverageProfilesModal extends React.Component<IProps, IState> {
...
 
+        // Make sure `add_to_workflow is always first so
+        // it renders first in the editor
+        profileCloned.editor.add_coverage_to_workflow.index = 0;
+
         fields.forEach((item, index) => {
             profileCloned.editor[item.name] = {...item.field};
             profileCloned.schema[item.name] = {...item.schema};
-            profileCloned.editor[item.name].index = index;
+            profileCloned.editor[item.name].index = index + 1;
```
tomaskikutis, line 180: would be good to add a comment here too that you're moving all fields "add to coverage" is first.

client/components/Coverages/CoverageEditor/index.tsx
```diff
@@ -248,7 +249,13 @@ export class CoverageEditorComponent extends React.PureComponent<IProps> {
                 },
             ];
 
-            if (planningUtils.canCancelCoverage(value, diff)) {
+            if (
+                planningUtils.canCancelCoverage(value, diff)
+
+                // Show only if adding coverages to workflow is handled automatically
+                // thus toggle in the editor is hidden
+                && appConfig.planning_auto_assign_to_workflow === true
```
tomaskikutis, line 257: Is there a reason not to move this to `planning_auto_assign_to_workflow` function?

client/components/ContentProfiles/FieldTab/index.tsx
```diff
@@ -269,7 +269,9 @@ export class FieldTab extends React.Component<IProps, IState> {
                             <FieldList
                                 profile={this.props.profile}
                                 group={undefined}
-                                fields={getGroupFieldsSorted(this.props.profile)}
+                                fields={getGroupFieldsSorted(this.props.profile)
+                                    .filter((field) => field.name !== 'add_coverage_to_workflow')
```
tomaskikutis, line 273: wouldn't it be better to move filtering inside `getGroupFieldsSorted`?

thecalcc: no, because that was is also used when rendering fields inside the editor. I want to only remove this field from being shown in the profile, thus I decided to filter out the field here

client/components/ContentProfiles/CoverageProfileModal.tsx
thecalcc: isn't it obvious that if `add_coverage_to_workflow` is always first, then the rest of the fields should follow? and if `index` starts at `0` it's overlapping with the index I manually set on `add_coverage_to_workflow`

client/components/Coverages/CoverageEditor/index.tsx
thecalcc: Not sure I get this one. Are you asking if I could move it inside `canCancelCoverage`? If that's the case - I won't move it because it's used in other places. The purpose of the check here is to decide whether generally a coverage can be canceled, and then follow the new requirement - if `planning_auto_assign_to_workflow === true` we hide the button. That shouldn't be tied to the actual logic that says whether you can cancel a button.

client/components/ContentProfiles/CoverageProfileModal.tsx
tomaskikutis: it's obvious enough when looking at this diff, but if I opened the entire file without seeing the changed parts then it's less obvious that the 2 are connected.

client/components/Coverages/CoverageEditor/index.tsx
tomaskikutis: Yes I meant `canCancelCoverage` - pasted the wrong thing.

tomaskikutis (approved):

PR #2508, opened 2025-09-18
[STT-1205] Add Search Panel to Assignments page

MarkLark86:
### Purpose
Add search panel to the Assignments page.

### What has changed
* Move Assignment elasticsearch query from front-end to back-end
* Add Assignment search profile to `planning_types.advanced_search` document
* Re-use the same filter panel From Planning in the Assignments page

### Screenshots

image

image

### Front-end checklist

Resolves: [STT-1205](https://sofab.atlassian.net/browse/STT-1205)

[STT-1205]: https://sofab.atlassian.net/browse/STT-1205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/components/Assignments/AssignmentFilterPanel.tsx
```diff
@@ -0,0 +1,140 @@
...
+            set(params, field, null);
+        } else {
+            set(params, field, value);
+        }
+
+        this.setState({params: params});
+    }
+
+    onChangeMultiple(updates: IAssignmentSearchParams) {
+        const params = cloneDeep(this.state.params);
```
tomaskikutis: consider using immer where mutation is needed

client/components/Assignments/AssignmentFilterPanel.tsx
```diff
@@ -0,0 +1,140 @@
...
+                                onChange={this.onChangeHandler}
+                                onChangeMultiple={this.onChangeMultiple}
+                                searchProfile={this.props.searchProfile}
+                                enabledField="enabled"
+                                filterUsersByDesk={this.props.currentDeskId}
+                            />
+                        </ContentBlockInner>
+                    </ContentBlock>
+                </Content>
+                <Footer className="side-panel__footer--button-box">
```
tomaskikutis, line 122: convert `Footer` prop-types to typescript intefaces - it is currently showing a compilation error `Type 'Element' is not assignable to type 'ReactNodeLike'.`

client/components/Assignments/AssignmentFilterPanel.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    <h3 className="side-panel__heading">
+                        {gettext('Advanced filters')}
+                    </h3>
+                </Header>
+                <Content>
+                    <ContentBlock>
+                        <ContentBlockInner>
+                            <AdvancedSearch
+                                params={this.state.params}
+                                onChange={this.onChangeHandler}
```
tomaskikutis, line 113: this doesn't compile

server/features/search_assignments.feature
```diff
@@ -0,0 +1,627 @@
...
+        When we get "/events_planning_search?repo=assignments&subject=soccer,locality:international"
+        Then we get list with 1 items
+        """
+        {"_items": [
+            {"_id": "#ASSIGNMENT_1_ID#"}
+        ]}
+        """
+
+        # Custom Texts
+        When we get "/events_planning_search?repo=assignments&custom_text=my_name:Elephants"
```
petrjasek, line 508: how would it work when searching multiple fields possible using multiple words in each at the same time?

server/features/search_assignments.feature
```diff
@@ -0,0 +1,627 @@
...
+        Then we get list with 1 items
+        """
+        {"_items": [
+            {"_id": "#ASSIGNMENT_1_ID#"}
+        ]}
+        """
+
+    @auth @wip
```
petrjasek: extra `@wip`

server/planning/search/queries/assignments.py
```diff
@@ -0,0 +1,285 @@
...
+        query.must.append(
+            elastic.terms(
+                field="assigned_to.desk",
+                values=desk_ids,
+            )
+        )
+
+
+def search_users(params: dict, query: elastic.ElasticQuery) -> None:
+    desk_ids = common.str_to_array(params.get("user_ids"))
```
petrjasek: here it should be rather `user_ids`

client/components/AdvancedSearch/index.tsx
```diff
@@ -102,6 +142,29 @@ export class AdvancedSearch extends React.PureComponent<IProps> {
...
+                genre: {
+                    clearable: true,
+                },
+                assignment_priority: {
+                    field: 'assignmentPriority',
+                    noTitleInPopup: true,
+                    clearable: true,
+                },
+                anpa_category: {
+                    field: this.props.type === 'assignments' ? 'anpaCategory' : 'anpa_category',
```
tomaskikutis, line 166: In many places you use a ternary where it should be a switch instead - even better if in combination with `assertNever`. Or multiple `if` statements [like here](https://github.com/superdesk/superdesk-planning/blob/feature/embedded-events-editor-async/client/components/editor-standalone/authoring-storage-in-memory.ts#L14).

At the moment it's not causing issues since there are only 2 possible string values, but adding more is much easier when using `assertNever` since one only needs to follow compiler errors to find all places that need to be updated.

tomaskikutis (approved):

server/features/search_assignments.feature
MarkLark86: It uses a comma delimiter to split field/value queries. This does mean the search value does not support `,` in the search text. If this is an issue, we could use an escape sequence to support this.

I have added more tests for this scenario, with 2 planning items with:
```json
"fields": [
    {"field": "my_name", "value": "Elephants being creative"},
    {"field": "my_description", "value": "Can paint cats dancing"}
]
```
and
```json
"fields": [
    {"field": "my_name", "value": "Elephants singing"},
    {"field": "my_description", "value": "Can paint cats singing"}
]
```

https://github.com/superdesk/superdesk-planning/blob/bead77075f2984384780d95fbee36a10b01197d1/server/features/search_assignments.feature#L516-L544

petrjasek (approved):

PR #2511, opened 2025-09-18
show remote timezone details inside tooltip

tomaskikutis:
SDESK-7766

- [ ] update client-core to receive ui-framework update https://github.com/superdesk/superdesk-client-core/pull/4952

image

### DIFF
--- a/client/components/Events/EventDateTime.tsx
+++ b/client/components/Events/EventDateTime.tsx
@@ -8,9 +8,8 @@ import {eventUtils, timeUtils} from '../../utils';
 import {DateTime} from '../UI';
 
 import './style.scss';
-import {Spacer} from 'superdesk-ui-framework/react';
+import {Icon, Spacer, Tooltip} from 'superdesk-ui-framework/react';
 import {isSameDay} from './../../helpers';
-import {SpacerBlock} from '@sourcefabric/common';
 
 interface IProps {
     item: IEventItem;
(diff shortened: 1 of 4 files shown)


client/components/UI/DateTime.tsx
```diff
@@ -79,8 +81,19 @@ function DateTime({
     const displayDate = !isEndEventDateTime ? eventStartDate : eventEndDate;
     const tz = timeUtils.getTimeZoneAbbreviation(date.format('z')) + ' ';
 
+    const style: React.CSSProperties = {color};
+
+    if (color != null) {
+        style.color = color;
+    }
```
thecalcc: needs a fix

client/components/Events/EventDateTime.tsx
```diff
@@ -97,32 +96,43 @@ export class EventDateTime extends React.PureComponent<IProps> {
...
-                                withYear={remoteEndWithYear}
-                                date={remoteEnd}
-                                withTime={!isFullDay}
-                                isEndEventDateTime={true}
-                                {...commonProps}
-                            />
-                        </span>
-                    )}
+                    {
+                        isRemoteTimeZone && (
```
thecalcc: wouldn't it be better to extract this tooltip functionality into its own component so this one is easier to read?

tomaskikutis: There are lots of props I'd need to copy around, so I'd rather keep it in this component. I've moved tooltip content into a separate variable so return content is quicker to scan

thecalcc (approved):
ci needs a fix

PR #2516, opened 2025-09-23
Update store after change

thecalcc:
STT-1333

**Purpose:** 
Keep track of changes in the `planning_export_templates` resource. That way views using export templates can have the latest data without re-fetching. 

**What's changed:**
- Enhanced the redux store to also have the ability to get updated data. 
- Removed the template filtering in the export templates modal. 
- Added an event listener that calls a store update once it receives an update.

### DIFF
--- a/client/actions/exportTemplates.ts
+++ b/client/actions/exportTemplates.ts
@@ -1,6 +1,7 @@
+import {ExportTemplatesActions} from '../constants/exportTemplates';
 import {IPlanningExportTemplate} from '../interfaces';
 
 export const updateTemplates = (updatedTemplates: Array<IPlanningExportTemplate>) => ({
-    type: 'UPDATE_EXPORT_TEMPLATES',
+    type: ExportTemplatesActions.UPDATE_EXPORT_TEMPLATES,
     payload: updatedTemplates,
 });
--- a/client/actions/exportTemplates.ts
+++ b/client/actions/exportTemplates.ts
@@ -1,0 +1,6 @@
+import {IPlanningExportTemplate} from '../interfaces';
+
+export const updateTemplates = (updatedTemplates: Array<IPlanningExportTemplate>) => ({
+    type: 'UPDATE_EXPORT_TEMPLATES',
+    payload: updatedTemplates,
+});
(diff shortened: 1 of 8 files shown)


client/actions/exportTemplates.ts
```diff
@@ -0,0 +1,6 @@
+import {IPlanningExportTemplate} from '../interfaces';
+
+export const updateTemplates = (updatedTemplates: Array<IPlanningExportTemplate>) => ({
+    type: 'UPDATE_EXPORT_TEMPLATES',
```
eos87: We could use an Enum here instead of hardcoding the same string twice. Thoughts?
```
    type: 'UPDATE_EXPORT_TEMPLATES',
```

eos87 (commented):
I only have one comment. It looks good. I would only ask you to give some love to the PR description :)

client/actions/exportTemplates.ts
thecalcc: yep, makes total sense

MarkLark86:
@thecalcc this won't work with multi users. Other users would still need to reload the page right?

eos87 (approved):

MarkLark86:
We use [notifications](https://github.com/superdesk/superdesk-planning/blob/develop/client/actions/notifications.ts) file to listen to websocket notifications, and update the redux store when certain resources are updated

client/actions/multiSelect.ts
```diff
@@ -294,7 +296,7 @@ const exportAsArticle = (items = [], download) => (
                 items: sortableItems,
                 action: exportArticlesDispatch,
                 desks: [...selectors.general.userDesks(getState()), personalWorkspace],
-                templates: templates.filter((t) => download ? t.download : !t.download),
+                templates: templates,
```
MarkLark86: Why this change? Some templates are for download only, while others are for exporting as articles

MarkLark86: See https://github.com/superdesk/superdesk-aap/blob/master/server/data/planning_export_templates.json#L20

thecalcc: I see, I didn't ask what the requirements were when I removed it, I've reverted it with a comment that explains what it's meant for

MarkLark86 (approved):

PR #2518, opened 2025-09-24
[SDESK-7774] - Show event schedule summary for all event types

thecalcc:
SDESK-7774

### Purpose
Show event schedule summary after event was created.

### What has changed
Show event schedule summary component regardless of the instance's link config. 

### Steps to test
Create an event. 
You can configure it to be recurring to test if full details render.
Event summary should be rendered at the top of the event after saving it. 

Resolves: #[SDESK-7774](https://sofab.atlassian.net/browse/SDESK-7774)

@MarkLark86 let me know if this needs changes, I'm not sure what the logic behind hiding it was when the changes were made, but now it'll show up always.

[SDESK-7774]: https://sofab.atlassian.net/browse/SDESK-7774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Events/EventEditor/index.tsx
+++ b/client/components/Events/EventEditor/index.tsx
@@ -105,10 +105,6 @@ class EventEditorComponent extends React.PureComponent<IProps> {
     }
 
     renderHeader() {
-        if (appConfig.planning_event_link_method === 'many_secondary') {
-            return null;
-        }
-
         return !this.props.itemExists ? null : (
             <React.Fragment>
                 <EventEditorHeader item={this.props.item} />
--- a/client/components/Events/EventEditor/index.tsx
+++ b/client/components/Events/EventEditor/index.tsx
@@ -17,6 +17,7 @@ import {EditorForm} from '../../Editor/EditorForm';
 import {EventEditorHeader} from './EventEditorHeader';
 import {ContentBlock} from '../../UI/SidePanel';
 import {RepeatEventSummary} from '../RepeatEventSummary';
+import {EventScheduleSummary} from '../EventScheduleSummary';
 import {appConfig} from 'appConfig';
 
 interface IProps {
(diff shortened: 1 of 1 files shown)


MarkLark86 (commented):
Doesn't this also show the event schedule summary as well as the date inputs?
Seems redundant to me to show both the schedule summary and date inputes.

I think the idea is to show the recurring summary. https://github.com/superdesk/superdesk-planning/blob/686189c46d5e78751ab8c5d942b61b91a27c1902/client/components/Events/EventScheduleSummary/index.tsx#L36

thecalcc:
> Doesn't this also show the event schedule summary as well as the date inputs? Seems redundant to me to show both the schedule summary and date inputes.
> 
> I think the idea is to show the recurring summary.
> 
> https://github.com/superdesk/superdesk-planning/blob/686189c46d5e78751ab8c5d942b61b91a27c1902/client/components/Events/EventScheduleSummary/index.tsx#L36

yep you're right, I talked to Gideon, since we now allow users to edit the date after the event was created we should show just the recurring summary if the event is recurring

client/components/Events/EventEditor/index.tsx
```diff
@@ -105,19 +105,20 @@ class EventEditorComponent extends React.PureComponent<IProps> {
     }
 
     renderHeader() {
-        if (appConfig.planning_event_link_method === 'many_secondary') {
```
MarkLark86, line 108: The default system configuration doesn't allow editing of the Event date or time once it's created.
So these changes should should be linked to the `many_secondary` config value.
(sorry I missed that one yesterday)

MarkLark86 (approved):

PR #2521, opened 2025-09-25
Improve assignment related list item fields

thecalcc:
SDESK-7778

### Purpose
Improve the assignment list item.

### What has changed
Changed a few styles, adding ellipsis mostly to text fields, flexbox adjustments, avatar and tag sizes were made consistent with the planning/events list item.


eos87 (approved):
Looks good

PR #2524, opened 2025-09-29
[STT-1343] - Use multiple content field conifg in coverage editor

thecalcc:
### Purpose
Use config options from profile for `multiple_content` coverage field. So if `read_only` is `true`, you won't be able to interact with the field through the editor.

### What has changed
Get the full coverage profile, read the config value and pass it to the field.

### Steps to test
1. Configure a coverage profile to have Multiple content field. Set that field to `read_only = true` using the field's side panel. 
2. Open a coverage with this profile.
3. Coverage `multiple_content` field should be disabled.

Resolves: #[STT-1343](https://sofab.atlassian.net/browse/STT-1343)

[STT-1343]: https://sofab.atlassian.net/browse/STT-1343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Coverages/CoverageEditor/CoverageForm.tsx
+++ b/client/components/Coverages/CoverageEditor/CoverageForm.tsx
@@ -498,7 +498,7 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
             },
             multiple_content: {
                 disabled: this.props.readOnly
-                    ?? (coverageProfile?.schema?.['multiple_content'] as IProfileSchemaTypeList).read_only
+                    ?? (coverageProfile?.schema?.['multiple_content'] as IProfileSchemaTypeList)?.read_only
                     ?? false,
                 field: 'multiple_content',
             },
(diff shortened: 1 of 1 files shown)


client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -419,6 +420,9 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
                 },
             }), {});
 
+        const allProfiles = coverageProfiles(planningApi.redux.store.getState());
```
eos87, line 423: I wonder how could we grab `allProfiles` (planningApi) from the redux mapStateToProps to avoid coupling the component like how we have right now. I've seen this a couple times in other components.

eos87 (approved):

client/components/Coverages/CoverageEditor/CoverageForm.tsx
thecalcc: we can do it for sure, just need someone to decide whether we'll continue to use redux or not, or whether we'll switch to a different library for state management

PR #2531, opened 2025-10-02
Fix cv select not allowing selection of a entire category when cv config allows it

tomaskikutis:
STT-1351

The issue was that `selectBranchWithChildren=true` means `allow selecting`, but `cv.disable_entire_category_selection` was being passed it it.


tomaskikutis:
I didn't mean to change ui-framework version - that was a mistake.

thecalcc (approved):

PR #2539, opened 2025-10-06
[SDESK-7782] - Allow existing non recurring events to be made recurrent

thecalcc:
### Purpose
Allow existing non recurring events to be made into recurring ones.

### What has changed
Event Recurring Rules input. Now if the event is saved but not recurring you can reconfigure it to be one. 

### Steps to test
Case 1:

Open an existing recurring event. 
Recurring summary should be showing in place of the recurring event inline toggle and input.

Case 2:
Open a new event. Save it without making it a recurring one.
Open it again -> you should see the recurring events toggle.
Toggle it on, and configure the recurring settings.
Event should now be turned into a recurring one.

Resolves: #[SDESK-7782](https://sofab.atlassian.net/browse/SDESK-7782)

image
image
image
image
image
image

[SDESK-7782]: https://sofab.atlassian.net/browse/SDESK-7782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Events/EventEditor/index.tsx
+++ b/client/components/Events/EventEditor/index.tsx
@@ -112,30 +112,15 @@ class EventEditorComponent extends React.PureComponent<IProps> {
             return null;
         }
 
-        const eventSchedule = item.dates ?? {};
-        const doesRepeat = eventSchedule.recurring_rule != null;
-        const isManySecondary = appConfig.planning_event_link_method === 'many_secondary';
-
         return (
             <>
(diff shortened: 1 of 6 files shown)


eos87 (approved):
Looks good. I updated the title so it would sound better in release change log :)

PR #2542, opened 2025-10-07
Remove add to workflow from coverage profile

thecalcc:
### Purpose
Remove add to workflow field from coverage profile, because we don't need it; also because existing instances might get in a bad state because we no longer allow users to configure it but was a system required field

Resolves: #[SDESK-7781](https://sofab.atlassian.net/browse/SDESK-7781)

[SDESK-7781]: https://sofab.atlassian.net/browse/SDESK-7781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


petrjasek (approved):

thecalcc:
hi @petrjasek could you check backend test failing?

PR #2555, opened 2025-10-14
[STT-1397] - Assignments view: Filtering by ANPA category (Osasto)

BrianMwangi21:
### Purpose
This PR adds some changes to the planning -> coverage flow whereby now : 
- Coverages now inherit the Planning’s category if none is provided explicitly . They can both be different.
- Newly created stories from assignments inherit from the coverage first then fall back to the planning item if coverage is null.

### What has changed
- In the planning service, added logic so a coverage without its own `anpa_category` now inherits from the parent planning item.
- Updated merge logic to pull `anpa_category` from the coverage if it set before falling back to the planning item’s category.

### Steps to test
1. Create a Planning item and set its anpa category.
2. Add a Coverage to the planning item without anpa category. 
3. In the Assignments list, filter by that category → the assignment must appear. It should also show the anpa category inherited from the planning item in the list view.
4. Click “Start working” on that assignment → the new story’s anpa category should match the Coverage’s one.
6. Repeat with a Coverage that has its own anpa category. Observe it retains its own category even if different from the planning item. Filtering on Assignments list should also work. Also story must inherit the Coverage’s category as expected.

Resolves: #[STT-1379](https://sofab.atlassian.net/browse/STT-1379)

[STT-1379]: https://sofab.atlassian.net/browse/STT-1379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/assignments/assignments_content.py
+++ b/server/planning/assignments/assignments_content.py
@@ -94,7 +94,14 @@ async def get_item_from_assignment(assignment, template=None):
 
                 await merge_subject(item, planning)
                 merge_list("place", item, planning)
-                merge_list("anpa_category", item, planning)
+
+                # Inherit anpa_category from coverage first, then fallback to planning item
+                coverage = await get_coverage_for_assignment(assignment) or {}
+                coverage_anpa_category = (coverage.get("planning") or {}).get("anpa_category")
+                if coverage_anpa_category:
+                    item["anpa_category"] = deepcopy(coverage_anpa_category)
+                else:
+                    merge_list("anpa_category", item, planning)
 
             if assignment.get("description_text"):
                 item["abstract"] = "<p>{}</p>".format(assignment["description_text"])
--- a/server/planning/assignments/assignments_content.py
+++ b/server/planning/assignments/assignments_content.py
@@ -95,11 +95,9 @@ async def get_item_from_assignment(assignment, template=None):
                 await merge_subject(item, planning)
                 merge_list("place", item, planning)
 
-                # Inherit anpa_category from coverage first, then fallback to planning item
-                coverage = await get_coverage_for_assignment(assignment) or {}
-                coverage_anpa_category = (coverage.get("planning") or {}).get("anpa_category")
-                if coverage_anpa_category:
-                    item["anpa_category"] = deepcopy(coverage_anpa_category)
+                anpa_category = planning_data.get("anpa_category")
+                if anpa_category:
+                    item["anpa_category"] = deepcopy(anpa_category)
                 else:
                     merge_list("anpa_category", item, planning)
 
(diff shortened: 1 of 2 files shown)


server/planning/assignments/assignments_content.py
```diff
@@ -94,7 +94,12 @@ async def get_item_from_assignment(assignment, template=None):
 
                 await merge_subject(item, planning)
                 merge_list("place", item, planning)
-                merge_list("anpa_category", item, planning)
+
+                anpa_category = planning_data.get("anpa_category")
+                if anpa_category:
+                    item["anpa_category"] = deepcopy(anpa_category)
+                else:
+                    merge_list("anpa_category", item, planning)
```
eos87: nit: you can take advantage of the walrus operator here 😁 
```suggestion
                if anpa_category := planning_data.get("anpa_category"):
                    item["anpa_category"] = deepcopy(anpa_category)
                else:
                    merge_list("anpa_category", item, planning)
```

server/planning/planning/planning.py
```diff
@@ -674,6 +674,10 @@ async def add_coverages(self, updates, original):
                 coverage.setdefault("planning", {})
                 coverage["planning"].setdefault("scheduled", planning_date)
 
+                # Inherit anpa category from planning item if not explicitly set/missing in coverage profile
+                if not coverage["planning"].get("anpa_category"):
+                    coverage["planning"]["anpa_category"] = updates.get("anpa_category", [])
+
```
eos87: This code seems duplicated. What about extracting it into a separate method? Something like this:

```python
def inherit_anpa_category(self, coverage, updates):
    if not coverage["planning"].get("anpa_category"):
        coverage["planning"]["anpa_category"] = updates.get("anpa_category", [])
```

eos87 (commented):
Left a couple of comments. Please take a look.

server/planning/planning/planning.py
petrjasek: I would actually use setdefault here:
```py
if updates.get("anpa_category"):
    coverage["planning"].setdefault("anpa_category", updates["anpa_category"])
```

petrjasek: but avoiding that duplication would be good, but it should probably handle more metadata than just `anpa_category`, so something like `inherit_planning_metadata` with the repeated parts would be good

BrianMwangi21:
@petrjasek I've picked the overlapping metadata fields to inherit. Let me know if those are ok.

eos87 (approved):
Looks good. We just need the confirmation from Petr that the inherited fields are good.

petrjasek (approved):

PR #2556, opened 2025-10-15
[SDBELGA-965] - Add contacts config support

thecalcc:
### Purpose
Add support for optional integration for editing and creating contacts from an external source. 

Resolves: #[SDBELGA-965](https://sofab.atlassian.net/browse/SDBELGA-965)

[SDBELGA-965]: https://sofab.atlassian.net/browse/SDBELGA-965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/config.ts
+++ b/client/config.ts
@@ -100,4 +100,6 @@ export function updateConfigAfterLoad() {
     }
 }
 
+export const contactsConfig: null | {editUrl: string; createUrl: string;} = appConfig.planning?.contactsProcessing;
+
 export const planningConfig = appConfig as IPlanningConfig;
--- a/client/config.ts
+++ b/client/config.ts
@@ -100,6 +100,6 @@ export function updateConfigAfterLoad() {
     }
 }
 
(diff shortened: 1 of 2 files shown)


client/config.ts
```diff
@@ -100,4 +100,6 @@ export function updateConfigAfterLoad() {
     }
 }
 
+export const contactsConfig: null | {editUrl: string; createUrl: string;} = appConfig.planning?.contactsProcessing;
```
petrjasek: just wondering about the naming `contactsProcessing`, maybe it should somehow imply it's external system, maybe like `externalContacts.editUrl` ?

thecalcc: yep makes sense

thecalcc: much better

petrjasek (approved):

PR #2557, opened 2025-10-15
[SDESK-7781] - Fix auto add to workflow

thecalcc:
### Purpose
Fix add to workflow toggle showing in the editor, when auto add to workflow is on

Resolves: #[SDESK-7781](https://sofab.atlassian.net/browse/SDESK-7781)

[SDESK-7781]: https://sofab.atlassian.net/browse/SDESK-7781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Coverages/CoverageEditor/CoverageForm.tsx
+++ b/client/components/Coverages/CoverageEditor/CoverageForm.tsx
@@ -544,29 +544,26 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
         };
 
         const isAutoAddToWorkflowOn = appConfig.planning_auto_assign_to_workflow;
+        const shouldDisableToggle = () => {
+            if (isAutoAddToWorkflowOn != true) {
+                return !(isCoverageDraft(this.props.value)
+                    && isCoverageAssigned(this.props.value)
+                    && !isItemExpired(this.props.diff));
+            } else {
+                return !planningUtils.canCancelCoverage(this.props.value, this.props.item);
+            }
+        };
 
-        if (isAutoAddToWorkflowOn === false) {
-            const shouldDisableToggle = () => {
-                if (this.props.value.add_coverage_to_workflow != true) {
-                    return !(isCoverageDraft(this.props.value)
-                        && isCoverageAssigned(this.props.value)
-                        && !isItemExpired(this.props.diff));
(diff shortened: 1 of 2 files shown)


client/components/Coverages/CoverageEditor/CoverageForm.tsx
```diff
@@ -544,29 +544,26 @@ export class CoverageFormComponent extends React.Component<IProps, IState> {
...
         const isAutoAddToWorkflowOn = appConfig.planning_auto_assign_to_workflow;
+        const shouldDisableToggle = () => {
+            if (isAutoAddToWorkflowOn != true) {
+                return !(isCoverageDraft(this.props.value)
+                    && isCoverageAssigned(this.props.value)
+                    && !isItemExpired(this.props.diff));
+            } else {
+                return !planningUtils.canCancelCoverage(this.props.value, this.props.item);
+            }
+        };
```
eos87, line 551: This piece seems hard to follow, maybe it's because of the inversion of the values not sure. I would suggest to destructure the props to avoid repetition and make it easier to read. Also, maybe invert the first `if` condition and  use an early return. Something like this below (I did not test though):

```suggestion
        const isAutoAddToWorkflowOn = appConfig.planning_auto_assign_to_workflow;
        const shouldDisableToggle = () => {
            const {value, item, diff} = this.props;
            
            if (isAutoAddToWorkflowOn === true) {
                return !planningUtils.canCancelCoverage(value, item);
            }
            
            return !(isCoverageDraft(value) && isCoverageAssigned(value) && !isItemExpired(diff));
        };
```

eos87 (commented):
looks good, I just left one comment

client/components/Coverages/CoverageEditor/CoverageForm.tsx
thecalcc: ok, and actually I reworked it because when auto add to workflow is on, we don't show the toggle anyway, so it's not relevant if it should be disabled or not in that case. 

Thus only the second execution branch is needed

eos87 (approved):

PR #2559, opened 2025-10-15
fix ingested events state after manual editing

petrjasek:
- avoid setting `draft` state when editing ingested events
- if it's already set update it on next update

CPCN-1155

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/server/planning/common.py
+++ b/server/planning/common.py
@@ -498,20 +498,6 @@ def list_uniq_with_order(list):
     return [x for x in list if not (x in seen or seen_add(x))]
 
 
-def set_ingested_event_state(updates, original):
-    """Set the ingested event state to draft"""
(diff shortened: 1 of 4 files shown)


server/planning/common.py
```diff
@@ -498,20 +498,6 @@ def list_uniq_with_order(list):
     return [x for x in list if not (x in seen or seen_add(x))]
 
 
-def set_ingested_event_state(updates, original):
```
MarkLark86, line 501: I'm confused, why are we now not wanting this?

petrjasek: that's where it was setting ingested item state to draft when those events were already published and then later it would publish them with draft state

MarkLark86 (approved):

PR #2563, opened 2025-10-16
[SDCP-969] - Add to Planning not pulling slugline automatically

eos87:
### Purpose
Resolves race condition where field values arrive after component initialization, causing "Add to Planning" to show empty slugline (and other fields) instead of auto-populating from the news item.

### What has changed
- Adds `getDerivedStateFromProps` to sync late-arriving prop values while preserving user input once they start typing.

### To reproduce the issue
- Go to SD CP
- Go to Monitoring menu 
- Click the three dot and then “Add to Planning” and then click + to add a new planning item
- Add to Planning modal dialog does not populate the `slugline`

Resolves: [SDCP-969]

[SDCP-969]: https://sofab.atlassian.net/browse/SDCP-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/fields/editor/base/text.tsx
+++ b/client/components/fields/editor/base/text.tsx
@@ -24,6 +24,7 @@ interface IState {
     key: string;
     value: string;
     suggestions: string[];
+    userHasModified: boolean;
 }
 
 export class EditorFieldText extends React.Component<IEditorFieldTextProps, IState> {
@@ -45,7 +46,29 @@ export class EditorFieldText extends React.Component<IEditorFieldTextProps, ISta
             key: uniqueId(),
             suggestions: [],
             value: get(props.item, props.field, props.defaultValue || ''),
+            userHasModified: false,
         };
+    }
+
+    /**
+     * Handles scenarios when the field has been initially rendered with empty value
+     * and then the prop value arrives later (like with 'add-to-planning').
+     * It only updates the state if user hasn't started typing to preserve user input
+     * while allowing late-loading data to populate empty fields.
+     */
+    static getDerivedStateFromProps(nextProps: IEditorFieldTextProps, prevState: IState): Partial<IState> | null {
+        const nextValue = get(nextProps.item, nextProps.field, nextProps.defaultValue || '');
+
+        // Only update if:
+        // 1. The prop value is different from current state value, AND
+        // 2. The user hasn't manually modified the field
+        if (nextValue !== prevState.value && !prevState.userHasModified) {
+            return {
+                value: nextValue
+            };
+        }
+
+        return null;
     }
 
     componentDidMount(): void {
@@ -59,7 +82,7 @@ export class EditorFieldText extends React.Component<IEditorFieldTextProps, ISta
     }
 
     onChange(value: string) {
-        this.setState({value}, () => {
+        this.setState({value: value, userHasModified: true}, () => {
             this.propsOnChange(this.state.value);
         });
     }
--- a/client/components/fields/editor/base/text.tsx
+++ b/client/components/fields/editor/base/text.tsx
@@ -84,7 +84,7 @@ export class EditorFieldText extends React.Component<IEditorFieldTextProps, ISta
     componentDidUpdate = (prevProps: Readonly<IEditorFieldTextProps>): void => {
         // Make sure to reset user modification state when item changes
         // so that late-arriving prop values can populate the field again
-        if (!isEqual(prevProps.item, this.props.item)) {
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -22,6 +22,7 @@
     "@sourcefabric/common": "0.0.65",
     "@sourcefabric/draft-js": "^0.11.5",
     "dompurify": "^1.0.11",
+    "immer": "^10.1.1",
```
thecalcc, line 25: why do we need immer here?

client/components/fields/editor/base/text.tsx
```diff
@@ -58,8 +81,16 @@ export class EditorFieldText extends React.Component<IEditorFieldTextProps, ISta
         }
     }
 
+    componentDidUpdate = (prevProps: Readonly<IEditorFieldTextProps>): void => {
+        // Make sure to reset user modification state when item changes
+        // so that late-arriving prop values can populate the field again
+        if (!isEqual(prevProps.item, this.props.item)) {
```
thecalcc: i'm just wondering isn't there a better way to see if we have changes - the `this.props.diff` if we have access to it, should hold the changes

thecalcc (commented):
small comments

package.json
eos87: it's breaking the build and I saw it was included in release/3.1 but I couldn't cherry pick it (it's used from superdesk-client-core)

client/components/fields/editor/base/text.tsx
eos87: I'm not sure but I think the `diff` would be different when you change an input so it's not what I'm looking for here. My intention is to reset the `userHasModified` when the item being edited/created is different from the previous so the data arriving late from props can update the state in `getDerivedStateFromProps`

petrjasek (commented):
I wonder if some parent component shouldn't rather wait until it has all data it needs and only render those components after

eos87:
> I wonder if some parent component shouldn't rather wait until it has all data it needs and only render those components after

Ideally yes and I considered it but, it is not simple. The problem is that one of the parent components intentionally does multiple state updates, therefore achieving that would be very tricky without refactoring many pieces. 

This is what I'm referring to:
https://github.com/superdesk/superdesk-planning/blob/154db9922837f824112435b1cea6f34d5155d5e6/client/components/Main/ItemEditor/Editor.tsx#L107-L158

eos87:
This PR depends on https://github.com/superdesk/superdesk-planning/pull/2567

client/components/fields/editor/base/text.tsx
```diff
@@ -58,8 +81,16 @@ export class EditorFieldText extends React.Component<IEditorFieldTextProps, ISta
...
+            this.setState({userHasModified: false});
```
petrjasek, line 88: I wonder now why not set it in the `getDerivedStateFromProps` ?

eos87: because `getDerivedStateFromProps` is static and do not have access to current props or state.

eos87: but thinking it twice, I think I could solve it with `getDerivedStateFromProps`. Let me give it a try.

eos87: @petrjasek nope, my idea did not work. It prevents the user from typing.

client/components/fields/editor/base/text.tsx
petrjasek: would be good not to use `isEqual` here but check using `===` or compare item id if that's when it should reset the flag

eos87: I didn't use `===` because it does not perform deep equality in dict like objects. As for the `_id`, from what I saw, the `_id` is not always provided/reliable. Although the fact that was provided or not could be sufficient in this case. Do you think that's the case?

petrjasek: it seems like overkill to use deep equality on item if it's editing single field, and as that editor could be there multiple times and item structure with associations can be quite deep I would avoid it. if it's about switching when item changes imo `===` should be enough, we probably don't have 2 different items which would pass the `===` test. using `_id` would be ok too in that case probably

eos87: makes sense. I've made the change and went with `_id`

petrjasek (approved):

PR #2565, opened 2025-10-17
[STT-1364] - Port internal note fix

thecalcc:
### Purpose
Cherry pick internal note fix

Resolves: #[STT-1364](https://sofab.atlassian.net/browse/STT-1364)

[STT-1364]: https://sofab.atlassian.net/browse/STT-1364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/InternalNoteLabel/index.tsx
+++ b/client/components/InternalNoteLabel/index.tsx
@@ -3,6 +3,7 @@ import {get} from 'lodash';
 import classNames from 'classnames';
 import {Text, Tooltip} from 'superdesk-ui-framework/react';
 import {gettext, getItemWorkflowStateLabel} from '../../utils';
+import {cleanHtmlAndReplacePlaneNewLine} from 'helpers';
 
 import './style.scss';
 
@@ -37,12 +38,7 @@ export const InternalNoteLabel: React.FC<IInternalNoteLabelProps> = ({
         return null;
     }
 
-    // Strip out any existing HTML tags and convert \n to <br> tags for HTML rendering
-    const internalNoteHtml = internalNoteRaw
-        .replace(/<[^>]*>/g, '')
-        .split('\n')
-        .map((line) => line)
-        .join('<br>');
+    const internalNoteHtml = cleanHtmlAndReplacePlaneNewLine(internalNoteRaw);
 
     const iconColor = stateField ? get(getItemWorkflowStateLabel(item, stateField), 'iconType') : 'red';
 
@@ -65,7 +61,7 @@ export const InternalNoteLabel: React.FC<IInternalNoteLabelProps> = ({
         return (
             <div className={className}>
                 {icon}
-                {showText && <span dangerouslySetInnerHTML={{__html: internalNoteHtml}} />}
+                {showText && internalNoteRaw}
             </div>
         );
     }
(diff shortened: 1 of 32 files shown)


client/components/InternalNoteLabel/index.tsx
```diff
@@ -31,12 +31,19 @@ export const InternalNoteLabel: React.FC<IInternalNoteLabelProps> = ({
...
-    const internalNote = get(item, `${prefix}${noteField}`);
+    const internalNoteRaw = get(item, `${prefix}${noteField}`);
 
-    if ((internalNote ?? '').length < 1) {
+    if ((internalNoteRaw ?? '').length < 1) {
         return null;
     }
 
+    // Strip out any existing HTML tags and convert \n to <br> tags for HTML rendering
+    const internalNoteHtml = internalNoteRaw
```
petrjasek: imo this logic should be in some helper and not implemented here

client/components/InternalNoteLabel/index.tsx
```diff
@@ -58,21 +65,31 @@ export const InternalNoteLabel: React.FC<IInternalNoteLabelProps> = ({
         return (
             <div className={className}>
                 {icon}
-                {showText && internalNote}
+                {showText && <span dangerouslySetInnerHTML={{__html: internalNoteHtml}} />}
```
petrjasek: also wondering if the markup is converted to plain text list why using join there and not map here?

thecalcc: i thought for one source of truth but yes it can just be internalNote

client/helpers.ts
```diff
@@ -35,3 +35,11 @@ export const partitionLineItems = (items: Array<ILineConfig>) => partition(items
         return superdeskApi.helpers.assertNever(position);
     }
 });
+
+// Strip out any existing HTML tags and convert \n to <br> tags for HTML rendering
+export const cleanHtmlAndReplacePlaneNewLine = (text: string): string => {
```
petrjasek: we have already stripHtmlTags in `core/utils`, and that split and joining wouldn't make reusable, that's specific to the usage, and imo can be avoided

client/components/InternalNoteLabel/index.tsx
```diff
@@ -31,12 +32,14 @@ export const InternalNoteLabel: React.FC<IInternalNoteLabelProps> = ({
...
 }) => {
-    const internalNote = get(item, `${prefix}${noteField}`);
+    const internalNoteRaw = get(item, `${prefix}${noteField}`);
 
-    if ((internalNote ?? '').length < 1) {
+    if ((internalNoteRaw ?? '').length < 1) {
         return null;
     }
 
+    const internalNoteHtml = cleanHtmlAndReplacePlaneNewLine(internalNoteRaw);
```
petrjasek: would probably do rather:
```
internalNoteText = stripHtml(internalNote);

const content = (
    <div>
        {internalNoteText.split("\n").map((line) => <p>line</p>)}
    </div>
);
```
so there is no need to use dangerouslySetInnerHTML

petrjasek (approved):

PR #2570, opened 2025-10-20
increase the limit when fetching linked items for assignments

petrjasek:
atm it's using elastic default 10

STT-1402

### Front-end checklist

Resolves: #[issue-number]


MarkLark86 (approved):

server/planning/assignments/assignments.py
```diff
@@ -153,14 +156,11 @@ async def get_archive_items_for_assignment(self, assignment):
...
-                            "must": {"term": {"assignment_id": str(assignment[ID_FIELD])}},
-                        }
-                    }
-                }
-            }
+                "bool": {
+                    "must": {"term": {"assignment_id": str(assignment[ID_FIELD])}},
+                },
+            },
+            "size": get_app_config("ASSIGNMENTS_LINKED_ITEMS_SIZE", 500),
```
MarkLark86: For future reference, we also have the newer [get_config](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/app.py#L53) which is more type safe.

PR #2571, opened 2025-10-21
[STT-1384] fix: Re-write AssignmentEditor

MarkLark86:
### Purpose
Removing Coverage provider from the `AssignmentEditor` raised an exception. After investigating the component, it was unnecessarily complex than it needed to be.

### What has changed
* Add ts interfaces to `AssignmentEditor`, `EditCoverageAssignmentModal` and `UpdateAssignmentComponent` components
* Re-write `AssignmentEditor`, so it doesn't need local state, reducing react lifecycle updates
* Set Assignment priority when creating the Coverage (before the components are even rendered - it was being set on component mount, which caused a re-render)
* Update `onChange` to support changing multiple fields in the one step
* Move more redux state connected props to the `AssignmentEditor` (previously props were not used by parent components, just passed through)
* Remove old Assignment modal component that is no longer used

### Front-end checklist

Resolves: [STT-1384](https://sofab.atlassian.net/browse/STT-1384)

[STT-1384]: https://sofab.atlassian.net/browse/STT-1384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/components/ItemActionConfirmation/forms/updateAssignmentForm.tsx
```diff
@@ -1,17 +1,19 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import {connect} from 'react-redux';
 import {get, set, isEqual, cloneDeep} from 'lodash';
 
-import {IAssignmentItem} from '../../../interfaces';
+import {IDesk} from 'superdesk-api';
+import {IAssignmentItem, IAssignmentPriority, ICoverageProvider} from '../../../interfaces';
```
eos87: nit:
```suggestion
import type {IDesk} from 'superdesk-api';
import type {IAssignmentItem, IAssignmentPriority, ICoverageProvider} from '../../../interfaces';
```

client/components/ItemActionConfirmation/forms/updateAssignmentForm.tsx
```diff
@@ -20,37 +22,64 @@ import {AbsoluteDate} from '../..';
...
-            valid: true,
+            diff: cloneDeep(this.props.original),
+            valid: isAssignmentDeskValid(this.props.original.assigned_to?.desk),
         };
 
         this.dom = {popupContainer: null};
         this.onChange = this.onChange.bind(this);
-        this.setValid = this.setValid.bind(this);
+        this.submit = this.submit.bind(this);
+        this.getPopupContainer = this.getPopupContainer.bind(this);
```
eos87: this binds are not needed if you declare your functions like this

```typescript
getPopupContainer = () => {
   ...
}
```

eos87 (approved):

client/components/ItemActionConfirmation/forms/updateAssignmentForm.tsx
MarkLark86: I did not know about this one, thanks :+1:

PR #2573, opened 2025-10-21
[STT-1405] fix: EventRelatedPlanning crashes on update when disabled

MarkLark86:
### Purpose
The `EventRelatedPlanning` component crashes on update or unmount if the component is in a read-only state.

### What has changed
* Only process associated items if the component is not disabled

Resolves: [STT-1405](https://sofab.atlassian.net/browse/STT-1405)

[STT-1405]: https://sofab.atlassian.net/browse/STT-1405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
+++ b/client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
@@ -39,18 +39,22 @@ export class EditorFieldEventRelatedPlanningsComponent extends React.PureCompone
     }
 
     componentDidUpdate(prevProps: Readonly<IRelatedPlanningProps>): void {
-        const prevPlanIds = prevProps.item.associated_plannings.map((x) => x._id);
-        const currentPlanIds = this.props.item.associated_plannings.map((x) => x._id);
+        if (this.props.disabled !== true) {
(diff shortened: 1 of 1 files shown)


thecalcc (approved):

client/components/fields/editor/EventRelatedPlannings/EventRelatedPlannings.tsx
```diff
@@ -39,18 +39,22 @@ export class EditorFieldEventRelatedPlanningsComponent extends React.PureCompone
...
+            }
         }
     }
 
     componentWillUnmount(): void {
-        this.props.item.associated_plannings
-            .filter((x) => x._temporary != null)
-            .forEach((x) => planningApi.locks.unlockEmbeddedItem(x as IPlanningItem));
+        if (this.props.disabled !== true) {
+            this.props.item.associated_plannings
```
thecalcc, line 54: is it crashing because `item.associated_plannings.map` throws an error?

MarkLark86: Yeah, because the associated items is not defined.
But also, if the component is disabled, no locks should be involved anyway

PR #2574, opened 2025-10-21
[STT-1380] - Use multiple content default value in add to planning

thecalcc:
### Purpose
Use the default value coming from coverage profile for `multiple_content` coverage field when creating a planning with coverage item from "Add to planning" modal

### What has changed
Now the function responsible for creating a planning item that gets triggered from the add to planning modal also reads the default value for `multiple_content` field, and adds it to the newly created coverage item.

### Steps to test
1. Configure `multiple_content` default value to be `true` for text coverage profile. 
2. Create an article
3. Click add to planning in 3 dots menu of the article
4. Create a new planning item, instead of using an existing one
5. The `multiple_content` field of the coverage of the newly created planning item should be set to `true` (toggled on)

Resolves: #[STT-1380](https://sofab.atlassian.net/browse/STT-1380)

[STT-1380]: https://sofab.atlassian.net/browse/STT-1380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Coverages/CoverageArrayInput.tsx
+++ b/client/components/Coverages/CoverageArrayInput.tsx
@@ -5,7 +5,10 @@ import {isEmpty} from 'lodash';
 import {
     EDITOR_TYPE,
     IAssignmentPriority,
-    ICoverageProvider, IEventItem, IFile,
+    ICoverageProvider,
+    ICoverageType,
+    IEventItem,
+    IFile,
     IG2ContentType,
     IGenre,
     IInputArrayHocModeOptions,
@@ -63,6 +66,8 @@ interface IReduxStateProps {
     newsCoverageStatus: Array<IPlanningNewsCoverageStatus>;
     coverageAddAdvancedMode: boolean;
     defaultDesk: IDesk;
+    getCoverageProfile: (coverageType?: ICoverageType) =>
+        ReturnType<typeof selectors.coverageProfiles.getCoverageProfileByContentType>;
 }
 
 interface IReduxDispatchProps {
@@ -76,6 +81,8 @@ interface IState {
 }
 
 const mapStateToProps = (state): IReduxStateProps => ({
+    getCoverageProfile: (contentType: ICoverageType) =>
+        selectors.coverageProfiles.getCoverageProfileByContentType(state, contentType),
     users: selectors.general.users(state),
     desks: selectors.general.desks(state),
     genres: state.genres,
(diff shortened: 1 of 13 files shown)


client/utils/planning.tsx
```diff
@@ -1057,15 +1058,26 @@ function createCoverageFromNewsItem(
...
         ...newCoverage.planning,
-        g2_content_type: get(contentType, 'qcode', PLANNING.G2_CONTENT_TYPE.TEXT),
+        g2_content_type: contentTypeQcode,
         slugline: get(addNewsItemToPlanning, 'slugline', ''),
         ednote: get(addNewsItemToPlanning, 'ednote', ''),
         scheduled: moment().add(1, 'hour')
             .startOf('hour'),
     };
 
+    const allCoverageProfiles = coverageProfiles(planningApi.redux.store.getState());
```
eos87: I think we need to stop doing this. It tighly couples the function to the external redux store making it much harder to write any tests. I would suggest to pass the computed values (`allCoverageProfiles`) as another parameter to the function. Ideally, you should use the `reselect` selectors we already have to grab this values and then pass them down. With that, it would be fairly simple to write a test and cover this change :)

eos87 (changes requested):
I left a comment, please take a look.

client/components/Coverages/CoverageArrayInput.tsx
```diff
@@ -76,6 +81,8 @@ interface IState {
 }
 
 const mapStateToProps = (state): IReduxStateProps => ({
+    getCoverageProfile: (contentType: ICoverageType) =>
+        selectors.coverageProfiles.getCoverageProfileByContentType(state, contentType),
     users: selectors.general.users(state),
     desks: selectors.general.desks(state),
     genres: state.genres,
```
eos87, line 87: Returning an anonymous function here will create a new prop function every time state changes causing unnecessary re-renders (I think even when nothing has changed). 

I would suggest this: 
1. build a selector that returns a map of the profiles (maybe we already have one, not sure) so it's a stable object that won't cause re-renders
2. Look up the specific profile by key instead of calling a function. 

Something like this below 

1. 
```typescript
export const getCoverageProfilesMap = createSelector(
    coverageProfiles,
    (profiles) => profiles.reduce<Record<string, ICoverageContentProfile>>((acc, p) => {
        if (p?.content_type) acc[p.content_type] = p;
        return acc;
    }, {})
);
```

2.a - in your mapStateToProps replace the anonymous function for something like 
```typescript
    coverageProfilesMap: selectors.coverageProfiles.getCoverageProfilesMap(state),
    // ...
});
```

2.b - in the component you would get the profile like this
```typescript
const profile = this.props.coverageProfilesMap[coverageType];
```

eos87 (commented):
In general we need to avoid returning anonymous functions in `mapStateToProps` to avoid unnecessary re-renders. Please take a look at my comment, maybe it would do the trick, it probably needs some tweaking.

eos87 (approved):
Looks good!

PR #2578, opened 2025-10-22
[STT-1399] - Bugfix: embedded planning coverages saving

thecalcc:
### Purpose
Fix a saving bug where when you add a coverage to an embedded planning, then also make changes to the event and then save all changes from the event header, the embedded planning item wouldn't have. 

### What has changed
Now after changes occur in the embedded planning items after they're saved via `handleEmbeddedItems`, they are synced to the event's `associated_plannings` so that the endpoint that saves all the data for the event, doesn't override the newly saved changes, and also cause `_etag`s of the embedded planning items to be out of sync.

### Steps to test
Add an event and save it.
Add an embedded planning item to it and save it.
Add a coverage inside the embedded planning item and make a change in the event.
Click the save from the event header. All changes should save now.

Resolves: #[STT-1399](https://sofab.atlassian.net/browse/STT-1399)

[STT-1399]: https://sofab.atlassian.net/browse/STT-1399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/editor-standalone/authoring-autosave.ts
+++ b/client/components/editor-standalone/authoring-autosave.ts
@@ -1,10 +1,10 @@
+import type {IAuthoringAutoSave, IBaseRestApiResponse} from 'superdesk-api';
+import type {IEventOrPlanningItem} from '../../interfaces';
 import moment from 'moment';
-import {IAuthoringAutoSave, IBaseRestApiResponse} from 'superdesk-api';
 import * as selectors from '../../selectors';
 import {throttle, DebouncedFunc} from 'lodash';
 import {planningApi, superdeskApi} from '../../superdeskApi';
 import {omitFields} from './utils';
-import {IEventOrPlanningItem} from '../../interfaces';
 
 export class AutoSaveHttp<T extends IBaseRestApiResponse & IEventOrPlanningItem> implements IAuthoringAutoSave<T> {
     private autoSaveThrottled: DebouncedFunc<typeof this.autosave>;
--- a/client/components/editor-standalone/authoring-autosave.ts
+++ b/client/components/editor-standalone/authoring-autosave.ts
@@ -4,8 +4,9 @@ import * as selectors from '../../selectors';
 import {throttle, DebouncedFunc} from 'lodash';
 import {planningApi, superdeskApi} from '../../superdeskApi';
 import {omitFields} from './utils';
+import {IEventOrPlanningItem} from '../../interfaces';
 
-export class AutoSaveHttp<T extends IBaseRestApiResponse> implements IAuthoringAutoSave<T> {
+export class AutoSaveHttp<T extends IBaseRestApiResponse & IEventOrPlanningItem> implements IAuthoringAutoSave<T> {
     private autoSaveThrottled: DebouncedFunc<typeof this.autosave>;
 
     private autosavePromise: Promise<void> | null;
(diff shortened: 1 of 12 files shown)


thecalcc:
I've also done some type improvements and cleanup to the save handling of embedded items

client/components/editor-standalone/authoring-autosave.ts
```diff
@@ -4,8 +4,9 @@ import * as selectors from '../../selectors';
 import {throttle, DebouncedFunc} from 'lodash';
 import {planningApi, superdeskApi} from '../../superdeskApi';
 import {omitFields} from './utils';
+import {IEventOrPlanningItem} from '../../interfaces';
```
eos87: nit:
```suggestion
import type {IEventOrPlanningItem} from '../../interfaces';
```

client/components/editor-standalone/utils.ts
```diff
@@ -1,10 +1,9 @@
-import {IEventOrPlanningItem, IProfileSchemaType, IProfileSchemaTypeString} from 'interfaces';
+import {IEventOrPlanningItem, IPlanningItem, IProfileSchemaType, IProfileSchemaTypeString} from 'interfaces';
```
eos87:
```suggestion
import type {IEventOrPlanningItem, IPlanningItem, IProfileSchemaType, IProfileSchemaTypeString} from 'interfaces';
```

client/components/editor-standalone/authoring-storage-planning-http.ts
```diff
@@ -32,26 +33,30 @@ export const authoringStoragePlanningItemHttp: IAuthoringStorage<IPlanningItem>
     saveEntity: (current, original) => {
         const {httpRequestJsonLocal} = superdeskApi;
         const {generatePatch} = superdeskApi.utilities;
+        const patchData = omitFields(
+            generatePatch(
+                planningUtils.modifyForServer(original),
+                planningUtils.modifyForServer(current, original),
```
eos87: could you please explain about this part? just curious what's going on.

client/components/editor-standalone/save-handling.ts
```diff
@@ -46,29 +46,29 @@ export const handleEmbeddedItems = async<T extends IEventItem | IPlanningItem>(
...
         if (action === 'SAVE') {
-            updatedItems.push(exposed.save());
+            updatedItems.push(await exposed.save());
         } else if (action === 'DISCARD') {
             await exposed.discardUnsavedChanges();
-            updatedItems.push(Promise.resolve(latestItem));
+            updatedItems.push(latestItem);
         } else {
-            updatedItems.push(exposed.handleUnsavedChanges());
+            updatedItems.push(await exposed.handleUnsavedChanges());
```
eos87: Wondering if we would loose parallelism here by switching to `await` instead of keeping `Promise.all`. I would suggest to revert this back to promises so the items are saved concurrently. Or was it on purpose?

client/components/editor-standalone/authoring-autosave.ts
```diff
@@ -91,11 +91,7 @@ export class AutoSaveHttp<T extends IBaseRestApiResponse> implements IAuthoringA
         return httpRequestJsonLocal<T>({
             method: 'GET',
             path: `/${this.resource}/${id}`,
-        }).then((res) => {
-            const result = this.modifyForClient(res);
-
-            return result;
-        });
+        }).then(this.modifyForClient);
```
eos87: I'm just a little worried that `this.modifyForClient` inside `.then(` might lose the instance context (`this`). I'd suggest this below to be on the safe side:

```suggestion
        }).then((res) => this.modifyForClient(res));
```

client/components/editor-standalone/authoring-storage-in-memory.ts
```diff
@@ -3,7 +3,7 @@ import {IAuthoringAutoSave, IAuthoringStorage} from 'superdesk-api';
 import {getProfile} from './profile';
 import {autosave} from '../../api/autosave';
 import {superdeskApi} from '../../superdeskApi';
-import {IEventOrPlanningItem} from 'interfaces';
+import {IEventOrPlanningItem} from '../../interfaces';
```
eos87, line 6: what happened here, was it not finding it?

client/components/editor-standalone/authoring-storage-event-http.ts
```diff
@@ -52,11 +53,14 @@ export const authoringStorageEventItemHttp: IAuthoringStorage<IEventItem> = {
...
     },
+
     getContentProfile: (item) => {
         return Promise.resolve(getProfile('event', item.language));
     },
+
     closeAuthoring: (_current, original, hasUnsavedChanges, _cancelAutosave, doClose) => {
-        return Promise.resolve();
+        return Promise.resolve({cancelled: false});
     },
```
eos87, line 63: it would be nice to type the return types of these functions if it's not too much work.

client/components/editor-standalone/authoring-storage-planning-http.ts
```diff
@@ -32,26 +33,30 @@ export const authoringStoragePlanningItemHttp: IAuthoringStorage<IPlanningItem>
...
     },
+
     getContentProfile: (item) => {
         return Promise.resolve(getProfile('planning', item.language));
     },
+
     closeAuthoring: (_current, _original, _hasUnsavedChanges, _cancelAutosave, _doClose) => {
-        return Promise.resolve();
+        return Promise.resolve({cancelled: false});
     },
```
eos87, line 59: same about the return type here

eos87 (changes requested):

client/components/editor-standalone/authoring-autosave.ts
thecalcc: let me double check if that's any different than my version, from what I understood my version is just syntactic sugar for yours

client/components/editor-standalone/authoring-storage-event-http.ts
thecalcc: the whole object is fully typed here because it's defined like this:

```js
export const authoringStorageEventItemHttp: IAuthoringStorage<IEventItem> 
```

client/components/editor-standalone/authoring-storage-in-memory.ts
thecalcc: yeah, not sure why it was written with 'interfaces'. we use relative imports in planning always

client/components/editor-standalone/authoring-storage-planning-http.ts
thecalcc: essentially `modifyForServer` in this case is taking the planning item and adjusting genre to the format the server expects. there's also a bit of code taking care of the coverage scheduled updates. 

Then we generate an update/patch object that contains item updates. And omitFields just strips some fields like `_created`, `_links`, etc. 

One thing that's wrong is this line here 

```js
planningUtils.modifyForServer(current, original)
```

We don't need to pass the `original` there, so I'll refactor that

The bigger thing - in the first place ideally we should get rid of the client/server transformations, but that's deeply tied into planning code.

client/components/editor-standalone/authoring-storage-planning-http.ts
thecalcc: everything in the object is typed because of `IAuthoringStorage`

client/components/editor-standalone/save-handling.ts
thecalcc: great point, we would loose all parallelism, I did it like this now because historically I could only get the saving to work in parallel and not sequentially like it is now. I see the performance downside though of having it sequential, let me know if you prefer it in parallel. I personally think it's not a big deal because even if we're saving 10 items that wouldn't take too much time, and we have proper loaders rendering in the UI

client/components/editor-standalone/authoring-autosave.ts
eos87: it actually makes a difference, but yeah double check it.

client/components/editor-standalone/authoring-storage-event-http.ts
eos87: aha, perfect!

client/components/editor-standalone/save-handling.ts
eos87: let's keep it concurrently/parallel so we also take advantage of our async backend now :)

eos87 (approved):

PR #2579, opened 2025-10-22
[STT-1403] Remove required field validation from `Add to Planning`

eos87:
### Purpose
"Add to planning" action should be executed without the topic/slugline field being present or filled in. So in this PR we're removing any validation when the user "clicks" the add to planning button as the required field will still be there in the editor form.

### What has changed
- Eliminated logic enforcing required fields when adding to planning
- Remove related test cases

Resolves: [STT-1403]

[STT-1403]: https://sofab.atlassian.net/browse/STT-1403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/controllers/AddToPlanningController_test.ts
+++ b/client/controllers/AddToPlanningController_test.ts
@@ -71,46 +71,21 @@ describe('AddToPlanningController', () => {
         api,
         lock,
         session,
-        userList
+        userList,
+        $timeout
     ) => {
         api.find = sinon.stub().returns($q.reject({}));
-        return new AddToPlanningController(null,
+        return (new AddToPlanningController(null,
             scope, sdPlanningStore, notify,
-            gettext, api, lock, session, userList
-        )
+            gettext, api, lock, session, userList,
+            $timeout, {}
+        ) as any)
             .then(() => { /* no-op */ }, () => {
                 expect(api.find.callCount).toBe(1);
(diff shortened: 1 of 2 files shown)


eos87:
Some context here https://sourcefabric.slack.com/archives/C1XBKE70C/p1761049520444699

client/controllers/AddToPlanningController_test.ts
```diff
@@ -87,47 +89,22 @@ describe('AddToPlanningController', () => {
...
     it('notifies the user if the item is already linked to an assignment', inject((
         sdPlanningStore,
         notify,
         gettext,
         api,
         lock,
         session,
-        userList
+        userList,
+        $timeout
```
thecalcc, line 100: how did this work before without those two injected?

thecalcc (approved):

client/controllers/AddToPlanningController_test.ts
eos87: no idea but the test had a bunch of warnings because of the missing parameters

thecalcc: i see, just checked it - one usage of `$timeout` in the controller and one of `userList`, so should be all good

PR #2582, opened 2025-10-23
avoid deep copying editor state in reducers

petrjasek:
to avoid editor rerendering on each update

SDCP-1001

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/reducers/forms.ts
+++ b/client/reducers/forms.ts
@@ -5,8 +5,11 @@ import {IPlanningContentProfile} from '../interfaces';
 import {AUTOSAVE, ITEM_TYPE, MAIN, FORMS} from '../constants';
 import {createReducer} from './createReducer';
 import {eventUtils, getItemId, getItemType, planningUtils} from '../utils';
-import {cloneDeep, get, set} from 'lodash';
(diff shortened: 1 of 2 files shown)


client/reducers/forms.ts
```diff
@@ -5,8 +5,11 @@ import {IPlanningContentProfile} from '../interfaces';
 import {AUTOSAVE, ITEM_TYPE, MAIN, FORMS} from '../constants';
 import {createReducer} from './createReducer';
 import {eventUtils, getItemId, getItemType, planningUtils} from '../utils';
-import {cloneDeep, get, set} from 'lodash';
+import {get, set} from 'lodash';
 import {EDITOR_TYPE, IEditorFormState, IFormState} from '../interfaces';
+import {produce, setAutoFreeze} from 'immer';
+
+setAutoFreeze(false);
```
thecalcc, line 12: we can't avoid state mutation in other places?

petrjasek: we can, but since this is hotfix for an older release I want to reduce the changes

thecalcc (approved):

PR #2584, opened 2025-10-27
[SDESK-7785] fix: Copy Coverage assignee changes to Assignment in workflow

MarkLark86:
### Purpose
When changing the Coverage assignee details (desk, user, priority, coverage_provider), if the Assignment is already in workflow these changes aren't copied to the Assignment as well.

Additionally there were business logic differences between the `Reassign` button on the Coverage form and the same action in the Assignments page.

### What has changed
* Copy Coverage assignee changes to Assignments already in workflow
* Add tooltips in Coverage form header, as to why the `Reassign` or `Remove` action is not available
* Don't allow re-assign or remove if Assignment is completed or cancelled
* Don't allow re-assigning the desk if the Assignment is in progress

Resolves:
* [SDESK-7785](https://sofab.atlassian.net/browse/SDESK-7785)
* https://sofab.atlassian.net/browse/STT-1384

[SDESK-7785]: https://sofab.atlassian.net/browse/SDESK-7785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
+++ b/client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
@@ -117,15 +117,78 @@ class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
         }
     }
 
+    renderAssignmentFunctionButtons() {
+        const {
+            value,
+            addNewsItemToPlanning,
+            readOnly,
+            lockedItems,
+        } = this.props;
+
+        if (addNewsItemToPlanning != null
+            || (value as ICoverageScheduledUpdate).scheduled_update_id != null
+            || readOnly === true
+        ) {
+            return null;
+        }
+
+        const assignmentState = value.assigned_to?.state;
+        const isAssignmentLocked = lockedItems?.assignment
+            && value.assigned_to?.assignment_id in lockedItems.assignment;
+        const buttonsDisabled = [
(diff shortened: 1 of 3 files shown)


client/components/Coverages/CoverageEditor/CoverageFormHeader.tsx
```diff
@@ -222,28 +223,62 @@ class CoverageFormHeaderComponent extends React.PureComponent<IProps> {
...
+                                    style="hollow"
+                                    size="small"
+                                    expand
+                                    disabled={buttonsDisabled}
+                                    tooltip={removeTooltip}
+                                />
+                            </ListRow>
+                        </Column>
+                    );
+                })()}
```
eos87: I would recommend extracting this into a separate render function or simple component. 
The benefits: 
- more of a standard approach when it comes to writing JSX ("convention")
- the IIFE creates a new function on every render therefore it has a slight performance cost almost negligible though I know)
- avoid the mixing concerns 
- better debugging info (react dev tools) if separate simple component is chosen

eos87 (commented):
Looks good, just have one comment. Please take a look. Thanks!

eos87 (approved):

PR #2585, opened 2025-10-28
[SDESK-7790] - Fix event unlock notification payload

thecalcc:
### Purpose
Unlock the original event that was converted to a recurring one, via the "Convert to recurring event" action modal. 

### What has changed
Add `item_type` to the payload of `event:unlock` notification, so that the notification handler on the frontend can remove the item lock

Resolves: #[SDESK-7790](https://sofab.atlassian.net/browse/SDESK-7790)

[SDESK-7790]: https://sofab.atlassian.net/browse/SDESK-7790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/events/events.py
+++ b/server/planning/events/events.py
@@ -627,6 +627,7 @@ class EventsService(AsyncBaseService):
                 etag=updates["_etag"],
                 recurrence_id=original.get("recurrence_id") or None,
                 from_ingest=from_ingest,
+                type=original.get("type"),
             )
 
         await self.delete_event_files(updates, original)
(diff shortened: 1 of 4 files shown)


thecalcc:
After the initial fix, I adjusted also the planning unlocking. Also the assignments seem to be using the unlock from [lock service](https://github.com/superdesk/superdesk-planning/blob/50f7b3d3eae347d870a872e93bc7c56b42625616/server/planning/item_lock.py#L159) which already passes type param.

MarkLark86 (approved):

PR #2588, opened 2025-10-30
[STT-1427] - Fix list reloading on assignments notification

thecalcc:
### Purpose
Improve the assignments list UI so groups of lists don't rerender completely when there's an assignment change - update an existing one , or on creation of a new one.

### What has changed

**Feature related:**

Logic was introduced to control the triggering of groups UI loaders. 
Reworked the reducer handling assignments actions to use `produce` so we can stop deep copying the entire state in each action, causing unnecessary component re-rendering. 
Trigger the list UI loaders when a search is executed from the advanced search sidebar. Previously that was getting triggered only when user made a query from the top search bar.

**Small improvements:**

I also used the opportunity to just very barely make the code more readable, making conditional logic easier to read in places.

### Steps to test
1. Open two browser windows side by side. 
2. On the first one, go to the assignments page, on the second one open planning.

Case 1:
1. Create a few assignments to the user you're logged in with on the screen with assignments view.
2. Notice the loading mechanism doesn't get triggered, but items count and items in the TODO list view change, with the newly added assignment. 

Case 2:
1. Mark an existing assignment as completed 
2. Notice that the group loaders aren't rendering, rather:
 - marked item should be removed from the "In Progress" group, and get added to the "Completed" one
 - groups' scroll position doesn't change, if you've got enough items so that the 2 groups are overflowing of items and are scrollable. 

### Screenshots
https://github.com/user-attachments/assets/88a12e6f-627e-499c-ab0f-8b69552dd545

Resolves: #[STT-1427](https://sofab.atlassian.net/browse/STT-1427)

[STT-1427]: https://sofab.atlassian.net/browse/STT-1427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/actions/assignments/notifications.ts
+++ b/client/actions/assignments/notifications.ts
@@ -10,7 +10,6 @@ import * as selectors from '../../selectors';
 import assignments from './index';
 import main from '../main';
 import {hideModal, showModal} from '../index';
-import * as actions from '../../actions';
 import planningApis from '../planning/api';
 
 const _notifyAssignmentEdited = (assignmentId) => (
@@ -59,7 +58,7 @@ const onAssignmentCreated = (_e, data) => (
         if (querySearchSettings.deskIds == null || querySearchSettings.deskIds.length > 0 &&
             (currentDesk === data.assigned_desk || currentDesk === data.original_assigned_desk)
         ) {
-            dispatch(assignments.ui.reloadAssignments([data.assignment_state]));
+            dispatch(assignments.ui.reloadAssignments([data.assignment_state], false));
         }
 
         return Promise.resolve();
@@ -103,7 +102,7 @@ const onAssignmentUpdated = (_e, data) => (
             currentDesk === data.original_assigned_desk
         ) {
             dispatch(assignments.api.fetchAssignmentHistory({_id: data.item}));
-            dispatch(assignments.ui.reloadAssignments([data.assignment_state]));
+            dispatch(assignments.ui.reloadAssignments([data.assignment_state], false));
 
             dispatch(assignments.api.fetchAssignmentById(data.item))
                 .then((assignmentInStore) => {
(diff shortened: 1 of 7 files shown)


thecalcc:
Most changes are with the reducer and how we handle state updates. The other part is not triggering the UI loaders for the cases where we agreed, that is - updating an assignment and a new assignment arriving via web socket. 

The other changes are nearly unrelated, just took the opportunity to kind of clean up the code just a bit.

On the changes in the assignments reducer - ideally we'd remove this pattern we use currently to create a reducer and use switch statements within all reducers. That way we can easily migrate all reducers for all entities by wrapping them into a single `produce` function instead of doing so at action level.

client/actions/assignments/notifications.ts
```diff
@@ -122,18 +121,20 @@ const onAssignmentUpdated = (_e, data) => (
...
 
             if (data.assignment_state === ASSIGNMENTS.WORKFLOW_STATE.CANCELLED ||
-                 data.assignment_state === ASSIGNMENTS.WORKFLOW_STATE.IN_PROGRESS) {
+                data.assignment_state === ASSIGNMENTS.WORKFLOW_STATE.IN_PROGRESS) {
                 // If we are in authoring workspace (fulfilment) and assignment is previewed,
                 // close it
-                if (selectors.general.currentWorkspace(getState()) === WORKSPACE.AUTHORING &&
-                        selectors.getCurrentAssignmentId(getState()) === data.item) {
+                if (selectors.general.currentWorkspace(getState()) === WORKSPACE.AUTHORING
+                    && selectors.getCurrentAssignmentId(getState()) === data.item
```
eos87: nit:
```suggestion
                const state = getState();
                if (selectors.general.currentWorkspace(state) === WORKSPACE.AUTHORING
                    && selectors.getCurrentAssignmentId(state) === data.item
```

client/actions/assignments/ui.ts
```diff
@@ -161,42 +166,47 @@ const updatePreviewItemOnRouteUpdate = () => (
...
-                    },
-                    () => {
+
+                        // Preview only if user is a member of that assignment's desk
+                        const currentUserId = selectors.general.currentUserId(getState());
+                        const user = desks.deskMembers[item.assigned_to.desk].find(
+                            (u) => u._id === currentUserId);
+
+                        if (!user) {
+                            notify.error('Insufficient privileges to view the assignment');
```
eos87: we should probably use gettext here to allow translations?

client/actions/assignments/ui.ts
```diff
@@ -161,42 +166,47 @@ const updatePreviewItemOnRouteUpdate = () => (
...
+                            return dispatch(self.closePreview());
+                        }
+
+                        // For previewing, add it to the store even though it might be from another
+                        dispatch(assignments.api.receivedAssignments([item]));
+
+                        return dispatch(self.preview(item));
+                    })
+                    .catch(() => {
                         notify.error('Assignment does not exist');
```
eos87: same here

client/reducers/assignment.ts
eos87, line 1: This looks great! My only suggestion would be that for `produce` we use `draftState` instead of `state` and `action` instead `payload`. I think that would bring better meaning. What do you think?

eos87 (approved):
Great job 👏 I left a couple of small comments.

client/reducers/assignment.ts
thecalcc: yep i agree `draftState` is better, but not sure about `action` since it's really data that we're trying to extract from that prop. Action is used somewhere up the chain of reducers here probably, and inside this reducer action should be the key of the object e.g.
image

But if you want maybe we do `actionPayload`?

eos87: yep, `actionPayload` sounds good!

PR #2595, opened 2025-11-11
[SDBELGA-1005] fix: SelectUser not updating with Desk selection

MarkLark86:
### Purpose
The SelectUser input was not updating the list of available users when the DeskId changes.

### What has changed
Force a re-mount of the SelectUser component.

Resolves: [SDBELGA-1005](https://sofab.atlassian.net/browse/SDBELGA-1005)

[SDBELGA-1005]: https://sofab.atlassian.net/browse/SDBELGA-1005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Assignments/AssignmentEditor.tsx
+++ b/client/components/Assignments/AssignmentEditor.tsx
@@ -327,6 +327,7 @@ export class AssignmentEditorComponent extends React.Component<IProps, IState> {
                     <Row>
                         <div data-test-id={this.FIELDS.USER}>
                             <SelectUser
+                                key={this.props.value.assigned_to?.desk ?? null}
                                 disabled={disableUserSelection}
                                 deskId={this.props.value.assigned_to?.desk ?? null}
                                 selectedUserId={this.state.userId}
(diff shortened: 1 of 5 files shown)


MarkLark86:
@thecalcc Although this solution works, I don't like the solution. After digging deeper it looks like an issue with the TreeSelect component in the UI-Framework only using the `getOptions` prop in the [constructor](https://github.com/superdesk/superdesk-ui-framework/blob/v4/app-typescript/components/TreeSelect/TreeSelect.tsx#L114) and using internal state thereafter. 

Looks like we'll need a fix on the UI-Framework side for the proper fix

thecalcc (approved):
i agree with you comment, you bring up a great point. this is a known issue with the TreeSelect we never got around to fixing

thecalcc:
CI would need a fix

MarkLark86:
@thecalcc The Ci-Client 14 issue is not related to this PR, I will look into it afterwards (it's passing on NodeJS v22)

PR #2605, opened 2025-11-14
[SDESK-6761] improve: Loading of publish queue preview

MarkLark86:
### Purpose
When opening the preview of a `Planning` or `Event` item in the Publish Queue page, the preview area would open but show up blank until all data it loaded.

Loading of data includes [many things](https://github.com/superdesk/superdesk-planning/blob/develop/client/services/PlanningStoreService.ts#L178), to ensure all required data is in the redux store.

### What has changed
Improve the user experience by providing a loading animation while the required data is fetched, showing the user that their interaction has started.

There was also a gap at the top of the preview that was caused by css classes in the SidePanel from core. Added a style prop to fix this (should ideally be looked at in the UI-Framework/Superdesk-Client-Core).

### Screenshots
**Previous Experience**
[publish_queue_loading_planning_previous_experience.webm](https://github.com/user-attachments/assets/bb8d09ec-c211-42dc-a4e9-aa8135d75bba)

**New Experience**
[publish_queue_loading_planning_new_experience.webm](https://github.com/user-attachments/assets/b0426c76-9141-400e-aade-58d725adfe08)

### Front-end checklist

Resolves: [SDESK-6761](https://sofab.atlassian.net/browse/SDESK-6761)

[SDESK-6761]: https://sofab.atlassian.net/browse/SDESK-6761?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/index.ts
+++ b/client/index.ts
@@ -113,7 +113,9 @@ export default angular.module('superdesk-planning', [])
                         scope.$watch('selected.preview', (newValue) => {
                             store.dispatch(actions.main.onQueueItemChange(newValue));
                         });
-                        setTimeout(() => resolve(store), 500);
+
+                        // Use a timeout as resolving the promise straight away causes the app to crash
+                        setTimeout(() => resolve(store));
                     });
                 });
 
--- a/client/index.ts
+++ b/client/index.ts
@@ -105,15 +105,20 @@ export default angular.module('superdesk-planning', [])
 
             ng.register($injector);
 
-            const callback = (extension, scope) => (
-                sdPlanningStore.initWorkspace(WORKSPACE.AUTHORING, (store) => {
-                    store.dispatch(actions.fetchAgendas());
-                    extension.props.store = store;
-                    scope.$watch('selected.preview', (newValue) => {
-                        extension.props.store.dispatch(actions.main.onQueueItemChange(newValue));
+            const callback = (extension, scope) => {
+                extension.props.storePromise = new Promise((resolve) => {
+                    sdPlanningStore.initWorkspace(WORKSPACE.AUTHORING, (store) => {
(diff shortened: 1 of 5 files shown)


client/index.ts
```diff
@@ -105,15 +105,22 @@ export default angular.module('superdesk-planning', [])
...
+                extension.props.storePromise = new Promise((resolve) => {
+                    sdPlanningStore.initWorkspace(WORKSPACE.AUTHORING, (store) => {
+                        store.dispatch(actions.fetchAgendas());
+                        actions.main.onQueueItemChange(scope.selected?.preview);
+                        scope.$watch('selected.preview', (newValue) => {
+                            store.dispatch(actions.main.onQueueItemChange(newValue));
+                        });
+
+                        // Use a timeout as resolving the promise straight away causes the app to crash
+                        setTimeout(() => resolve(store));
```
eos87, line 118: Ideally, we should find the root cause of why the app crashes without the `setTimeout`, since needing a `setTimeout` is usually a symptom of an underlying issue. But if it’s too tricky to figure out right now, we can leave it as is for the moment.

eos87 (approved):
Looks good, I just left one comment but, if it's too tricky to figure out we can leave for later investigation.

client/index.ts
MarkLark86: Unfortunately I was unable to fix it. I think the problem stems from the bridge between AngularJS and React codes.
When we remove AngularJS from the Publish Queue page, we should be able to fix this (and improve this PRs code, as it is hacky already just to make it work properly).

PR #2611, opened 2025-11-17
fix missing planning items in the list

petrjasek:
they were filtered out due to missing end date

STT-1470

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/utils/index.ts
+++ b/client/utils/index.ts
@@ -672,7 +672,13 @@ export const isDateInRange = (inputDate, startDate, endDate, allDay = false) => 
     }
 
     if (allDay) {
-        return moment.utc(inputDate).isBetween(startDate, endDate, 'day', '[)');
+        if (startDate && inputDate.isBefore(startDate, 'day') ||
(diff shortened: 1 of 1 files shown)


client/utils/index.ts
```diff
@@ -672,7 +672,13 @@ export const isDateInRange = (inputDate, startDate, endDate, allDay = false) =>
     }
 
     if (allDay) {
-        return moment.utc(inputDate).isBetween(startDate, endDate, 'day', '[)');
+        if (startDate && inputDate.isBefore(startDate, 'day') ||
+            endDate && inputDate.isSameOrAfter(endDate, 'day')
```
thecalcc, line 676: if it's all day isn't it single day? or it could also be multi day, and then we would filter out the last item ending on e.g. 18th if input date is also 18th?

thecalcc (approved):

client/utils/index.ts
petrjasek: all planning items are single day

PR #2617, opened 2025-11-19
Action modals improvements

thecalcc:
### Purpose
Make three dots menu action modals more consistent.

### What has changed
Change styles, rework agendas modal to use generic crud form, add preview/edit states, add click/doubleclick actions. 

Resolves: [SDESK-7267]

[SDESK-7267]: https://sofab.atlassian.net/browse/SDESK-7267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eos87:
Hi @thecalcc, there are some tests failing. Also, could you please add a couple of screenshots like before/after? Thanks!

thecalcc:
> Hi @thecalcc, there are some tests failing. Also, could you please add a couple of screenshots like before/after? Thanks!

yes, I'm waiting for test instances to come back up so I can record a video

thecalcc:
Before:

https://github.com/user-attachments/assets/f26d2cf8-14ef-482c-accf-f5ed337a3efc

After: 

https://github.com/user-attachments/assets/3860bebf-e4c1-4e5c-833b-d515f6f1b72a

Lots of changes around usability and consistency, all has been synced with @fritzSF

eos87:
@thecalcc thanks for adding the videos. What about the error shown in the second video 😅?

client/components/Agendas/AgendaListItem.tsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import moment from 'moment';
+import {IAgendaEntity} from '../../interfaces';
+import {IPropsGenericFormItemComponent} from 'superdesk-api';
+import {ListItem, ListItemColumn, ListItemActionsMenu} from 'superdesk-core/scripts/core/components/ListItem';
+import {getFormFieldPreviewComponent} from 'superdesk-core/scripts/core/ui/components/generic-form/form-field';
+import {gettext} from '../../utils';
+import {getNameField} from './ManageAgendasModal';
+import {IconButton, Label} from 'superdesk-ui-framework';
```
eos87: do we still need to import from 'superdesk-ui-framework/react'?
```suggestion
import {IconButton, Label} from 'superdesk-ui-framework/react';
```

thecalcc:
> @thecalcc thanks for adding the videos. What about the error shown in the second video 😅?

Good point, it's because of the promise reject, doesn't really affect anything. Regardless, I'll address it after I fix E2E, which comes after I can fix the E2E on the client-core PR and update deps here. Wanted to get you videos before that.

client/components/Agendas/index.tsx
```diff
@@ -1,5 +1 @@
-export {AgendaList} from './AgendaList';
-export {AgendaItem} from './AgendaItem';
 export {AgendaNameList} from './AgendaNameList';
-export {EditAgenda} from './EditAgenda';
-export {ManageAgendasModal} from './ManageAgendasModal';
```
eos87, line 5: are we sure these are not used anywhere else?

client/components/Events/ManageEventTemplatesModal.tsx
```diff
@@ -17,32 +18,35 @@ interface IEventTemplate extends IBaseRestApiResponse {
...
+                        <div style={{display: 'flex'}}>
+                            <IconButton
+                                icon="pencil"
+                                onClick={() => page.startEditing(item._id)}
+                                ariaValue={superdeskApi.localization.gettext('')}
+                                size="small"
+                            />
+                            <IconButton
+                                icon="trash"
+                                onClick={() => page.deleteItem(item._id)}
```
eos87: I saw two ways of calling `page.deleteItem`, one with `_id` and another only with `item` somewhere above. Can you ensure we're passing the right param on each cases?
```
                                onClick={() => page.deleteItem(item._id)}
```

client/components/Agendas/ManageAgendasModal.tsx
```diff
@@ -1,169 +1,102 @@
...
+            >
+                <AgendasView
+                    ItemComponent={AgendaListItem}
+                    getFormConfig={() => this.config}
+                    fieldForSearch={getNameField()}
+                    getId={(item) => item._id}
+                    defaultSortOption={{field: 'is_enabled', direction: 'descending'}}
+                    hideItemsCount
+                    disallowCreatingNewItem={
+                        this.props.privileges[PRIVILEGES.AGENDA_MANAGEMENT] != 1 ? true : undefined
```
eos87: maybe we should have some helper like `hasPrivilege(): boolean` so we don't have to use a "magic number" here (`1`). What do you think?

eos87 (commented):
It looks good, I left some minor comments. I was thinking if we should instead have created this PR against `develop` given that this is a significant refactor and `release/3.1` is what STT is using in production already. Thoughts?

client/components/Agendas/index.tsx
thecalcc: yes

client/components/Agendas/AgendaListItem.tsx
thecalcc: yes, I just keep forgetting it since it's manual. In some cases the app just crashes at runtime if we don't so I always try to do it

client/components/Agendas/ManageAgendasModal.tsx
thecalcc: good point, with that we can refactor all other usages which are a lot probably

thecalcc:
> It looks good, I left some minor comments. I was thinking if we should instead have created this PR against `develop` given that this is a significant refactor and `release/3.1` is what STT is using in production already. Thoughts?

i think the same but since it's planned for release/3.2 i'm not sure what the flow is in terms of branches. I'll see how easy it is to move it to develop

client/components/Agendas/ManageAgendasModal.tsx
thecalcc: actually doesn't need to be, rather we use the `1` for just making the key available in the object. if there's no privilege for e.g. agenda mgmt we just don't have that key, so a `== null` check is fine in this case

eos87: if property `disallowCreatingNewItem` expects a boolean, it could simply become something like

```tsx
disallowCreatingNewItem={hasPrivilege(privileges, AGENDA_MANAGEMENT)}
```

the helper would return a boolean. Inside the helper `hasPrivilege` (or any other more meaningful name) would check against `1` which could be made a constant to give it more meaning instead of being a magic number. Does that make sense?

client/components/Agendas/AgendaListItem.tsx
eos87: so we should import from `superdesk-ui-framework/react` correct?

client/components/Agendas/ManageAgendasModal.tsx
thecalcc: yep that function makes sense, I'll do it, but for the question about `disallowCreatingNewItem` expected type, it's actually `true` or `undefined`. that's why I have to have it like I have it now

thecalcc:
Note: I've moved this unit test for agendas to an E2E one since we haven't setup infra to use the whole `superdeskApi` properly in unit tests thus you can't easily mock a whole component coming from it.

client/components/EventsPlanningFilters/EditFilter.tsx
```diff
@@ -51,6 +51,25 @@ export class EditFilter extends React.Component<IEventsPlanningContentPanelProps
...
+                filter: filter,
+                invalid: false,
+                errors: {},
+                profile: this.getProfile(filter.item_type),
+            }, () => {
+                this.props.onPristineChange?.(true);
+            });
+        }
+    }
+
```
eos87: I'm wondering if instead of this change we could use a `key` prop to reset the state without the lifecycle tricks. You know what I mean? Thoughts?

client/components/EventsPlanningFilters/ManageFiltersModal.tsx
```diff
@@ -71,13 +72,20 @@ export class ManageFiltersComponent extends React.Component<IProps, IState> {
...
         this.editFilter = this.editFilter.bind(this);
+        this.editFilterWithConfirmation = this.editFilterWithConfirmation.bind(this);
         this.previewFilter = this.previewFilter.bind(this);
         this.editFilterSchedule = this.editFilterSchedule.bind(this);
         this.closeEditor = this.closeEditor.bind(this);
         this.handleKeydown = this.handleKeydown.bind(this);
+        this.onPristineChange = this.onPristineChange.bind(this);
+        this.onDontSave = this.onDontSave.bind(this);
+        this.onSaveAndSwitch = this.onSaveAndSwitch.bind(this);
+        this.onCancelSwitch = this.onCancelSwitch.bind(this);
```
eos87: nit: we could remove this and make it cleaner by declaring the members with arrow functions.

client/components/Agendas/AgendaListItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export class AgendaListItem extends React.PureComponent<IPropsGenericFormItemComponent<IAgendaEntity>> {
+    render() {
+        const {item, page, inEditMode, inPreviewMode} = this.props;
+        const borderState = item.is_enabled || item.is_enabled == null ? 'active' : 'idle';
+
+        return (
+            <ListItem
+                onClick={() => page.openPreview(item._id)}
+                onDoubleClick={() => page.startEditing(item._id)}
```
eos87: we need to avoid passing arrow functions as properties when possible. Here, these two, it could be easily extracted into a stable method which would grab the items from props. 

```typescript
handleClick = () => {
    const {item, page} = this.props;
    page.openPreview(item._id)
}
```

same with double click. I've seen a couple of case like this below too. Also, a function component could be used here instead

I mean, it could be nothing burger but who knows, small performance optimizations are needed these days. 
Thoughts?

thecalcc: i agree on the performance, it's not about it being small or big, but we need to be mindful, and in this case it's an easy benefit. i'll change it to a functional component and define functions before the render method

client/components/EventsPlanningFilters/EditFilter.tsx
thecalcc: yep good point, i'll do it, shouldn't sacrifice too much readability/understanding for other devs

eos87 (approved):
:ship: it

PR #2620, opened 2025-11-26
[SDESK-7522] Add initial Playwright framework and tests

MarkLark86:
### Purpose
Support creating end-to-end tests using Playwright.

### What has changed
* Create initial Playwright test file structure
* Port Cypress utilities to Playwright
* Move basic Event & Planning edit e2e tests to Playwright

Resolves: [SDESK-7522](https://sofab.atlassian.net/browse/SDESK-7522)

[SDESK-7522]: https://sofab.atlassian.net/browse/SDESK-7522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


e2e/playwright/utils/common/inputs/index.ts
```diff
@@ -0,0 +1,17 @@
+export {Input} from './input';
+export {ToggleInput} from './toggleInput';
+export {SelectInput} from './selectInput';
```
thecalcc, line 3: not sure if we need this export file at all, maybe since it's a cleanup we can get rid of this also

e2e/playwright.config.ts
```diff
@@ -0,0 +1,54 @@
...
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+    testDir: './playwright',
+    /* Run tests in files in parallel */
+    fullyParallel: false,
+    /* Fail the build on CI if you accidentally left test.only in the source code. */
+    forbidOnly: !!process.env.CI,
+    /* Retry on CI only */
+    retries: 0,
```
thecalcc, line 13: i think when in CI env we might find it useful to retry 1-2 more times, given the historical flakiness, other option might be to go hard on the tests and try to get rid of all flakiness

thecalcc (approved):
Looking good!! pretty cool we'll have a unified testing infra.  I saw `any` types in a couple places in the utils, not sure @MarkLark86  if the intent is to be more progressive with this whole migration and polish as we go.

also seems like regular e2e CI is stuck?

e2e/playwright.config.ts
eos87: yeah I rather go second option, get rid of flakiness and make tests faster 🚀

e2e/playwright/utils/common/inputs/index.ts
MarkLark86: This allows to import multiple input helpers in the one line, such as:
```ts
import {Input, SelectMetaTerms, ToggleInput, UrgencyTreeSelectInput} from '../../common';
```

Otherwise it would need to be 1 per import:
```ts
import {Input} from '../../common/input';
import {SelectMetaTerms} from '../../common/selectMetaTerms';
import {ToggleInput} from '../../common/toggleInput';
import {UrgencyTreeSelectInput} from '../../common/urgencyTreeSelectInput';
```

This makes the `inputs` a module, allowing to import directly from it and not the implementation file.

e2e/playwright.config.ts
MarkLark86: This is the same config from client-core. And in my opinion, if we find flakiness we should fix it.
@eos87 thoughts?

MarkLark86:
> Looking good!! pretty cool we'll have a unified testing infra. I saw `any` types in a couple places in the utils, not sure @MarkLark86 if the intent is to be more progressive with this whole migration and polish as we go.
> 
> also seems like regular e2e CI is stuck?

I'll have a look at the `any` usage in my next PR. As for the older e2e CI stuck, that's because the test name has changed. I'll fix it once this is merged

e2e/playwright/utils/common/inputs/index.ts
thecalcc: yep i see it being useful, though that we don't want to use it anymore after talking with @eos87, but I might have gotten it wrong

e2e/playwright.config.ts
```diff
@@ -0,0 +1,54 @@
...
+    use: {
+        /* Base URL to use in actions like `await page.goto('/')`. */
+        baseURL: 'http://localhost:9000',
+
+        viewport: {width: 1280, height: 800},
+
+        /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+        trace: 'retain-on-failure',
+
+        screenshot: 'only-on-failure',
```
eos87, line 28: we should add our test id attribute so we can use `getByTestId` more often and keep our tests more stable. 

```suggestion
        screenshot: 'only-on-failure',
        
        /* our custom test id attribute */
        testIdAttribute: 'data-test-id',
```

e2e/playwright/events/edit_event.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+        // Make sure both recurring Events now have the 'Scheduled' badge
+        await expect(
+            list.item(0).locator('.label--success'),
+        ).toContainText('Scheduled');
+        await expect(
+            list.item(1).locator('.label--success'),
+        ).toContainText('Scheduled');
+    });
+
+    test('SDESK-5982: Post updates the initial values', async ({page}) => {
```
eos87: probably we don't need the ticket number here
```suggestion
    test('Post updates the initial values', async ({page}) => {
```

e2e/playwright/planning/edit_planning.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+            description_text: 'Other Desc. Text',
+            internal_note: 'Other Int. Note',
+            ednote: 'Other Ed. Note',
+        });
+        await editor.waitForAutosave();
+        await editor.waitTillOpen();
+        await expect(editor.postButton).toBeVisible();
+    });
+
+    test('SDESK-5982: Post updates the initial values', async () => {
```
eos87: same here
```suggestion
    test('Post updates the initial values', async () => {
```

e2e/playwright/utils/common/app.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Changes the current workspace by interacting with the workspace navigation element on the given page.
+ *
+ * @param {Page} page - The Playwright Page object representing the active browser page.
+ * @param {string} name - The name of the workspace to navigate to.
+ * @return {Promise<void>} A promise that resolves when the workspace has been successfully changed.
+ */
+export async function changeWorkspace(page: Page, name: string): Promise<void> {
+    await page.locator('[data-test-id=workspace-navigation]')
+        .locator(`[data-test-id=${name}]`)
```
eos87: we could use `getByTestId` here

e2e/playwright/utils/common/editor.ts
```diff
@@ -0,0 +1,144 @@
...
+        await clickAll(this.element, '.toggle-box.toggle-box--circle.hidden');
+    }
+
+    async clickBookmark(bookmarkId: string): Promise<void> {
+        await this.element.locator(`[data-test-id="editor--bookmarks__${bookmarkId}"]`).click();
+    }
+
+    getFormGroup(groupId: string): Locator {
+        return this.element.locator(`[data-test-id="editor--group__${groupId}"]`);
+    }
```
eos87, line 143: nit:
```suggestion
    async clickBookmark(bookmarkId: string): Promise<void> {
        await this.element.getByTestId(`editor--bookmarks__${bookmarkId}`).click();
    }

    getFormGroup(groupId: string): Locator {
        return this.element.getByTestId(`editor--group__${groupId}`);
    }
```

e2e/playwright/utils/common/inputs/linkInput.ts
```diff
@@ -0,0 +1,35 @@
+import {Locator, expect} from '@playwright/test';
+import {Input} from './input';
+
+export class LinkInput extends Input {
+    get addButton(): Locator {
+        return this.parent.locator('[data-test-id="event-links__add-new-button"]');
```
eos87: nit:
```suggestion
        return this.parent.getByTestId('event-links__add-new-button');
```

eos87 (approved):
Great job, Mark! I left a few minor comments. In general we should prefer using `data-test-id` and `getByTestId` instead of css selector to keep our tests more stable but we can polish that more along the way. Thanks for taking care of this 👏

e2e/playwright.config.ts
eos87: Totally agree. I just added one more config to use our custom test id attribute `data-test-id`

e2e/playwright/utils/common/inputs/index.ts
eos87: given these are the test utils we are going to be reusing throughout the tests I think it's good to have it so the imports look cleaner.

PR #2624, opened 2025-11-26
fix reducer creating new state on unknown action

petrjasek:
happens a lot on eg. websocket notifications

[SDCP-1037]

[SDCP-1037]: https://sofab.atlassian.net/browse/SDCP-1037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/constants/index.ts
+++ b/client/constants/index.ts
@@ -69,6 +69,7 @@ export const SPIKED_STATE = {
 
 export const RESET_STORE = 'RESET_STORE';
 export const INIT_STORE = 'INIT_STORE';
+export const INITIAL_STATE = '_INITIAL_STATE';
 export const FORM_NAMES = {
     PlanningForm: 'planning',
     EventForm: 'event',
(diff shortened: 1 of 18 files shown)


eos87 (approved):
Looks good. I kinda worries me a little bit the shallow copying of the initial state but it should be ok I think.

thecalcc (approved):

thecalcc:
not sure why `@sourcefabric/date-fns-tz` doesn't work in node 14, i remember doing a fix some time ago in the webpack config

petrjasek:
> not sure why `@sourcefabric/date-fns-tz` doesn't work in node 14, i remember doing a fix some time ago in the webpack config

it actually works when building for e2e, but when building for karma it might have a different config. but I wouldn't spend time on fixing 14 now anyhow

PR #2625, opened 2025-11-28
[STT-1491] - Fix pagination infinite loop

eos87:
### What has changed
- Prevent loading more items once all list items are fetched by also checking isAllListItemsLoaded in the scroll handler.
- Ensure isNextPageLoading is only reset after loadMore completes by moving the call into a setState callback.
- Refactor eventsplanning reducer cases to use immer’s produce instead of object spreads, cloneDeep, and uniq.
- Keep eventsAndPlanningInList deduplicated by manually checking for existing IDs before pushing new ones.

Resolves: [STT-1491]

[STT-1491]: https://sofab.atlassian.net/browse/STT-1491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/reducers/eventsplanning.ts
+++ b/client/reducers/eventsplanning.ts
@@ -85,10 +85,12 @@ const eventsPlanningReducer = createReducer<IEventsPlanningState>(initialState, 
             draft.currentFilter = payload === EVENTS_PLANNING.FILTER.ALL_EVENTS_PLANNING ? null : payload;
         }),
 
-    [MAIN.ACTIONS.CLEAR_SEARCH]: (state, payload: keyof IMainState['search']) =>
-        payload !== 'COMBINED' ? state : produce(state, (draft) => {
-            draft.currentFilter = null;
-        }),
+    [MAIN.ACTIONS.CLEAR_SEARCH]: (state, payload: keyof IMainState['search']) => (
+        payload !== 'COMBINED' ? state : {
+            ...state,
+            currentFilter: null,
+        }
+    ),
 });
 
 export default eventsPlanningReducer;
--- a/client/reducers/eventsplanning.ts
+++ b/client/reducers/eventsplanning.ts
@@ -1,4 +1,5 @@
-import {cloneDeep, get, uniq, sortBy} from 'lodash';
+import {get, sortBy} from 'lodash';
+import {produce} from 'immer';
 
 import {IEventsPlanningState, IMainState, ISearchFilter} from '../interfaces';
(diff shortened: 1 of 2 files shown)


client/reducers/eventsplanning.ts
```diff
@@ -35,66 +36,59 @@ const eventsPlanningReducer = createReducer<IEventsPlanningState>(initialState,
...
+        produce(state, (draft) => {
+            const newIds = (payload || []).map((e) => e._id);
+            const existingIds = new Set(draft.eventsAndPlanningInList || []);
+
+            newIds.forEach((id) => {
+                if (!existingIds.has(id)) {
+                    draft.eventsAndPlanningInList.push(id);
+                }
+            });
+        }),
```
eos87, line 54: Most relevant piece is this one. Before we were replacing the entire array and now we're just adding those that do not exist in the list yet.

client/reducers/eventsplanning.ts
```diff
@@ -35,66 +36,59 @@ const eventsPlanningReducer = createReducer<IEventsPlanningState>(initialState,
...
+            draft.filters = sortBy([...payload], [(f) => f.name.toLowerCase()]);
+        }),
+
+    [EVENTS_PLANNING.ACTIONS.SELECT_EVENTS_PLANNING_FILTER]: (state, payload: ISearchFilter['_id']) =>
+        produce(state, (draft) => {
+            draft.currentFilter = payload === EVENTS_PLANNING.FILTER.ALL_EVENTS_PLANNING ? null : payload;
+        }),
+
+    [MAIN.ACTIONS.CLEAR_SEARCH]: (state, payload: keyof IMainState['search']) =>
+        payload !== 'COMBINED' ? state : produce(state, (draft) => {
```
petrjasek: I saw some immer overhead so might make sense to handle those simple changes via:
```
payload !== 'COMBINED' ? state : {...state, currentFilter: null};
```
but it might be handling those small changes well, just wondering

thecalcc (approved):
cool stuff, one thing to note is that we can avoid doing: 

```js 
(state, payload) => 
produce(state, (draft) => {
...
```

by using the shorthand - https://immerjs.github.io/immer/example-setstate/#redux--immer

petrjasek (approved):

client/reducers/eventsplanning.ts
eos87: just wanted it to be consistent to you're right, kinda overkill. Let me adjust it for this one.

PR #2627, opened 2025-12-02
[SDESK-7522] Migrate Cypress E2E tests to Playwright

MarkLark86:
Resolves: [SDESK-7522](https://sofab.atlassian.net/browse/SDESK-7522)

[SDESK-7522]: https://sofab.atlassian.net/browse/SDESK-7522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86:
Hey @eos87 & @thecalcc I had to disable 3 tests as I could not figure out what is causing them to fail in github/CI but pass locally. So that I can get on with other tasks, and as this is for v3.2, we can come back at a later date and fix these.

On a positive note, I enabled the accessibility tests which was previously disabled in Cypress. So over all, we actually have more E2E tests running now :smile: 

Also the `cypress-a`, `cypress-b` and `playwright` actions can be ignore, I reverted back to using the `a` and `b` E2E matrix in the config

e2e/playwright/assignments/edit_priority.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        await editor.waitLoadingComplete();
+
+        await coverageEditor.element.click();
+        await expect(
+            coverageEditor.element.locator('.sd-collapse-box__content-block--top')
+        ).toContainText('Politic Desk');
+        await expect(
+            coverageEditor.element
+                .locator('.sd-collapse-box__content-block--top')
+                // .locator('.label')
```
thecalcc: dead code

e2e/playwright/assignments/edit_priority.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        await coverageEditor.toggleAddToWorkflow();
+
+        await editor.saveButton.click();
+
+        await editor.waitForAutosave();
+        await expect(
+            coverageEditor.element
+                .locator('.sd-collapse-box__content-block--top')
+                .locator('.label')
```
thecalcc, line 63: i guess uncommenting at the previous line would work given this checks for a different string in the same manner?

e2e/playwright/events/event_action_create_planning.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+        await editors.planning.type({
+            internal_note: 'Created from Event',
+        });
+
+        // Test that we can save the Planning item
+        await editors.planning.waitForAutosave();
+        await editors.planning.saveButton.click();
+
+        // Wait for autosave to be completed
+        await editors.planning.waitForAutosave();
```
thecalcc: if we're waiting for save beforehand shouldn't that also flush the autosave, so no need to wait for it again? or autosave and save is coupled here?

e2e/playwright/utils/common/inputs/contactsInput.ts
```diff
@@ -34,7 +36,8 @@ export class ContactsInput extends Input {
     async remove(index: number): Promise<void> {
         await this.list
             .nth(index)
-            .locator('.icon-trash')
+            .getByRole('button', {name: 'Remove Contact', exact: true})
+            // .locator('.icon-trash')
```
thecalcc: dead code

e2e/playwright/utils/common/inputs/contactsInput.ts
```diff
@@ -44,10 +47,13 @@ export class ContactsInput extends Input {
...
     async editContact(index: number): Promise<void> {
+        await this.list.nth(index).hover();
         await this.list
             .nth(index)
-            .locator('.icon-pencil')
-            .click({force: true}); // the icon is only shown on hover
+            .getByRole('button', {name: 'Edit Contact', exact: true})
+            // .locator('.icon-pencil')
+            .click(); // the icon is only shown on hover
+            // .click({force: true}); // the icon is only shown on hover
```
thecalcc: dead code

e2e/playwright/utils/common/inputs/treeSelect.ts
```diff
@@ -22,10 +22,10 @@ export class TreeSelect extends Input {
         if (this.allowMultiple) {
             for (const val of values) {
                 await this.addButton.click();
-                await this.page.keyboard.type(val);
                 await this.page
-                    .locator('[data-test-id="tree-select-popover"] ul li')
-                    .first()
+                    .getByTestId('tree-select-popover')
```
thecalcc, line 26: in the future it'd probably be good to get this whole tree select functionality into a "driver" utility so we can reuse it across everywhere

e2e/playwright/utils/common/ui/subNavBar.ts
```diff
@@ -33,14 +33,16 @@ export class SubNavBar {
     async createEvent(): Promise<void> {
         await this.plusBtn.click();
         await this.createMenu
-            .locator('#create_event')
+            .getByRole('button', {name: 'Event', exact: true})
+            // .locator('#create_event')
```
thecalcc: dead code

thecalcc (approved):
legendary change log, i'm nitpicking with some comments on dead code, but this is great stuff. I'll try out the distrobox setup at some point, and I think we could benefit from the same migration in client core and have everything else completely gone

Also, what do you guys @eos87 @MarkLark86  think in terms of making tests consistent for UI framework related code - maybe we can have a "driver" for each bigger component that are defined in that project and reused in client core and here?

thecalcc:
by the way sorry @MarkLark86 to keep you waiting on the agendas E2E test that I did, but it seems flaky and while locally it passes, through GH actions it doesn't ;D

Edit: it's actually not flaky, rather seems that GH actions instance uses some other client core which has outdated data test ids for some elements

.github/workflows/ci-e2e.yml
```diff
@@ -47,31 +47,16 @@ jobs:
...
-      - name: Run Playwright tests
-        if: matrix.e2e == 'playwright'
-        run: npx playwright test
+        run: npx playwright install --with-deps chromium
+      - name: E2E Suite A
+        if: matrix.e2e == 'a'
+        run: npm run playwright -- --grep events
+      - name: E2E Suite B
+        if: matrix.e2e == 'b'
+        run: npm run playwright -- --grep-invert events
```
eos87, line 56: why do we have two suits for this?

e2e/index.js
```diff
@@ -1,6 +1,34 @@
...
+    //     animation-duration: 0s !important;
+    //     transition-duration: 0s !important;
+    // }`;
+    //
+    // const styleSheet = document.createElement("style");
+    // styleSheet.textContent = styles;
+    //
+    // // Use Arial font in tests.
+    // // With Roboto screenshots don't match when generated using vscode extension vs headless run
+    // document.head.appendChild(styleSheet);
```
eos87: I imagine we can remove this

eos87 (approved):
The PR is very fat which makes is a bit hard to review. We should maybe aim for multiple smaller ones whenever possible. It looks great though, I just left a couple of minor comments. Great job and thanks for taking care of this 👏

eos87:
there are some conflicts, I assume they are because of this PR https://github.com/superdesk/superdesk-planning/pull/2617

.github/workflows/ci-e2e.yml
MarkLark86: So that can run tests concurrently, as each suite takes 8+ minutes to run.

We should ideally do the build step once, store the generated `dist` folder, and have e2e steps use the build step, so we aren't downloading, compiling and bundling multiple times.

MarkLark86:
> Also, what do you guys @eos87 @MarkLark86 think in terms of making tests consistent for UI framework related code - maybe we can have a "driver" for each bigger component that are defined in that project and reused in client core and here?

Agree'd @thecalcc. Let's have a meeting next week to discuss our desires and options regarding this.

MarkLark86:
> The PR is very fat which makes is a bit hard to review. We should maybe aim for multiple smaller ones whenever possible. It looks great though, I just left a couple of minor comments. Great job and thanks for taking care of this 👏

My apologies. As this work was not about re-creating e2e from scratch but just to migrate what we already had, I thought it would be okay. If tests pass, then we should be in the same point as before this PR. That was my thinking

.github/workflows/ci-e2e.yml
eos87: I realized that after asking 😅 thanks for clarifying.

eos87:
> > The PR is very fat which makes is a bit hard to review. We should maybe aim for multiple smaller ones whenever possible. It looks great though, I just left a couple of minor comments. Great job and thanks for taking care of this 👏
> 
> My apologies. As this work was not about re-creating e2e from scratch but just to migrate what we already had, I thought it would be okay. If tests pass, then we should be in the same point as before this PR. That was my thinking

Nothing to apologize. I think you're totally right, my comment seems kinda silly now that you explained it, makes total sense.

PR #2635, opened 2025-12-10
[STT-1516] Improve web request performance for Assignments resource

MarkLark86:
### Purpose
While processing a web request for Assignments, we're getting the list of linked content items (archive, publised etc). The issue is we're retrieving the entire content items from the DB when we only need just a couple fields, which increases the processing required for the elasticsearch json response.

### What has changed
When enhancing Assignment items on fetch:
* Disable aggregations, as they are not needed
* Use projection to define what fields we need
* Tell `search` endpoint not to run it's on fetched signals.

Depends on: https://github.com/superdesk/superdesk-core/pull/3091
Resolves: [STT-1516](https://sofab.atlassian.net/browse/STT-1516)

[STT-1516]: https://sofab.atlassian.net/browse/STT-1516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -118,7 +118,7 @@ class AssignmentsService(AsyncBaseService):
     async def _enhance_assignments(self, docs):
         """Populate `item_ids` with ids for all linked Archive items for an Assignment"""
         assignment_archive_map: dict[str, tuple[list[str], list[dict]]] = {}
-        async for item in await self.get_archive_items_for_assignments([doc.get(ID_FIELD) for doc in docs]):
+        async for item in await self.get_archive_links_for_assignments([doc.get(ID_FIELD) for doc in docs]):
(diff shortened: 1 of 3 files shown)


server/planning/assignments/assignments.py
```diff
@@ -117,24 +117,28 @@ async def on_fetched_item_async(self, doc):
...
             self.set_type(doc, doc)
 
+            try:
+                linked_item_ids, linked_items = assignment_archive_map[str(doc.get("_id"))]
+                doc["item_ids"] = linked_item_ids
+                doc["linked_items"] = linked_items
+            except KeyError:
+                pass
+
     async def get_archive_items_for_assignments(self, assignment_ids):
```
petrjasek: I think now the name is a bit off with what it does, should be probably something like `get_archive_links_for_assignments`

petrjasek (approved):

PR #2636, opened 2025-12-10
[SDBELGA-1010] - As an admin ingesting Events in Json format I want to also have attachments uploaded to SD

BrianMwangi21:
### Purpose
This PR support ingesting json events with attachments in `EventJsonFeedParser`, enabling upload of attachments referenced in the event json

### What has changed
- Parser now resolves files entries which were initially ignored. It checks for existing IDs first, then if not, uploads local files relative to the json location
- Updated tests to test new attachment ingestion using `attachments/sunset.jpg`

### Steps to test
1. Create new ingest source which uses file feed service and Event Json as the parser. 
2. In the specified server folder, add the event_json file. Then create an attachments folder and add an attachment such as an image. Make sure the event json file attribute has a value such as `files: ['attachments/sunset.jpg']`
3. Activate the ingest source and observe the event is ingested and the attachment section has the file specified.
4. Else, you can run the update tests `event_json_tests.py.`

Resolves: #[SDBELGA-1010](https://sofab.atlassian.net/browse/SDBELGA-1010)

[SDBELGA-1010]: https://sofab.atlassian.net/browse/SDBELGA-1010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/feed_parsers/superdesk_event_json.py
```diff
@@ -163,6 +169,89 @@ def add_to_local_db(self, superdesk_event):
...
+            else:
+                logger.warning("Skipping unsupported file entry type: %s", type(entry))
+                continue
+
+            file_path_to_use = filename
+            if not os.path.isabs(file_path_to_use):
+                file_path_to_use = os.path.join(base_dir, filename)
+
+            try:
+                with open(file_path_to_use, "rb") as content:
```
petrjasek: imo it shouldn't be on the parser to handle the files, that's kind of service specific (eg. ftp will fetch it from there), so would rather pass the feeding service to parser and implement there some `fetch_file` method to get the binary

server/planning/feeding_services/event_file_service.py
petrjasek: instead of making new service instance in the parser it should rather pass it from there:
```
item = parser.parse(file_path, provider, feeding_service=self)
```
so it's already configured

server/planning/feeding_services/event_file_service.py
```diff
@@ -89,7 +89,7 @@ def _update(self, provider, update):
                             with open(file_path, "rb") as f:
                                 item = parser.parse_file(f, provider)
```
petrjasek: would be good to pass it also here

server/planning/feed_parsers/event_json_tests.py
```diff
@@ -99,3 +102,86 @@ def test_event_json_feed_parser_parse(self):
...
+                json.dump(sample, tmp)
+                tmp_path = tmp.name
+
+            feeding_service = EventFileFeedingService()
+
+            try:
+                events = EventJsonFeedParser().parse(tmp_path, provider, feeding_service=feeding_service)
+                self.assertNotIn("files", events[0])
+            finally:
+                os.remove(tmp_path)
```
petrjasek: personally would rather keep it inside that `with tempfile` block and let it delete it, is there any benefit to use this try/finally?

server/planning/feed_parsers/event_json_tests.py
```diff
@@ -99,3 +102,86 @@ def test_event_json_feed_parser_parse(self):
...
+            try:
+                events = EventJsonFeedParser().parse(tmp_path, provider, feeding_service=feeding_service)
+                files = events[0].get("files", [])
+                self.assertEqual(2, len(files))
+                self.assertEqual(existing_file_id, files[0])
+
+                new_file_id = files[1]
+                saved_file = events_files_service.find_one(req=None, _id=new_file_id)
+                self.assertIsNotNone(saved_file)
+                self.assertEqual("image/jpeg", saved_file.get("mimetype"))
```
petrjasek, line 143: would be also good to check if the binary was stored properly

server/planning/feed_parsers/superdesk_event_json.py
```diff
@@ -163,6 +166,51 @@ def add_to_local_db(self, superdesk_event):
...
         return superdesk_event
 
+    def _process_files(self, superdesk_event, file_path, feeding_service=None):
+        """Process event attachments by reusing existing IDs or uploading new files via the feeding service."""
+        files = superdesk_event.get("files")
+        if not files:
+            superdesk_event.pop("files", None)
+            return superdesk_event
+
+        base_dir = os.path.dirname(file_path)
```
petrjasek: I wouldn't pass this one, it's kind of filesystem specific, but that shouldn't be concern for the parser

server/planning/feed_parsers/superdesk_event_json.py
```diff
@@ -163,6 +166,51 @@ def add_to_local_db(self, superdesk_event):
...
+
+            if isinstance(entry, str):
+                try:
+                    object_id = ObjectId(entry)
+                    existing_file = events_files_service.find_one(req=None, _id=object_id)
+                    if existing_file:
+                        logger.info("Reusing existing event file %s", existing_file.get("_id"))
+                        processed_file_ids.append(existing_file.get("_id"))
+                        continue
+                except Exception:
```
petrjasek: can you use more specific exception here?

server/planning/feeding_services/event_file_service.py
```diff
@@ -101,3 +106,53 @@ def _update(self, provider, update):
                 raise ParserError.parseFileError("{}-{}".format(provider["name"], self.NAME), filename, ex, provider)
 
         push_notification("ingest:update")
+
+    def fetch_file(self, base_dir, filename):
```
petrjasek: base dir should be based on the ingest config imo, filename should do

server/planning/feeding_services/event_file_service.py
```diff
@@ -101,3 +106,53 @@ def _update(self, provider, update):
...
+
+        :param base_dir: Directory to resolve relative paths against
+        :param filename: Filename or absolute path to the file
+        :return: The created events_files document ID, or None on failure
+        """
+        file_path_to_use = filename
+        if not os.path.isabs(file_path_to_use):
+            file_path_to_use = os.path.join(base_dir, filename)
+
+        events_files_service = get_resource_service("events_files")
```
petrjasek: that dealing with event files is spread now in both parser and the service, should be in one place, probably in the parser if it's events specific one, this service can be type agnostic

server/planning/feeding_services/event_file_service.py
```diff
@@ -101,3 +106,53 @@ def _update(self, provider, update):
                 raise ParserError.parseFileError("{}-{}".format(provider["name"], self.NAME), filename, ex, provider)
 
         push_notification("ingest:update")
+
+    def fetch_file(self, base_dir, filename):
+        """
+        Fetch a local file, upload to media storage, and create an events_files record.
```
petrjasek: imo it should fetch local file, yield the binary, and then move the file like it does with the json.
all the other parts should be handled elsewhere, we can trigger it from the parser for now as it's events files specific, but would be better probably to have a method in `events_files` service, something like `ingest_file()` that would return media id

server/planning/feeding_services/event_file_service.py
```diff
@@ -107,52 +103,26 @@ def _update(self, provider, update):
...
-                payload = {"media": media_id, "mimetype": content_type}
-                if metadata:
-                    payload["filemeta"] = metadata
-
-                ids = events_files_service.post([payload])
-                saved_id = next(iter(ids or []), None)
-                if saved_id:
-                    logger.info("Attached event file %s as %s", file_path_to_use, saved_id)
-                    return saved_id
+            stream = open(file_path, "rb")
```
petrjasek: would rather do something like:
```py
with open(file_path, 'rb') as f:
    yield f
```
so it will be closed when the processing is done

server/planning/feeding_services/event_file_service.py
```diff
@@ -107,52 +103,26 @@ def _update(self, provider, update):
 
         push_notification("ingest:update")
 
-    def fetch_file(self, base_dir, filename):
+    def fetch_file(self, filename):
```
petrjasek: would be good to add types to new methods

server/planning/feeding_services/event_file_service.py
```diff
@@ -112,34 +115,31 @@ def _update(self, provider, update):
         self._move_attachment_files()
         push_notification("ingest:update")
 
-    def fetch_file(self, filename):
+    @contextmanager
```
petrjasek: I don't think context manager is needed, rather use for loop to consume the data

server/planning/feed_parsers/superdesk_event_json.py
```diff
@@ -195,13 +196,20 @@ def _process_files(self, superdesk_event, feeding_service=None):
                 logger.warning("Skipping unsupported file entry type: %s", type(entry))
                 continue
 
+            normalized = os.path.normpath(filename)
```
petrjasek: you can use [secure_filename](https://werkzeug.palletsprojects.com/en/stable/utils/#werkzeug.utils.secure_filename) on it, that should be enough

server/planning/feeding_services/event_file_service.py
```diff
@@ -37,6 +37,7 @@ def __init__(self):
         super().__init__()
         self._fetched_attachments = set()
         self._failed_attachments = set()
+        self._lock = threading.Lock()
```
petrjasek: you don't have to handle threading, we're not using multiple threads

server/planning/feeding_services/event_file_service.py
```diff
@@ -115,39 +117,45 @@ def _update(self, provider, update):
...
+        :return: Tuple of (stream, content_type) or None on failure
         """
         file_path = os.path.join(self.path, filename)
         try:
             with open(file_path, "rb") as f:
                 guessed_type = mimetypes.guess_type(file_path)[0]
                 content_type = guessed_type or "application/octet-stream"
-                self._fetched_attachments.add(filename)
-                yield (f, content_type)
+                stream = io.BytesIO(f.read())
```
petrjasek: it should still use yield imo, there is no need to read it here:
```py
with open(...) as f:
    yield f
```

in parser:
```
for file in service.fetch_file(filename):
    ...
```
would also consider moving that guess type logic from here to ingest_file method to simplify
the signature

server/planning/feeding_services/event_file_service.py
BrianMwangi21: I actually added that cause of this review https://github.com/superdesk/superdesk-planning/pull/2636#discussion_r2622284017 but its alright, I can remove it

petrjasek: yep I know, copilot is not aware of it probably

petrjasek (approved):

PR #2648, opened 2025-12-15
[SDBELGA-1011] - Allow users to edit coverage language from advanced modal

thecalcc:
### Purpose
Allow users to edit coverage language from advanced modal

### What has changed
Refactor the modal to use Modal component from UI framework. 
Separate editable fields into their own component.
Update save functionality to take into account the language field.
Add a new Select component for the language field.

### Screenshots
image

**Note:** I know that the `UserSelect` is inconsistent with the rest of the UI elements. This might be because it's using TreeSelect under the hood and the other elements are use `Select`, or because we render a `medium` sized `Avatar` component inside. The heights of these two are inconsistent by ±3 pixels. I will see if I can update the component in client-core without touching the TreeSelect itself. Otherwise we can make a new ticket to address this inconsistency issue as it's been there for a long time.

Resolves: #[SDBELGA-1011]

[SDBELGA-1011]: https://sofab.atlassian.net/browse/SDBELGA-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Coverages/CoverageAddAdvancedModal.tsx
+++ b/client/components/Coverages/CoverageAddAdvancedModal.tsx
@@ -2,169 +2,145 @@ import React from 'react';
 import {connect} from 'react-redux';
 import {cloneDeep, get} from 'lodash';
 
-import {IG2ContentType, IPlanningCoverageItem, IPlanningNewsCoverageStatus, IWorkflowState} from '../../interfaces';
+import {IG2ContentType, IPlanningNewsCoverageStatus, IPlanningCoverageItem} from '../../interfaces';
 import {IDesk, IUser} from 'superdesk-api';
 
 import {gettext, planningUtils, getUsersForDesk, getDesksForUser} from '../../utils';
 import {getUserInterfaceLanguageFromCV} from '../../utils/users';
 import {getVocabularyItemFieldTranslated} from '../../utils/vocabularies';
 
-import Modal from '../Modal';
-import {superdeskApi} from '../../superdeskApi';
+import * as selectors from '../../selectors';
 import * as actions from '../../actions';
-import {Button, Checkbox, IconButton, Option, Select, Spacer, Tooltip} from 'superdesk-ui-framework/react';
+
+import {Button, Checkbox, Modal, Spacer, Tooltip} from 'superdesk-ui-framework/react';
+import {CoverageEditableFields} from './CoverageFieldsRow';
 
 interface IReduxDispatchProps {
     setCoverageAddAdvancedMode(enable: boolean): void;
 }
 
-type IReduxStateProps = {};
+type IReduxStateProps = {
+    languages: Array<any>;
+};
+
+export interface ICoverageLineItem extends IPlanningCoverageItem {
+    enabled: boolean;
+    qcode: string;
+    desk: IDesk;
+    user: IUser;
+    status: IPlanningNewsCoverageStatus;
+    icon: string;
+    filteredDesks: Array<IDesk>;
+    filteredUsers: Array<IUser>;
+}
 
 interface IOwnProps {
     field: string;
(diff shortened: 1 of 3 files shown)


petrjasek:
we need this in release/2.11 @thecalcc

client/components/Coverages/CoverageFieldsRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    label={gettext('Desk')}
+                    value={coverage.desk?._id}
+                    onChange={(newDeskId) => {
+                        handleDeskChange(
+                            coverage,
+                            coverage.filteredDesks.find(({_id}) => _id === newDeskId),
+                        );
+                    }}
+                >
+                    <Option />
```
eos87, line 56: Is this just an empty option? if so, maybe we should add a label to it like: "Select an option"?

client/components/Coverages/CoverageFieldsRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    {coverage.filteredDesks.map((desk) => (
+                        <Option key={desk._id} value={desk._id}>{desk.name}</Option>
+                    ))}
+                </Select>
+                <div style={{width: '100%'}}>
+                    <SelectUser
+                        key={`${coverage.desk?._id}-${index}`}
```
eos87, line 63: is the key really needed here? it could end up being the same key (`-${index}`) for two (or more) different coverages if the desk is missing in them.

client/components/Coverages/CoverageFieldsRow.tsx
```diff
@@ -0,0 +1,122 @@
...
+                    onChange={(value) => {
+                        updateCoverage(coverage, {
+                            planning: {
+                                ...coverage.planning,
+                                language: value,
+                            }
+                        });
+                    }}
+                >
+                    {languages.map((cov) => (
```
eos87, line 87: This might crash if the languages are empty

client/components/Coverages/CoverageAddAdvancedModal.tsx
```diff
@@ -407,14 +264,66 @@ class CoverageAddAdvancedModalComponent extends React.Component<IProps, IState>
...
-                </Modal.Footer>
+                )}
+            >
+                <Spacer v gap="8" justifyContent="center" alignItems="center">
+                    {this.state.coverages.map((coverage, index) => {
+                        const isActive = coverage.workflow_status === 'active';
+
+                        return (
+                            <div
+                                key={index}
```
eos87, line 282: using loop indexes as keys is not recommended. We should instead use a stable value. Maybe like this it would be a bit better
```suggestion
                                key={coverage.coverage_id ?? index}
```

thecalcc:
closing in favour of https://github.com/superdesk/superdesk-planning/pull/2653

client/components/Coverages/CoverageAddAdvancedModal.tsx
thecalcc: it should be ok here since there's no reordering

client/components/Coverages/CoverageFieldsRow.tsx
thecalcc: that's good, if languages are not set the whole instance won't work properly. they should always be set beforehand

client/components/Coverages/CoverageFieldsRow.tsx
thecalcc: yeah I know, we had it before, I just moved it. I think there was a bug inside the `SelectUser` so that's why I or someone else introduced it in the first place, we have this because when you select a desk user might also change, etc. and in that case we should reset the user select

PR #2659, opened 2025-12-24
[STT-1513] - Planning item detail displays wrong "last modified by" information

BrianMwangi21:
### Purpose
Fix inconsistent "Updated by" information between Planning Details and History tabs. Coverage-only edits were overwriting planning editor attribution, causing user confusion.

### What has changed
Modified `version_creator` and `versioncreated` to only update when planning-level fields change. Coverage-only edits, system field changes, and assignment workflow actions no longer affect these fields

### Steps to test
1. Create a planning item as User A and add some planning details. Create the coverage as User A too. 
2. Switch to User B and edit ONLY coverage fields.
3. Verify Planning Details shows "Updated by User A" (not User B)
4. Verify History tab also shows last planning edit by User A
5. Both tabs should now display consistent information
6. You can also run tests: `pytest server/planning/tests/planning_attribution_test.py`

Resolves: #[STT-1513](https://sofab.atlassian.net/browse/STT-1513)

[STT-1513]: https://sofab.atlassian.net/browse/STT-1513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/planning/planning.py
+++ b/server/planning/planning/planning.py
@@ -123,8 +123,8 @@ class PlanningService(AsyncBaseService):
         update_ingest_on_patch(document, original)
 
         if self._should_update_planning_editor(document, original):
-            document["last_planning_editor"] = None
-            document["last_planning_edit_at"] = None
+            document["version_creator"] = None
+            document["versioncreated"] = None
 
         response = await self.backend.update_in_mongo_async(self.datasource, id, document, original)
         await self.on_updated_async(document, original, from_ingest=True)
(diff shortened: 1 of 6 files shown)


server/planning/planning/planning.py
```diff
@@ -120,6 +121,11 @@ async def patch_in_mongo(self, id, document, original):
         """Patch an ingested item onto an existing item locally"""
         prepare_ingested_item_for_storage(document)
         update_ingest_on_patch(document, original)
+
+        if self._should_update_planning_editor(document, original):
+            document["version_creator"] = None
+            document["versioncreated"] = None
```
petrjasek: not sure if this should be conditional, there is always a version in history created probably, so it should match that

petrjasek (commented):
can you make it on `release/3.2` branch? should be enough to edit the base here

server/planning/planning/planning.py
```diff
@@ -1838,6 +1846,33 @@ def inherit_planning_metadata(self, coverage, updates):
             if updates.get(field):
                 coverage["planning"].setdefault(field, updates[field])
 
+    @staticmethod
+    def _should_update_planning_editor(updates, original):
```
MarkLark86: This name is a bit misleading. At first I read `editor` as the front-end form. After looking at the code and its use, I can see it's related to the version creator information. Instead of `editor` in the name, maybe something with `version_creator`

MarkLark86 (approved):

PR #2663, opened 2026-01-07
add edit coverages action to planning items

petrjasek:
SDBELGA-1012

### Front-end checklist

Resolves: #[issue-number]


client/components/Planning/PlanningItem.tsx
```diff
@@ -321,10 +412,63 @@ class PlanningItemComponent extends React.Component<IProps, IState> {
...
                 )}
+                {this.state.showCoverageModal && (
+                    <div onClick={(e) => e.stopPropagation()}> {/* avoid opening preview on click in the modal */}
+                        <CoverageAddAdvancedModal
+                            onCancel={this.closeCoverageModal}
+                            contentTypes={contentTypes}
+                            newsCoverageStatus={this.props.newsCoverageStatus}
+                            field="coverages"
+                            value={get(item, 'coverages', [])}
+                            onSave={(field, value) => {
```
thecalcc: could you move this on save handling to a function?

client/components/Coverages/CoverageAddButton/AddCoveragesWrapper.tsx
```diff
@@ -146,6 +146,11 @@ class AddCoveragesWrapperComponent extends React.Component<IProps, IState> {
         this.setState({advanced: false});
     }
 
+    onAdvancedSave = (field: string, value: Array<DeepPartial<IPlanningCoverageItem>>) => {
+        this.props.onChange(field, value);
```
thecalcc, line 150: this `onAdvancedSave` is passed to a button later down the component, so I wonder would it fail given that the button onClick is passing event object probably to this one?

petrjasek: it's not used as an onclick handler directly, it's called from the handler with right data

thecalcc (approved):
CI needs a fix, IIRC you can just reinstall node modules so package lock updates, then you'll probably get a build error and fix should be - https://github.com/superdesk/superdesk-planning/blob/5d77e07f196d8852ba88601ca3f3676db61446bf/webpack.config.js#L4

petrjasek:
> CI needs a fix, IIRC you can just reinstall node modules so package lock updates, then you'll probably get a build error and fix should be -
> 
> https://github.com/superdesk/superdesk-planning/blob/5d77e07f196d8852ba88601ca3f3676db61446bf/webpack.config.js#L4

won't fix it here as it's already failing on base branch

PR #2664, opened 2026-01-08
Assignment item configurable avatar

thecalcc:
### Purpose
Make avatar in assignment list item configurable

Resolves: #[STT-1508]

[STT-1508]: https://sofab.atlassian.net/browse/STT-1508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/planning-extension/src/globals.d.ts
+++ b/client/planning-extension/src/globals.d.ts
@@ -44,7 +44,7 @@ declare module 'superdesk-api' {
             // Controls whether planning should have date only
             all_day?: boolean;
 
-            assignmentItemAvatarDisplayMode?: 'tooltip' | 'inline' | 'none';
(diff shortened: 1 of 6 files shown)


client/planning-extension/src/globals.d.ts
```diff
@@ -44,6 +44,8 @@ declare module 'superdesk-api' {
             // Controls whether planning should have date only
             all_day?: boolean;
 
+            assignmentItemAvatarDisplayMode?: 'tooltip' | 'inline' | 'none';
```
petrjasek: would be good to add some info what those do

petrjasek (approved):

PR #2675, opened 2026-01-16
Tweak language field coverage advanced modal

thecalcc:
### Purpose
- Make available languages in dropdown either ones configured in planning profile, if it's multilingual or if not, all languages are available
- When switching desk, if desk language matches any of the available languages, change the language
- Set user to `null` on desk change, making it consistent with how editor works

Resolves: #[SDBELGA-1011]

[SDBELGA-1011]: https://sofab.atlassian.net/browse/SDBELGA-1011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Coverages/CoverageAddAdvancedModal.tsx
+++ b/client/components/Coverages/CoverageAddAdvancedModal.tsx
@@ -8,6 +8,7 @@ import {IDesk, IUser} from 'superdesk-api';
 import {gettext, planningUtils, getUsersForDesk, getDesksForUser} from '../../utils';
 import {getUserInterfaceLanguageFromCV} from '../../utils/users';
 import {getVocabularyItemFieldTranslated} from '../../utils/vocabularies';
+import {planningApi} from '../../superdeskApi';
 
 import * as selectors from '../../selectors';
 import * as actions from '../../actions';
@@ -15,8 +16,10 @@ import * as actions from '../../actions';
 import {Button, Checkbox, Modal, Spacer, Tooltip} from 'superdesk-ui-framework/react';
 import {CoverageEditableFields} from './CoverageFieldsRow';
 
+import {IVocabularyItem} from 'superdesk-api';
+
 type IReduxStateProps = {
-    languages: Array<any>;
+    allLanguages: Array<{value: IVocabularyItem}>;
 };
 
 interface IReduxDispatchProps {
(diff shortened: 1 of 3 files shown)


client/components/Coverages/CoverageAddAdvancedModal.tsx
```diff
@@ -76,6 +79,35 @@ class CoverageAddAdvancedModalComponent extends React.Component<IProps, IState>
...
 
+    getDefaultLanguage() {
+        const {multilingual} = planningApi.contentProfiles;
+        const profile = planningApi.contentProfiles.get('planning');
+
+        if (!multilingual.isEnabled(profile)) {
+            return null;
+        }
+
+        return planningApi.contentProfiles.getDefaultLanguage(profile) ?? null;
```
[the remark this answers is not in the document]
thecalcc: basically software wont function then

thecalcc:
for context: failing CI is already failing on base branch, it's fixed in later versions

client/components/Coverages/CoverageAddAdvancedModal.tsx
```diff
@@ -73,9 +81,51 @@ class CoverageAddAdvancedModalComponent extends React.Component<IProps, IState>
...
+    }
+
+    getFilteredLanguages(allLanguages: Array<{value: IVocabularyItem}>) {
+        const {multilingual} = planningApi.contentProfiles;
+
+        const planningProfile = planningApi.contentProfiles.get('planning');
+        const isMultilingual = multilingual.isEnabled(planningProfile);
+
+        // If `multilingual` is not enabled, return all languages
+        // If `multilingual` is enabled, filter to only configured languages
```
eos87: nit: the code is already pretty self-explanatory here so probably not needed the two comment lines
```suggestion
        // If `multilingual` is enabled, filter to only configured languages
```

client/components/Coverages/CoverageAddAdvancedModal.tsx
```diff
@@ -73,9 +81,51 @@ class CoverageAddAdvancedModalComponent extends React.Component<IProps, IState>
...
+            const nextFilteredLanguages = this.getFilteredLanguages(this.props.allLanguages);
+
+            if (this.state.filteredLanguages !== nextFilteredLanguages) {
+                // eslint-disable-next-line react/no-did-update-set-state
+                this.setState({
+                    filteredLanguages: nextFilteredLanguages,
+                });
+            }
+        }
+    }
```
eos87: I wonder why do we need `filteredLanguages` in the state in the first place? Maybe I'm missing something, but I think, since the filteredLanguages is calculated from allLanguages, we may not need it in the state at all. Couldn't we just have a, let's say, `getFilteredLanguagesForRender` function that would filter them and just pass them into CoverageEditableFields like

```tsx
<CoverageEditableFields
    languages={this.getFilteredLanguages(this.props.allLanguages)}
    // ... other props
/>
```

eos87 (changes requested):

client/components/Coverages/CoverageAddAdvancedModal.tsx
thecalcc: attempt was to do some kind of caching, but I'll remove it

eos87 (approved):

thecalcc:
@eos87 do you have permission to merge this one without waiting for pytest? given I don't have changes to the backend i think it's ok. I've been waiting for it to start for 1h+

eos87:
> @eos87 do you have permission to merge this one without waiting for pytest? given I don't have changes to the backend i think it's ok. I've been waiting for it to start for 1h+

It won't start because the python 3.12 config is not available in the base branch 2.11, so it was just wasting your time.

PR #2680, opened 2026-01-20
sync coverage data when updating assignments

petrjasek:
STT-1561

### Front-end checklist

Resolves: #[issue-number]


server/planning/data_updates/00036_20260120-120000_sync_assignment_coverages.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class DataUpdate(BaseDataUpdate):
+    resource = "planning"
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        assignments_collection = mongodb_database["assignments"]
+        planning_service = get_resource_service("planning")
+
+        cursor = mongodb_collection.find(
```
MarkLark86: Do these fields have an index in Mongo? Wondering if we should use elastic instead

server/planning/data_updates/00036_20260120-120000_sync_assignment_coverages.py
```diff
@@ -0,0 +1,106 @@
...
+            {
+                "$or": [
+                    {"coverages.assigned_to.assignment_id": {"$exists": True}},
+                    {"coverages.scheduled_updates.assigned_to.assignment_id": {"$exists": True}},
+                ]
+            }
+        )
+
+        for planning in cursor:
```
MarkLark86: Depending on the number of items, could the cursor timeout while we're processing items?
Maybe we should use batched processing instead, get_all_batch I believe it's called

server/planning/data_updates/00036_20260120-120000_sync_assignment_coverages.py
```diff
@@ -0,0 +1,106 @@
...
+
+from superdesk.commands.data_updates import BaseDataUpdate
+from superdesk import get_resource_service
+
+from planning.common import copy_assignment_details_to_coverage
+
+logger = logging.getLogger(__name__)
+
+
+class DataUpdate(BaseDataUpdate):
```
MarkLark86: As this data migration has the potential to take a long time, would this be better off as a command so we can control when it get's run?

server/planning/assignments/assignments.py
```diff
@@ -90,6 +91,72 @@
...
     """Service class for the Assignments model."""
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._skip_planning_sync: bool = False
+
+    async def _update_planning_coverages_from_assignment(
+        self, assignment: Dict[str, Any], skip_planning_sync: bool = False
+    ) -> None:
+        if skip_planning_sync or self._skip_planning_sync:
```
MarkLark86, line 101: Down the track, we need to improve the logic handling between Planning & Assignment synchronisation. This has caused a few bugs in the past, and adding this function in adds to the complexity.

For now it will be fine

server/planning/assignments/assignments.py
```diff
@@ -90,6 +91,72 @@
...
+            coverages = current_planning.get("coverages") or []
+            updated = False
+            for coverage in coverages:
+                if str(coverage.get("coverage_id")) == str(coverage_id) and not is_scheduled_update:
+                    coverage.setdefault("assigned_to", {})
+                    copy_assignment_details_to_coverage(assignment, coverage)
+                    updated = True
+
+                for scheduled_update in coverage.get("scheduled_updates") or []:
+                    if is_scheduled_update and str(scheduled_update.get("scheduled_update_id")) == str(
```
MarkLark86: Could probably have this `is_scheduled_update` check before running the loop, otherwise it's needlessly iterating scheduled updates.

server/planning/assignments/assignments.py
```diff
@@ -90,6 +91,72 @@
...
+                        assignment.get("scheduled_update_id")
+                    ):
+                        scheduled_update.setdefault("assigned_to", {})
+                        copy_assignment_details_to_coverage(assignment, scheduled_update)
+                        updated = True
+
+            if not updated:
+                return False
+
+            result = await planning_service.backend.update_async(
```
MarkLark86, line 134: What is the reason to use the backend directly instead of using `system_update_async`?

server/planning/planning/planning.py
```diff
@@ -142,15 +142,14 @@ def generate_related_assignments(self, docs):
             sync_assignment_details_to_coverages(doc)
 
     async def on_fetched_async(self, docs):
-        self.generate_related_assignments(docs.get(ITEMS))
+        return
```
MarkLark86: We could probably remove these two `on_fetched` functions as we aren't doing anything with them anymore

MarkLark86 (commented):
As this modifies the behaviour of syncing data, it would be good to have some tests for the changes. i.e. patch Assignment, test Planning.

Other than that, it looks good with a few minor questions/suggestions

server/planning/assignments/assignments.py
petrjasek: that avoids syncing data from planning to assignments

server/planning/data_updates/00036_20260120-120000_sync_assignment_coverages.py
petrjasek: yep looking at CP it might take a while

server/planning/data_updates/00036_20260120-120000_sync_assignment_coverages.py
petrjasek: would get probably issue with pagination limit, using the mongo batch helper now

server/planning/commands/sync_assignment_coverages.py
```diff
@@ -0,0 +1,177 @@
...
+        planning_service = get_resource_service("planning")
+        query = {
+            "$or": [
+                {"coverages.assigned_to.assignment_id": {"$exists": True}},
+                {"coverages.scheduled_updates.assigned_to.assignment_id": {"$exists": True}},
+            ]
+        }
+        page_size = 500
+
+        for planning in planning_service.get_all_batch(size=page_size, lookup=query):
```
MarkLark86, line 88: Looks like there might be an issue with the [get_all_batch](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/services.py#L162) function's handling of the lookup. It works on first iteration, but afterwards wipes the lookup and just uses `_id`.

server/planning/commands/sync_assignment_coverages.py
```diff
@@ -0,0 +1,177 @@
...
+            yield planning
+
+    def collect_assignment_ids(self, coverages: list[dict[str, Any]]) -> list[Any]:
+        assignment_ids: list[Any] = []
```
MarkLark86: Ideally we should avoid using `Any` as it doesn't provide type hinting/checking of the values.
In this case it's probably `list[str | ObjectId]`

server/planning/commands/sync_assignment_coverages.py
```diff
@@ -0,0 +1,177 @@
...
+    def find_assignment_by_id(self, assignment_id: Any) -> dict[str, Any] | None:
+        assignments_service = get_resource_service("assignments")
+        query_ids = self.build_assignment_query_ids([assignment_id])
+        cursor = assignments_service.get_from_mongo(req=None, lookup={"_id": {"$in": list(query_ids)}})
+        for doc in cursor:
+            return doc
+        return None
+
+    def build_assignment_query_ids(self, assignment_ids: list[Any]) -> set[Any]:
+        query_ids: set[Any] = set()
```
MarkLark86: Again here, it would be good to define the data type instead of using `Any`

server/planning/commands/sync_assignment_coverages.py
```diff
@@ -0,0 +1,177 @@
...
+        for assignment_id in assignment_ids:
+            if assignment_id is None:
+                continue
+            query_ids.add(assignment_id)
```
MarkLark86: Looks like the `assignment_id` is potentially being added 3 times to `query_ids`. What does this achieve?

MarkLark86 (commented):
Looks good, just some minor type improvements and potential bug in core `get_all_batch` function

server/planning/commands/sync_assignment_coverages.py
petrjasek: fixing it in https://github.com/superdesk/superdesk-core/pull/3138

server/planning/commands/sync_assignment_coverages.py
petrjasek: avoided that, that wasn't needed at the end

MarkLark86 (approved):

PR #2687, opened 2026-01-22
Fix planning filtering for all day items

thecalcc:
### Purpose
Fix planning items disappearing when there's no coverages for them

### What has changed
When we have allDay true which for planning items is always, check if inputDate is only after the end, and not if it's the same. 

Resolves: #[STT-1556]

[STT-1556]: https://sofab.atlassian.net/browse/STT-1556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


petrjasek (approved):

petrjasek:
next release will be based on 3.3, so you can make it on develop now

thecalcc:
in favour of https://github.com/superdesk/superdesk-planning/pull/2689

PR #2703, opened 2026-01-28
add planning_source param to prod_api events

petrjasek:
SDBELGA-1027

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/server/planning/prod_api/events/service.py
+++ b/server/planning/prod_api/events/service.py
@@ -8,7 +8,12 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
+import json
+from typing import Dict, Iterable, List
+
 from eve.utils import config
+from superdesk.errors import SuperdeskApiError
+from werkzeug.datastructures import MultiDict
 
 from superdesk import get_resource_service
 from prod_api.service import ProdApiService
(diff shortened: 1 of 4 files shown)


server/planning/prod_api/events/service.py
```diff
@@ -24,6 +29,73 @@
...
+            raise SuperdeskApiError.badRequestError("planning_source must be an object")
+
+        return query
+
+    def _get_event_ids_from_planning(self, planning_query: Dict) -> List[str]:
+        # Query planning documents using the planning service
+        planning_service = get_resource_service("planning")
+
+        # Search for planning documents matching the query
+        results = planning_service.search(planning_query)
```
MarkLark86, line 89: It might be a good idea to use projection here so we're only getting `event_item` (or `related_events` from 3.0+) from es

MarkLark86: Also, looks like the `planning_source` param allows the entire es query.
This means the consumer could pass in `"size": 0` which would break this functionality, or unnecessarily provide `"aggs": {` which will never be used.

Would it be a good idea to treat the provided `planning_source` as the body of the `"query"` es param?
So
```json
{"query": {"query_string": {"query": "EVENT", "default_field": "slugline"}}}
```
would become:
```json
{"query_string": {"query": "EVENT", "default_field": "slugline"}}
```

MarkLark86 (commented):
Looks good. Just a couple minor improvement suggestions

server/planning/prod_api/events/service.py
petrjasek: I wanted to keep it consistent with the query you would use on `/planning` api, so you can test it there and then just use it. would be good to do some checks on those source params in both cases, we might do it separately

petrjasek:
will do the improvements later, they need to test it now

PR #2705, opened 2026-01-29
[STT-1552] fix: Use CoverageProfile ID for planning:default_coverage_desks preference

MarkLark86:
### Purpose
When creating a new Coverage, the system is meant to remember what Desk you lasted used for that Coverage type and assign that automatically for you. This no longer works.

### What has changed
* Use CoverageProfile ID when storing to the User's `planning:default_coverage_desks` preference.
* Use the PlanningAPI to set the default coverage values (which retrieves all data from the redux store required).
* Remove the `AddNewCoverages` component, as it is not used anywhere.

### Steps to test
1. Create a new Planning item
2. Add a Text coverage, and assign a Desk to it
3. Create another Planning item
4. Add a Text coverage

Notice how the Desk selected in step 2 is automatically applied in step 4 as well.

### Front-end checklist

Resolves: #[issue-number]


thecalcc (approved):
looks great!

thecalcc:
just one thing to watch out for as you mentioned in scrum is which branch we push this to, I think this should go to 3.3, but pls confirm with Petr

PR #2706, opened 2026-01-30
[STT-1562] fix: Dont copy Planning metadata to Coverage if field not enabled

MarkLark86:
### Purpose
When a Coverage is added to a Planning item, the back-end copied certain metadata from the Planning item to the Coverage (if not already defined by the Coverage).

Previously the metadata was copied even if that particular field was not enabled on the Coverage. This caused issues when creating an Assignment and then Content from it, if the Planning's Category field was changed, the Coverage's Content field was still used (but never shown to the end user).

### What has changed
* Add `get_coverage_schema`, `is_field_custom_vocabulary` and `get_custom_vocabulary_fields_from_profile` to ContentProfile utilities
* Use CoveargeProfile to determine what fields to copy from the Planning item to the Coverage (including custom vocabularies)
* Set `superdesk-core` version in dev requirements to `release/3.3`

Resolves: [STT-1562](https://sofab.atlassian.net/browse/STT-1562)

[STT-1562]: https://sofab.atlassian.net/browse/STT-1562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.travis-docker-compose.yml
+++ b/.travis-docker-compose.yml
@@ -6,17 +6,18 @@ services:
             - "6379:6379"
 
     mongo:
-        image: mongo:3.6
+        image: mongo:6
         ports:
             - "27017:27017"
         tmpfs:
             - /data/db
 
     elastic:
-        image: elasticsearch:7.6.2
+        image: elasticsearch:7.17.29
         ports:
             - "9200:9200"
         environment:
             - discovery.type=single-node
+            - xpack.security.enabled=false
         tmpfs:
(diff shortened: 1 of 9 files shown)


eos87 (approved):
Looks good. I think we should rename (or delete if not used) that `.travis-docker-compose.yml`, I don't think we use it any longer for CI.

PR #2707, opened 2026-02-01
[SDBELGA-1028] Add content privilege checks for assignment locks

eos87:
### Purpose
Users with "Assignments" and "Create Content" privileges but without "Planning - Planning item management" privilege were unable to use the "Start Working" action on assignments. When clicking the action, they received an "User does not have sufficient permissions." error, even though the action was correctly visible in the UI.

### What has changed
- Introduces `can_work_on_content` to `AssignmentsService` to check archive privilege for content operations.
- Updates `LockService` to use this method for assignment lock/unlock actions related to content, ensuring correct permission checks for `start_working` and `content_edit` actions.

> [!IMPORTANT]  
> The tests broken in this PR were already broken in branch `release/2.11`. I think those have been addressed on other branches so not fixing them here. 

Resolves: [SDBELGA-1028]

[SDBELGA-1028]: https://sofab.atlassian.net/browse/SDBELGA-1028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -1196,20 +1196,20 @@ class AssignmentsService(superdesk.Service):
             lock_service = get_component(LockService)
             lock_service.unlock(assignment, user_id, get_auth()["_id"], "assignments")
 
-    def can_work_on_content(self, item, user_id):
+    def can_work_on_content(self, _item, _user_id):
         """Check if user can work on assignment content (lock/unlock for content operations).
 
         This requires only the archive privilege, as working on content means creating/editing archive items.
         Used for: start_working action, content_edit action, and unlocking assignments.
         """
         if not current_user_has_privilege("archive"):
-            return False, "User does not have sufficient permissions."
+            return False, lazy_gettext("User does not have sufficient permissions.")
         return True, ""
 
     def can_edit(self, item, user_id):
         # Check privileges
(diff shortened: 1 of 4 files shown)


server/planning/assignments/assignments.py
```diff
@@ -1196,6 +1196,16 @@ def sync_assignment_unlock(self, item, user_id):
...
             lock_service.unlock(assignment, user_id, get_auth()["_id"], "assignments")
 
+    def can_work_on_content(self, item, user_id):
+        """Check if user can work on assignment content (lock/unlock for content operations).
+
+        This requires only the archive privilege, as working on content means creating/editing archive items.
+        Used for: start_working action, content_edit action, and unlocking assignments.
+        """
+        if not current_user_has_privilege("archive"):
+            return False, "User does not have sufficient permissions."
```
petrjasek: not sure if we display those later, but we should probably use gettext for those messages

petrjasek (approved):

server/planning/assignments/assignments.py
eos87: Good point. Yeah, we do. I'll adjust it.

PR #2715, opened 2026-02-05
[SDCP-1073] fix all-day end date timezone

BrianMwangi21:
### Purpose
This PR fixes the off‑by‑one day issue when selecting end dates for all‑day multi‑day events in negative timezones like Toronto

### What has changed
- Ensure the end date field is treated as all‑day when either `dates.all_day` or `dates.no_end_time` is true.
- Normalize all‑day date values to the event timezone before passing them into the date picker, so the calendar day remains stable across timezones.

### Steps to test
1. Set your local timezone to a negative offset like Toronto 
2. Create a multi‑day all‑day event
3. Select a start date and an end date several days apart
4. Verify the picker shows the same day you click
5. Repeat in positive timezones e.g Kenya or even Sydney to ensure nothing breaks

Resolves : [SDCP-1073](https://sofab.atlassian.net/browse/SDCP-1073)

[SDCP-1073]: https://sofab.atlassian.net/browse/SDCP-1073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/UI/Form/DateTimeInput/index.tsx
+++ b/client/components/UI/Form/DateTimeInput/index.tsx
@@ -77,6 +77,14 @@ export const DateTimeInput = ({
     ...props
 }: IProps) => {
     let timeValue = timeField ? get(diff, timeField, null) : value;
+    let dateValue = value;
+
+    if (props.allDay && value != null) {
+        const tz = remoteTimeZone || moment.tz.guess();
+        const momentValue = moment.isMoment(value) ? value : moment(value);
+
+        dateValue = moment.tz(momentValue.format('YYYY-MM-DD'), tz);
+    }
 
     if (props.allDay) {
         timeValue = null;
@@ -98,7 +106,7 @@ export const DateTimeInput = ({
                 row={false}
                 component={DateInput}
                 field={`${field}.date`}
-                value={value}
+                value={dateValue as moment.Moment}
                 item={item}
                 diff={diff}
                 readOnly={readOnly}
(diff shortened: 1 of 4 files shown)


client/components/UI/Form/DateTimeInput/index.tsx
```diff
@@ -77,6 +77,14 @@ export const DateTimeInput = ({
     ...props
 }: IProps) => {
     let timeValue = timeField ? get(diff, timeField, null) : value;
+    let dateValue = value;
+
+    if (props.allDay && value != null) {
+        const tz = remoteTimeZone || moment.tz.guess();
```
petrjasek: I'm not sure exactly what it does, when event is all day there is no timezone in general, so it's using browser timezone and converting the value to it? that's the fix?

BrianMwangi21: Yes. So it normalizes the display value to the event timezone or browser TZ if none, so the date picker doesn’t shift the day. But it doesn't change how all‑day dates are saved. That part still goes through `localDateToUtc`.

petrjasek: it might be better then to use:
```
moment(moment.utc(value).format('YYYY-MM-DD'));
```
so it creates local date with the event date unchanged without dealing with event timezone (in case it's all_day that info is irrelevant anyhow). not sure if there is a helper, might be good to add one

client/utils/time.ts
```diff
@@ -174,6 +174,10 @@ function getDateAsString(value: string | Date | moment.Moment): string {
     }
 }
 
+function normalizeAllDayDate(value: moment.MomentInput): moment.Moment {
```
petrjasek: not sure what normalize mean here, maybe `allDayDateToLocalDate` would be better

petrjasek (approved):

PR #2736, opened 2026-02-16
[SDESK-7305] - Week start day setting not respected globally

BrianMwangi21:
### Purpose
Align Planning week-start behavior with client-core so it follows `startingDay` when present.

### What has changed
Updated Planning config normalization to mirror `startingDay` into `start_of_week`, and backfill `startingDay` when it’s missing.

### Steps to test
1. Set `startingDay: 1` in `superdesk.config.js` and set backend `START_OF_WEEK=0`.
2. Rebuild and start the client.
3. Open Planning date pickers (events/planning) and confirm week starts on Monday.

Note: This PR goes hand in hand with this change in [superdesk-client-core](https://github.com/superdesk/superdesk-client-core/pull/5073)

Resolves: [SDESK-7305](https://sofab.atlassian.net/browse/SDESK-7305)

[SDESK-7305]: https://sofab.atlassian.net/browse/SDESK-7305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/config.ts
+++ b/client/config.ts
@@ -74,13 +74,11 @@ if (appConfig.start_of_week == null) {
 }
 
 // Align planning week start with core config
-if (appConfig.startingDay != null) {
-    appConfig.start_of_week = typeof appConfig.startingDay === 'string'
-        ? parseInt(appConfig.startingDay, 10)
-        : appConfig.startingDay;
-} else {
-    appConfig.startingDay = appConfig.start_of_week;
-}
+const startingDay = appConfig.startingDay;
+const normalizedStartingDay = typeof startingDay === 'string' ? parseInt(startingDay, 10) : startingDay;
+
+appConfig.start_of_week = normalizedStartingDay ?? appConfig.start_of_week;
+appConfig.startingDay = appConfig.start_of_week;
 
 if (appConfig.planning == null) {
     appConfig.planning = {};
--- a/client/config.ts
+++ b/client/config.ts
@@ -73,6 +73,15 @@ if (appConfig.start_of_week == null) {
     appConfig.start_of_week = parseInt(appConfig.start_of_week, 10);
 }
 
+// Align planning week start with core config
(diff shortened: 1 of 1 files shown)


client/config.ts
```diff
@@ -73,6 +73,13 @@ if (appConfig.start_of_week == null) {
     appConfig.start_of_week = parseInt(appConfig.start_of_week, 10);
 }
 
+// Align planning week start with core config
+const startingDay = appConfig.startingDay;
```
thecalcc: aren't we normalising it already in client-core? If so I'd expect it to be already normalised here no?

client/config.ts
```diff
@@ -73,6 +73,9 @@ if (appConfig.start_of_week == null) {
     appConfig.start_of_week = parseInt(appConfig.start_of_week, 10);
 }
 
+// Fallback when start_of_week is not provided.
+appConfig.start_of_week = appConfig.start_of_week ?? appConfig.startingDay ?? 0;
```
thecalcc, line 77: My point with the previous comment was that if you need to change something you should do it in `superdesk-client-core` repo since that's where normalisation is handled. Otherwise we're spilling the logic across all repos and that's not great

If you have to adjust something here in planning, that should rather be the usage of the config. So you can do this fallback functionality if you think it's suitable in client-core then in here (planning) you can just replace usages of e.g. `appConfig.startingDay` with the other config option - `appConfig.start_of_week`

That way the logic related to this stays in the same place, and then you can just consume the config however it suits in the other connected projects.

BrianMwangi21: Ahhhh, got it. I had confused it with just removing the normalization bit, but now I get you. Makes sense.

Then I'll close this PR.

PR #2779, opened 2026-03-24
[SDESK-7865] fix: Hourly scheduled filter has invalid hours value

MarkLark86:
### Purpose
It was possible for a Event/Planning Filter to have an hourly schedule with `hours` value set, which leads to the item running once a day and not hourly.

### What has changed
* Make sure if an hourly schedule is created/updated, set the `hours` value to empty
* Create a migration script to fix existing filters

Resolves: [SDESK-7865](https://sofab.atlassian.net/browse/SDESK-7865)

[SDESK-7865]: https://sofab.atlassian.net/browse/SDESK-7865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/data_updates/00038_20260324-131530_events_planning_filters.py
+++ b/server/planning/data_updates/00038_20260324-131530_events_planning_filters.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8; -*-
+# This file is part of Superdesk.
+# For the full copyright and license information, please see the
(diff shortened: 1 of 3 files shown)


eos87 (approved):
Looks good

PR #2817, opened 2026-04-28
[STT-1582] fix: Desk re-assignment allowed when multiple_content enabled

MarkLark86:
### Purpose
When attempting to change the Desk of an Assignment that is in progress, the backend raises an error. We should allow this for Assignments with the `multiple_content` enabled

### What has changed
* Check Assignment multiple_content attribute when validating desk change

Resolves: [STT-1582](https://sofab.atlassian.net/browse/STT-1582)

[STT-1582]: https://sofab.atlassian.net/browse/STT-1582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/assignments/assignments.py
+++ b/server/planning/assignments/assignments.py
@@ -248,14 +248,16 @@ class AssignmentsService(AsyncBaseService):
             if not assigned_to.get("desk"):
                 raise SuperdeskApiError.badRequestError(message="Assignment should have a desk.")
 
-        if original.get("assigned_to", {}).get("desk") != assigned_to.get("desk"):
-            if original.get("assigned_to", {}).get("state") in [
-                ASSIGNMENT_WORKFLOW_STATE.IN_PROGRESS,
-                ASSIGNMENT_WORKFLOW_STATE.SUBMITTED,
-            ]:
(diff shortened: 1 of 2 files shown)


server/planning/assignments/assignments.py
```diff
@@ -248,14 +248,16 @@ async def validate_assignment(self, updates: dict, original: dict) -> None:
...
-                )
+        multi_content_disabled = not assignment_allows_multiple_content_linked(original)
+        desk_changed = original.get("assigned_to", {}).get("desk") != assigned_to.get("desk")
+        in_progress_or_submitted = original.get("assigned_to", {}).get("state") in [
+            ASSIGNMENT_WORKFLOW_STATE.IN_PROGRESS,
+            ASSIGNMENT_WORKFLOW_STATE.SUBMITTED,
+        ]
+        if multi_content_disabled and desk_changed and in_progress_or_submitted:
+            raise SuperdeskApiError.forbiddenError(
+                message="Assignment linked to content. Desk reassignment not allowed."
```
petrjasek: we should use gettext for the error message

PR #2824, opened 2026-05-04
fix assignment info syncing in coverage editor

petrjasek:
saving changes to coverage triggers assignment:updated signal which would call onChange with dirty flag so save button stays active.

STT-1575

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/components/fields/editor/Coverages.tsx
+++ b/client/components/fields/editor/Coverages.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {get} from 'lodash';
+import {cloneDeep, get, isEqual} from 'lodash';
 
 import {superdeskApi, planningApi} from '../../../superdeskApi';
 import {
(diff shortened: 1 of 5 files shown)


client/components/fields/editor/Coverages.tsx
```diff
@@ -88,12 +102,26 @@ export class EditorFieldCoverages extends React.PureComponent<IPropsEditorFieldC
...
+                return;
+            }
+
             const fieldName = scheduledUpdateIndex === -1 ?
                 `coverages[${coverageIndex}].assigned_to` :
                 `coverages[${coverageIndex}].scheduled_updates[${scheduledUpdateIndex}].assigned_to`;
 
-            this.props.onChange(fieldName, coverageToUpdate.assigned_to);
+            // Keep assignment sync visible in the form without triggering Save button re-activation.
+            // saveAutosave=true ensures the autosave stays in sync so re-opening the editor
```
MarkLark86: I don't think `saveAutosave` should not be required, as we're syncing autosave data on the backend (especially for scenarios where the item isn't currently open in the editor, but still locked for editing - minimised).

petrjasek: saveAutosave is enabled, but not updateDirtyFlag

MarkLark86: If this is enabled, it calls [saveAutosave](https://github.com/superdesk/superdesk-planning/blob/develop/client/components/Main/ItemEditor/AutoSave.ts#L98), which then sends a patch request to the backend.
Which is not required in this case

MarkLark86 (approved):

PR #2834, opened 2026-05-13
[STT-1670] - Issues during file upload

BrianMwangi21:
### Purpose
This PR tries to improve attachment uploads in event and planning editors, especially drag-and-drop behavior and upload failure handling. Although not fully reproduced locally, the changes help make it more robust

### What has changed
- Hardened drag/drop handling in the shared file input.
- Replaced modal drag-target detection with closest().
- Added shared upload retry and timeout handling for event and planning file uploads.

### Steps to test
1. Open an event or planning item and use Attached Files.
2. Upload a file via drag-and-drop and via browse.
3. Confirm the file attaches successfully and the UI does not hang.
4. Throttle or interrupt the network and confirm the upload fails cleanly and retries once before showing an error.

Resolves: #[STT-1670](https://sofab.atlassian.net/browse/STT-1670)

[STT-1670]: https://sofab.atlassian.net/browse/STT-1670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -524,7 +524,7 @@ const fetchEventHistory = (eventId) => (
     ));
 
 const uploadFiles = (event) => (
-    (dispatch) => {
+    (dispatch, getState, {upload}) => {
         const clonedEvent = cloneDeep(event);
 
         // If no files, do nothing
@@ -543,6 +543,7 @@ const uploadFiles = (event) => (
 
         return Promise.all(filesToUpload.map((file) => (
             uploadFileWithRetry<IFile>(
+                upload,
                 '/events_files/',
                 file,
                 {timeoutMs: 90000, retries: 2},
--- a/client/actions/events/api.ts
+++ b/client/actions/events/api.ts
@@ -5,6 +5,7 @@ import {
     ISearchSpikeState,
     IEventSearchParams,
     IEventItem,
+    IFile,
     IPlanningItem,
     IEventTemplate,
     IEventUpdateMethod,
@@ -27,6 +28,7 @@ import {
     isPublishedItemId,
     gettext,
 } from '../../utils';
+import {uploadFileWithRetry} from '../../utils/upload';
 
 import planningApis from '../planning/api';
 import eventsUi from './ui';
@@ -522,7 +524,7 @@ const fetchEventHistory = (eventId) => (
     ));
 
 const uploadFiles = (event) => (
-    (dispatch, getState, {upload}) => {
+    (dispatch) => {
         const clonedEvent = cloneDeep(event);
 
         // If no files, do nothing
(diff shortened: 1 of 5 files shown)


client/utils/upload.ts
```diff
@@ -0,0 +1,112 @@
...
+
+function createUploadError(message: string, extra: Record<string, unknown> = {}) {
+    return Object.assign(new Error(message), extra);
+}
+
+function isRetryableUploadError(error: any) {
+    return error?.timeout === true || error?.retryable === true || error?.status >= 500;
+}
+
+function toUploadError(error: any) {
```
MarkLark86: What is the reason for this?

MarkLark86 (approved):
Looks okay to me.
Was `dropEffect` mostly the cause?
Also, would be good to get Helmy's review on the new upload service (should it be in client-core's API)?

client/utils/upload.ts
```diff
@@ -0,0 +1,112 @@
...
+    const timeoutPromise = new Promise<never>((_, reject) => {
+        setTimeout(() => {
+            reject(createUploadError('Upload timed out.', {
+                timeout: true,
+                retryable: true,
+            }));
+        }, options.timeoutMs);
+    });
+
+    const response = await Promise.race([uploadPromise, timeoutPromise]);
```
eos87: This triggers a concern in my mind. I think there is a chance of duplicate uploads here. We're racing both promises here but, if the `timeoutPromise` rejects first the `uploadPromise` is not necessarily cancelled or aborted, it only stops waiting for it, meaning the original `upload.start` may still complete in the background. Or am I missing something? do we have the same timeout in the backend?

client/actions/events/api.ts
```diff
@@ -540,16 +542,15 @@ const uploadFiles = (event) => (
...
-                url: appConfig.server.url + '/events_files/',
-                headers: {'Content-Type': 'multipart/form-data'},
-                data: {media: [file]},
-                arrayKey: '',
-            })
+            uploadFileWithRetry<IFile>(
+                upload,
+                '/events_files/',
+                file,
+                {timeoutMs: 90000, retries: 2},
```
eos87: given the default timeout is `90000` in `upload.ts` (`DEFAULT_TIMEOUT_MS`) we can remove it from here.

client/utils/upload.ts
BrianMwangi21: I have removed timeout from the retry path so we don't duplicate uploads. 

There isn't a matching backend timeout in this flow, cause this PR is mainly trying to improve the upload path since I wasn't able to reproduce the issue reliably myself locally and on the test instances we have.

eos87: but I still see the timeoutPromise here in the race 
```typescript
const response = await Promise.race([uploadPromise, timeoutPromise]);
```

client/utils/upload.ts
```diff
@@ -0,0 +1,129 @@
...
+        let settled = false;
+
+        const timeoutId = setTimeout(() => {
+            settled = true;
+
+            if (typeof uploadPromise.abort === 'function') {
+                uploadPromise.abort();
+            }
+
+            reject(createUploadError('Upload timed out.'));
```
eos87: I think you need to add `timeout: true` so it will be catch by isRetryableUploadError
```suggestion
            reject(createUploadError('Upload timed out.', {timeout: true}));
```

client/utils/upload.ts
```diff
@@ -0,0 +1,129 @@
...
+        let settled = false;
+
+        const timeoutId = setTimeout(() => {
+            settled = true;
+
+            if (typeof uploadPromise.abort === 'function') {
+                uploadPromise.abort();
+            }
+
+            reject(createUploadError('Upload timed out.', {timeout: true}));
```
eos87: please user the i18n helper for this string.

eos87 (approved):
Looks good. Left just one minor comment.

PR #2869, opened 2026-06-24
Bump python version in CI

eos87:
This pull request updates the project's CI workflows and configuration files to use Python 3.12, removing support for Python 3.10.

### DIFF
--- a/server/planning/planning_notifications.py
+++ b/server/planning/planning_notifications.py
@@ -381,7 +381,7 @@ async def _send_user_email(user_id, contact_id, source, meta_message, data):
         ics = data["assignment"]["planning"]["ics_data"]
         if ics:
             name = get_assginment_name(data.get("assignment"))
-            attachments.append(EmailAttachment(filename=name, content_type="text/calendar", data=ics))
+            attachments.append(EmailAttachment(filename=name, content_type="text/calendar", data=ics.encode("utf-8")))
 
     await send_email(
         subject=data["subject"],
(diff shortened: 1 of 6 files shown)


server/planning/planning_notifications.py
```diff
@@ -381,7 +381,7 @@ async def _send_user_email(user_id, contact_id, source, meta_message, data):
         ics = data["assignment"]["planning"]["ics_data"]
         if ics:
             name = get_assginment_name(data.get("assignment"))
-            attachments.append(EmailAttachment(filename=name, content_type="text/calendar", data=ics))
+            attachments.append(EmailAttachment(filename=name, content_type="text/calendar", data=ics.encode("utf-8")))
```
eos87: @petrjasek this is just to unblock the build, but the underlying issue is a regression in core: `_get_message_from_request` can't handle `str` attachments, even though `EmailAttachment.data` is typed `str | bytes`. I will open another PR in core shortly.

eos87: PR here https://github.com/superdesk/superdesk-core/pull/3240

petrjasek:
we might change it also in https://github.com/superdesk/superdesk-planning/blob/develop/e2e/server/Dockerfile but that doesn't seem used on CI

eos87:
> we might change it also in https://github.com/superdesk/superdesk-planning/blob/develop/e2e/server/Dockerfile but that doesn't seem used on CI

good catch, it is used from the e2e script to start up the services (the AI skill uses this). I will test it locally and update it.

eos87:
@petrjasek I've updated the Dockerfile, please take another look when you get a time.

petrjasek (approved):

PR #2899, opened 2026-07-21
[SDBELGA-1104] fix: Reload list using single Redux action

MarkLark86:
### Purpose
There were many redux actions being dispatched when a Planning item was updated, causing re-renders of React components many times over without a predictable order. 

### What has changed
* Create a new `planningApi.ui.list.reloadListPages` that reloads the current list, regardless of the filter type (Event, Planning or Combined).
* Retrieve all data required first, then dispatch it in a single action, letting the reducers do the work of what needs to be updated
* Use this new list reloading with websocket messages for Planning create & update (can be used for many more scenarios, but leaving it just for these 2 for now)

### Front-end checklist

Resolves: [SDBELGA-1104](https://sofab.atlassian.net/browse/SDBELGA-1104)


petrjasek (approved):
looks good to me

PR #2900, opened 2026-07-21
[SDBELGA-1104] fix: Reload Events in list using single Redux action

MarkLark86:
### Purpose
The previous fix seemed to work for Planning items, but the same bug occurred when creating or updating Events.

### What has changed
* Use the newer `reloadListPages` functionality when reacting to websocket messages for Event create and update
* Show user error message when reloading the list fails
* Add new param to `reloadListPages` that allows to reload only the list if it's in the Event or Planning view, or the combined view
* Fixes as error when there are too many Events with associated Planning items, leading to the `Request Line is too large` error (uses EventID request chunking instead)._

Resolves: [SDBELGA-1104]

Related PR https://github.com/superdesk/superdesk-planning/pull/2899

### DIFF
--- a/client/actions/events/notifications.ts
+++ b/client/actions/events/notifications.ts
@@ -1,7 +1,7 @@
 import {get} from 'lodash';
 
 import {planningApi} from '../../superdeskApi';
-import {IWebsocketMessageData, ITEM_TYPE} from '../../interfaces';
+import {IWebsocketMessageData, ITEM_TYPE, PLANNING_VIEW} from '../../interfaces';
 import * as selectors from '../../selectors';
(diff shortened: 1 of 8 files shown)


eos87 (approved):
Looks good. 

I added  two changes: 

1394e687 Address Copilot review comments

55e1fb31 Throttle reloadListPages to collapse websocket notification bursts

Unlike the old scheduleRefetch path, reloadListPages didn’t combine concurrent reload requests. I added a promise-aware throttle: the first call reloads immediately, calls during a reload or its 1s cooldown share one follow-up, and reloads never overlap.

PR #2903, opened 2026-07-22
[SDBELGA-1104] fix: Use PlanningAPI to reload and load next pages

MarkLark86:
### Purpose
The previous changes worked, to a degree. Other actions that updated an Event or Planning item still used the old code, which exposed the original bug. On top of that, reloading the pages did not work if the last page loaded was not a complete page (partial page due to end of search).

### What has changed
* Create more functions in `ui.list` to update search params and reload page, load next page etc.
* Update websocket notifications to use the newer PlanningAPI list functionality
* Update initial page load to use newer Planning API list functionality (speeds up initial render, as it no longer needs multiple dispatch -> render cycles.
* Update tests to expect PlanningAPI list functionality to be called
* Remove code no longer used (specifically the `loadMore`)

### Screenshots
Before:
image
After:
image
_Notice the Cumulative Layout Shift drastically improves_

### Screencasts
Add Planning / Modify Coverage to Event on last page:
https://github.com/user-attachments/assets/a0a2b574-08ae-4d6f-a67e-2b2431791db3

Reschedule Event In Later Pages
https://github.com/user-attachments/assets/fcb723c6-beda-4e16-b2ab-ecc309075b5d

### Front-end checklist

Resolves: [SDBELGA-1104](https://sofab.atlassian.net/browse/SDBELGA-1104)


eos87 (approved):
Looks good. Great improvements!

eos87 (approved):

PR #2925, opened 2026-08-04
[STT-1738] Add a `card_view` list config for entity cards in previews

eos87:
### Purpose

The cards for linked items inside a preview (the related Event on a Planning item, related Planning items on an Event, both in the Assignment preview) reuse the full list configuration, which is too dense for the narrow preview panel. They also render a content type icon that adds nothing there.

This adds an optional `card_view` configuration for those cards and removes the icon. Without `card_view` configured, nothing changes.

Showing existing content items on coverage cards is struck through in the ticket and deferred pending a design concept.

### What has changed

`card_view` sits alongside `firstLine`, `secondLine` and `compact_view` under `planning_list_item` and `event_list_item`.

One rule worth knowing: a configured `card_view` fully describes the card, so an omitted line renders nothing rather than inheriting the list line. Falling back to the list configuration, then to the defaults, happens only when `card_view` is absent entirely.

`RelatedEventListItem` and `RelatedPlanningListItem` take a `cardView` prop that selects the card configuration and drops the `ItemIcon` column. It is set in the three previews only. The editor bookmarks bar, the editor related-item fields, `ExportAsArticleModal` and the confirmation modals render the same components and are deliberately untouched; extending to them is a follow-up.

Also here: `List.Column` gained an optional `testId` (mirroring `List.Item`) so the e2e spec asserts on a stable hook rather than a styling class, and the README now documents the list item configuration family, which was undocumented.

### Steps to test

Nothing is visible until `card_view` is configured, so add a block to `superdesk.config.js` under `planning` that differs from the list configuration:

```js
event_list_item: {
    firstLine: [{fieldId: 'slugline'}, {fieldId: 'name'}, {fieldId: 'calendars'}],
    secondLine: [{fieldId: 'state'}],
    card_view: {
        firstLine: [{fieldId: 'name'}],
        secondLine: [{fieldId: 'state'}],
    },
},
```

1. Preview a Planning item linked to an Event: the Event card shows only the `card_view` fields (no slugline, no calendars) and no content type icon.
2. Repeat on an Event preview with related Planning items, and on an Assignment preview where both cards appear.
3. Open that Planning item in the editor: the associated Event uses the same component without the card configuration, so it must still show the full list configuration and the icon.
4. Remove `card_view` and confirm the cards look as they do today, except the icon is gone everywhere.

Covered by 16 unit tests and 7 Playwright tests (`e2e/playwright/preview_card_view.spec.ts`), including controls that the same component without `cardView`, and the list rows, are unaffected.

Resolves: [STT-1738](https://sofab.atlassian.net/browse/STT-1738)

[STT-1738]: https://sofab.atlassian.net/browse/STT-1738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ It allows to ingest and manage events, to create planning within agenda, and to 
     * [Assignments](#assignments-config)
     * [Authoring](#authoring-config)
     * [Search Filters](#search-filters-config)
+* [Client Config](#client-config)
+    * [List item config](#list-item-config)
 * [Slack Integration](#slack-integration)
 * [Celery Tasks](#celery-tasks)
     * [Expire Items](#celery-tasks-expire-items)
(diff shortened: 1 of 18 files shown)


MarkLark86 (approved):
Looks good to me

PR #2930, opened 2026-08-05
[SDBELGA-1113] Move Event & Planning service logic into the Unified service

MarkLark86:
### What has changed

Resolves: SDBELGA-1113
Depends on: https://github.com/superdesk/superdesk-core/pull/3282


MarkLark86:
@petrjasek those recurring item issues will be solved with a different ticket (TBC) along with metadata sync (ticket already exists)

petrjasek (approved):
looks good

PR #2940, opened 2026-08-10
[SDBELGA-1128] - Update commands to use the unified resource

BrianMwangi21:
### Purpose
Migrate all `planning:*` management commands off the legacy Eve-backed `events` / `planning` resources and onto the unified resource (`unified_planning`),

### What has changed
Migrated commands (`server/planning/commands/`):
- `planning:flag_expired` 
- `planning:delete_spiked` 
- `planning:export_to_newsroom`
- `planning:purge_expired_locks`
- `planning:replace_deprecated_event_item_attribute` 
- `planning:sync_assignment_coverages`
- `planning:delete_assignments`
- `planning:export_scheduled_filters`

Resolves: [SDBELGA-1128](https://sofab.atlassian.net/browse/SDBELGA-1128)

[SDBELGA-1128]: https://sofab.atlassian.net/browse/SDBELGA-1128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/commands/delete_spiked_items_test.py
```diff
@@ -81,23 +95,25 @@ class BaseDeleteSpikedItemsTest(TestCase):
...
+        await test_utils.post_items("vocabularies", planning_fixtures.cvs.all_cvs())
         await test_utils.post_items("desks", fixtures.desks.all_desks())
+        await test_utils.post_items("stages", fixtures.stages.all_stages())
 
     async def insert(self, item_type, items):
-        service = self.event_service if item_type == "events" else self.planning_service
-        await service.post_async(items)
+        item_type_value = "event" if item_type == "events" else "planning"
+        await self.planning_service.create([{**item, "type": item_type_value} for item in items])
+        # unified resource is not force-refreshed in tests; refresh so the command's search sees the items
```
MarkLark86: Not sure why this is happening, because it's [on by default](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/types/elastic.py#L50), and we aren't assigning a value to it in the U[nifiedResource config](https://github.com/superdesk/superdesk-planning/blob/develop/server/planning/unified/module.py#L47)

MarkLark86 (approved):
Looks good, just 1 comment

server/planning/commands/delete_spiked_items_test.py
BrianMwangi21: Elastic doesn't auto-refresh after insert, so the command's search would find nothing and that's why I added the manual refresh. Behave has auto-refresh on, but pytest doesn't. But I have turned it on in the test config instead and dropped the manual refreshes

MarkLark86: I'm OK with this being merged in it's current state, but this should not be happening.
https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/elastic/async_client.py#L41

We're manually doing a refresh if it is turned on in the Pydantic based async resource (which it is by default).

BrianMwangi21: I am a bit confused, I thought [L41](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/elastic/async_client.py#L41) should've handled it. But the `force_refresh` it checks comes from the `ELASTICSEARCH_FORCE_REFRESH` setting, and [settings.py sets it to `is_testing` here](https://github.com/superdesk/superdesk-planning/blob/feature/unified/server/settings.py#L121-L122). 

Under pytest, SUPERDESK_TESTING isn't set, so is_testing and therefore force_refresh is False  so the refresh at L41 never runs.

Am I reading it wrong, or missing something?

MarkLark86: Ah right, yes it can be set globally via that setting. I didn't realize it's disabled there. That's what's happening.

You can merge this PR, and we'll fix this later

PR #2941, opened 2026-08-11
[SDBELGA-1129] - Update Autocomplete to use the unified resource

BrianMwangi21:
### Purpose
Migrate the planning autocomplete providers off the legacy Eve-backed `events` / `planning` elastic indexes and onto the async unified resource (`unified_planning`).

### What has changed
`server/planning/search/planning_autocomplete.py`:
- Both providers (`get_planning_suggestions`, `get_event_suggestions`) are now async and query the single `unified_planning` index via the async elastic client instead of the sync Eve datalayer.
- Added an `item_type` filter (`{"term": {"type": "event"|"planning"}}`) so event and planning suggestions stay separate on the shared index.
- Added tests 

Resolves: [SDBELGA-1129](https://sofab.atlassian.net/browse/SDBELGA-1129)

[SDBELGA-1129]: https://sofab.atlassian.net/browse/SDBELGA-1129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/planning/search/planning_autocomplete.py
+++ b/server/planning/search/planning_autocomplete.py
@@ -1,7 +1,7 @@
 from typing import Dict, Any
 from datetime import timedelta
 
-from superdesk.core import get_app_config, get_current_async_app
+from superdesk.core import get_app_config
 from superdesk.utc import utcnow
 from apps.archive.autocomplete import (
     SETTING_LIMIT as AUTOCOMPLETE_LIMIT,
@@ -12,8 +12,6 @@ from apps.archive.autocomplete import (
 
 from planning.common import WORKFLOW_STATE, POST_STATE
 from planning.types.unified import PlanningItemType, UnifiedPlanningResource
-
-UNIFIED_RESOURCE = "unified_planning"
 
 
 async def get_planning_suggestions(field: str, language: str) -> Dict[str, int]:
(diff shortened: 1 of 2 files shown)


server/planning/search/planning_autocomplete.py
```diff
@@ -41,22 +44,28 @@ def get_planning_suggestions(field: str, language: str) -> Dict[str, int]:
...
 
-    app = get_current_app()
-    res = app.data.elastic.search(query, "events", params={"size": 0})
-    return _get_aggregation_values(res.hits["aggregations"])
+    return await _run_suggestions_query(query)
+
+
+async def _run_suggestions_query(query: Dict[str, Any]) -> Dict[str, int]:
+    async_app = get_current_async_app()
+    index = async_app.elastic.get_elastic_index_name(UNIFIED_RESOURCE)
```
MarkLark86: Instead of getting the index this way, could you just use the `UnifiedPlanningResource.get_service().search` directly?

This applied the index name automatically, and you won't need the `UNIFIED_RESOURCE` variable here

MarkLark86 (approved):

PR #2942, opened 2026-08-12
[SDBELGA-1114] Merge locking code to planning.locks module

MarkLark86:
### Purpose
Move all locking code to a single module, and use the UnifiedResource service instead of Event & Planning eve services.

### What has changed
* Fix/improve types:
  * Add missing fields to LocationAddress
  * Default `link_type` in `related_events` to `primary`
  * Change `unique_id` to `integer` (matches what's in superdesk-core/metadata)
  * Use `LockFields` model with AssignmentResourceModel (same lock model used in UnifiedResource)
  * Fix AssignmentResourceModel's coverage/planning fields
* Migrate PlanningFeaturedLock from Eve resource/service to a Pydantic based one
* Migrate all locking code to a `planning.locks` module, using UnifiedResourceModel and AssignmentResourceModel where possible
* Remove old Eve based locking resources & services
* Fix Event & Planning Eve service's `get_async` to add `item_type` lookup (used by API requests through Eve)
* Fix more behave & unit tests

Resolves: [SDBELGA-1114](https://sofab.atlassian.net/browse/SDBELGA-1114)

[SDBELGA-1114]: https://sofab.atlassian.net/browse/SDBELGA-1114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/locks/views.py
```diff
@@ -0,0 +1,176 @@
...
+    if not item_id:
+        raise SuperdeskApiError.badRequestError(gettext("Item ID is required"))
+
+    try:
+        item = await resource_model.get_service().find_by_id(item_id)
+    except Exception as e:
+        from pprint import pprint
+
+        pprint(e)
+        raise Exception("FUCK")
```
BrianMwangi21: I don't think this exception was supposed to make it to the PR 😅 funny though

server/planning/locks/views.py
```diff
@@ -0,0 +1,176 @@
...
+        {
+            STATUS: STATUS_OK,
+            "_links": get_hateoas_links(updated),
+        }
+    )
+    return Response(response, status_code=201)
+
+
+@planning_lock_endpoints.endpoint("assignments/<string:item_id>/unlock", name="planning_unlock", methods=["POST"])
+async def unlock_assignment_endpoint(request: Request) -> Response:
```
BrianMwangi21, line 72: Both endpoints have the same name `planning_unlock`. I think the assignments one was probably meant to be `assignment_unlock`

server/planning/locks/lock.py
```diff
@@ -0,0 +1,214 @@
...
+        await getattr(app, f"on_lock_{item_type}").call_async(original_dict, lock_data.lock_user)
+        await signals.on_item_lock.send(original, lock_data)
+
+        updated = await service.update(original.id, lock_data.to_dict(), original=original, skip_signals=True)
+        push_notification(
+            f"{item_type}:lock",
+            item=original.id,
+            user=str(lock_data.lock_user),
+            lock_time=updated.lock_time,
+            lock_sessoin=str(updated.lock_session),
```
BrianMwangi21:
```suggestion
            lock_session=str(updated.lock_session),
```

server/planning/locks/views.py
MarkLark86: Whoops!

eos87: if it makes it to the UI, it would pretty clear to the user though 🤣

server/planning/locks/unlock.py
```diff
@@ -0,0 +1,125 @@
...
+    await signals.on_item_unlock.send(original)
+
+    updated = await service.update(original.id, updates, original=original, skip_signals=True)
+    push_notification(
+        f"{item_type_name}:unlock",
+        item=original.id,
+        user=str(original.lock_user),
+        lock_session=(original.lock_session),
+        etag=updated.etag,
+        event_ids=get_related_event_ids(original),
```
eos87: I think this breaks unlocking for assignments: `get_related_event_ids(original)` and `original.recurrence_id` assume the item has `related_events` and `recurrence_id`, but `AssignmentResourceModel` doesn't define either, so this raises AttributeError. And since service.update on line 91 has already run, the item ends up unlocked in the DB but the request 500s and the notification/on_unlocked side effects never fire. The lock path (`_update_item_lock`) guards this with `isinstance(original, UnifiedPlanningResource)`, looks like unlock needs the same. Can you double check?

server/planning/locks/assignment.py
```diff
@@ -0,0 +1,68 @@
...
+        and assignment.lock_user
+        and assignment.lock_action != "content_edit"
+        and (assignment.lock_session != current_session_id or assignment.lock_user != user_id)
+    ):
+        raise SuperdeskApiError.badRequestError(gettext("Lock Failed: Related assignment is locked."))
+
+
+async def sync_content_lock_to_assignment(item: dict, user_id: ObjectId) -> None:
+    assignment = await get_assignment_from_content_dict(item)
+    if not assignment or (assignment.planning and assignment.planning.multiple_content is True):
```
eos87, line 46: I think this could crash with AttributeError. AssignmentCoverage doesn't declare multiple_content. That means assignment.planning.multiple_content fails. Can you double check?

MarkLark86: @eos87 This whole `AssignmentResourceModel` is not right. I have fixed the type for now so it has `planning.multiple_content`, but I will need to thoroughly re-write that whole model (especially to re-use as much fields from the UnifiedResourceModel etc).

I will create a ticket for this task

MarkLark86:
Thanks @eos87 Please have a look over my changes

eos87 (approved):

PR #2943, opened 2026-08-13
[SDBELGA-1127] Update search service to use UnifiedPlanningResource

MarkLark86:
### Purpose
Update the events, planning & assignment search service to use the UnifiedPlanningResource

### What has changed
* Remove redundant search service (there was an internal search service that provided the ability to search events, planning and assignments). The main search endpoint was just a proxy to this service
* Update the main search endpoint to directly use the UnifiedPlanningResource
* Fix/update types:
  * If a `language` is supplied, use that for `languages` if that is not defined
  * Set the `coverage_id` for `scheduled_updates`
  * Fix checking schedule with no end time
* Fix UnifiedPlanningResource service: set `_created`, `_updated` and `versioncreated` where appropriate
* Fix Planning eve service: map `description_text` to `definition_long`
* Copy and improve `get_hateoas_links` and `get_item_type_name` functions from the lock PR
* Run black util (2 files changed that had nothing to do with this PR, but `black` was failing locally)
* Fix cyclic imports in `planning.__init__py`
* Fix tests

Resolves: [SDBELGA-1127](https://sofab.atlassian.net/browse/SDBELGA-1127)
Depends on: https://github.com/superdesk/superdesk-core/pull/3286

[SDBELGA-1127]: https://sofab.atlassian.net/browse/SDBELGA-1127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/planning_article_export.py
```diff
@@ -67,22 +69,33 @@ class PlanningArticleExportResource(Resource):
...
-        resource_type,
-        {"item_ids": ",".join(ids_string), "only_future": False},
-        page_size=EXPORT_FETCH_PAGE_SIZE,
-    )
-    items = sorted([item async for item in items], key=lambda i: ids_string.index(str(i.get("_id"))))
 
-    events_service = EventsAsyncService()
+    query: dict = {"_id": {"$in": ids}}
+    if resource_type and resource_type != SearchItemType.COMBINED.value:
+        query["type"] = "event" if resource_type == SearchItemType.PLANNING.value else resource_type
```
BrianMwangi21:
```suggestion
        query["type"] = "event" if resource_type == SearchItemType.EVENT.value else resource_type
```

BrianMwangi21: Should this be `EVENT` instead ?

BrianMwangi21 (approved):
Looks good. Just that one comment

server/planning/search/queries/common.py
```diff
@@ -349,8 +353,8 @@ def add_field_exist_query():
             return
 
         async for item in locked_items:
-            related_primary_events = get_related_event_ids_for_planning(item, "primary")
-            if item.get("recurrence_id"):
+            related_primary_events = get_related_event_ids(item, RelatedEventLinkType.PRIMARY)
+            if item.recurrence_id:
                 # This item is associated with a recurring series of events
                 # Add `recurrence_id` to the query (common field to both events & planning)
                 recurrence_ids.add(item["recurrence_id"])
```
eos87: I think this would fail as `item` is a pydantic model instance and it is not subscriptable. Should probably be updated to 

```suggestion
                recurrence_ids.add(item.recurrence_id)
```

server/planning/planning/planning.py
```diff
@@ -118,6 +122,8 @@ async def find_one_async(self, req, **lookup):
         if item:
             dates = item.pop("dates", {})
             item["planning_date"] = dates.get("start")
+            if definition_long := item.pop("definition_long", None):
+                item["description_text"] = definition_long
```
eos87: This maps definition_long back to description_text for single-item GETs (also create and update), but the search endpoint (events_planning_search) and the collection GET return the raw docs, so planning items fetched through search carry definition_long and no description_text. Do you think this could break the frontend somehow? or you intent to update later on in a subsequent PR?

MarkLark86: I updated to use the eve hooks (on_fetched, on_create/d, on_update/d) to implement these changes now. These are used also by the search service, so should also be handled there as well

MarkLark86: I reverted some of it back, as `create_async` and `update_async` aren't always guarantee'd to be called along with their before/after hooks

eos87 (approved):

PR #2945, opened 2026-08-13
[SDESK-7989] - Cannot save coverage priority

BrianMwangi21:
### Purpose
This PR fixes an issue where setting a coverage priority and saving looked like it didn't persist. The value was actually saved but it just rendered empty after collapsing and re-expanding the coverage panel, making it appear unsaved.

### What has changed
- The priority/urgency field's `valueAdapter.getValue` used a flat `item[field]` lookup, which fails for the nested planning.priority` path. On remount (expand) the tree-select re-seeded empty. Switched to lodash `get(item, field)` so nested paths resolve.
- Added tests

### How to test
1. Enable the Priority field on a coverage profile.
2. Create a Planning item, add a Text coverage, set a priority, and save.
3. Collapse the coverage panel, then re-expand it.
4. The priority still shows its value (previously it went empty).

Resolves : [SDESK-7989](https://sofab.atlassian.net/browse/SDESK-7989)

[SDESK-7989]: https://sofab.atlassian.net/browse/SDESK-7989?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/fields/resources/common.tsx
+++ b/client/components/fields/resources/common.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable react/no-multi-comp */
 import React from 'react';
+import {get} from 'lodash';
 
 import {IVocabularyItem} from 'superdesk-api';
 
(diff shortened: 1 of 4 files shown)


eos87 (approved):
Love to see the E2E there :star_struck:

BrianMwangi21:
@eos87 nothing takes off without the e2e

PR #2946, opened 2026-08-13
[STT-1738] Fix list row spacing and trim the coverage card in previews

eos87:
### Purpose

Cards for linked events and planning items in the preview panel space their fields inconsistently: a location sits further right than the field above it, and the gap before a right-aligned field is double every other gap on the row. The coverage card also shows "Added to workflow" on coverages the system added automatically, where it says nothing, and labels the desk name with "Desk" where the name alone is clear.

### What has changed

**Row spacing.** Three things were adding space on top of the row's 8px gap:

- `LineItems` always rendered an empty `ms-auto` divider between the start and end items. It's zero-width but still a flex item, so it picked up a gap on both sides: 16px where 8 was intended. It now renders only when there are end items, and cancels one gap with a negative margin.
- `location` and `reference` each added their own 10px `sd-list-item__element-lm-10`. No other field does that. Removing it left `location`'s `noMargin` option as dead code (nothing ever set it), so that went too.
- `anpa_category` was the only field whose `Spacer` skipped `noWrap`, so with `hideLabel: true` its falsy label still emitted an empty div and 4px of indent.

Most visible on the narrow preview cards, but it's every list row in the module. Fields at `position: 'end'` now sit 8px closer to the start.

**Coverage card.**

- "Added to workflow" only shows when adding to workflow was manual, matching the server condition in `planning_service.py`. With auto assign on it appeared on every coverage and just repeated the "active" label beside it.
- The "Desk" prefix is gone in previews, the name stays.
- `title` attributes on coverage type and desk name, which have no visible label.

### Screenshots - Before // After
image

image

### Steps to test

Row spacing, any configuration:

1. Preview a planning item linked to an event that has a location. The location lines up with the field above it, and the gap before it matches every other gap on the row.
2. Open the Events and Planning lists. Any field configured at `position: 'end'` sits 8px closer to the start items; rows with no end fields are unchanged. Nothing else moves.
3. If a line in your configuration starts with `anpa_category` and `hideLabel: true`, it now starts flush with the line below it instead of 4px in.

Coverage card:

4. With `PLANNING_AUTO_ASSIGN_TO_WORKFLOW = False`, preview a planning item with a coverage in workflow. "Added to workflow" shows, as before.
5. Set it to `True`, restart the backend, reload. The label is gone.
6. Leave it on and switch "Forward Planning" on for that planning item. The label comes back.
7. The desk on a coverage card in a preview shows the name with no "Desk" prefix; hovering shows it as a tooltip. In the coverage editor the prefix is still there.

Resolves: [STT-1738]

[STT-1738]: https://sofab.atlassian.net/browse/STT-1738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/client/components/Coverages/CoverageItem.tsx
+++ b/client/components/Coverages/CoverageItem.tsx
@@ -206,9 +206,9 @@ export class CoverageItemComponent extends React.Component<IProps, IState> {
             isPreview,
         } = this.props;
 
-        // The label only says something when adding to workflow was a manual step. The server adds
-        // automatically when PLANNING_AUTO_ASSIGN_TO_WORKFLOW is on and the planning item does not
-        // set the override flag (see `planning_service.py`).
+        // Nobody chose to add this coverage to workflow: the server did it on creation, and the
+        // action that would have done it by hand is hidden in this state. The status label says
+        // "active" instead, which is all there is to say.
         const addedToWorkflowAutomatically = appConfig.planning_auto_assign_to_workflow === true &&
             item.flags?.overide_auto_assign_to_workflow !== true;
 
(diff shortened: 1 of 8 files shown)


eos87:
Closing in favour of a develop-first branch. The same commits will land on develop with e2e coverage, then get backported to release/3.5.

PR #2947, opened 2026-08-14
[SDBELGA-1115] - Merge Event & Planning Files

BrianMwangi21:
### Purpose
Merge the Event & Planning file resources, services and endpoints onto a single shared service over the unified resource.

### What has changed
- Fold `EventsFilesService` & `PlanningFilesService` into one shared `FilesAsyncService` (`planning/files.py`) over the single `events_files` collection.
- Keep both `events_files` and `planning_files` endpoints (and their privileges) backed by the shared service — no client changes.
- Unify the delete guard against `unified_planning` — protects a file used by an item's `files` or a coverage's `files`/`xmp_file`.
- Remove the old `events/events_files.py` & `planning/planning_files.py`.
- Add `tests/unified_resource/files_test.py`.

Resolves: [SDBELGA-1115](https://sofab.atlassian.net/browse/SDBELGA-1115)

[SDBELGA-1115]: https://sofab.atlassian.net/browse/SDBELGA-1115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MarkLark86 (approved):
Although Eve is deprecated and we would prefer using the newer framework, this would probably have taken a lot longer to achieve.
I suppose in this case we can accept using Eve still.

PR #2951, opened 2026-08-17
[SDBELGA-1117] Merge history resources

MarkLark86:
### Purpose
Merge Event & Planning history resources into a unified history resource

### What has changed
* Move all history code into a common module `planning.history`
* Merge Event & Planning history resources into the 1
* Update front-end code to use the unified resource endpoint and updated schema

Resolves: [SDBELGA-1117](https://sofab.atlassian.net/browse/SDBELGA-1117)

[SDBELGA-1117]: https://sofab.atlassian.net/browse/SDBELGA-1117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/history/planning.py
```diff
@@ -2,38 +2,71 @@
...
+        for item in items:
+            if item.get("type") == "event":
+                events.append(item)
+            elif item.get("type") == "planning":
+                planning.append(item)
+            else:
+                logger.warning(
+                    "Received an incorrect type for UnifiedPlanningHistory",
+                    extra=dict(item_id=item.get("_id"), item_type=item.get("type")),
+                )
```
eos87, line 41: I'm wondering if we should raise an exception or validation error instead of just logging here. Thoughts?

server/planning/history/planning.py
```diff
@@ -42,15 +75,24 @@ async def _save_history(self, item, update: dict[str, Any], operation: str | Non
...
+            "item_type": item.get("type"),
             "user_id": user,
             "operation": operation,
             "update": update,
         }
 
-        if operation == "create" and update.get("state", "") == "ingested":
-            history["operation"] = "ingested"
+        if not history.get("item_type"):
+            raise Exception("ITEM TYPE NOT SUPPLIED")
```
eos87: is this a left-over or intended behavior? I'm wondering because by the time this exception is raised the item is fully updated in DB. Also, for recurring events, it will blow up mid-loop, so some events in the series get saved and the rest don't

server/planning/history/assignments.py
```diff
@@ -48,7 +51,8 @@ async def _save_history(self, item, update: dict[str, Any], operation: str | Non
             update["assigned_to"] = {"user": user_id}
 
         history = {
-            "assignment_id": item[ID_FIELD],
+            "item_id": item[ID_FIELD],
```
eos87, line 54: This still needs updating from the frontend I think. In `assignments/api.ts` is still being called like `where: {assignment_id: assignment._id}`

eos87 (changes requested):
is the prod-api part going to be tackle separately? because I think it is still pointing to the old resource if I'm not mistaken.

server/planning/history/planning.py
MarkLark86: I had thought about that too, raise exception or log a warning.
As the history is just informational metadata for an item, and has no affect of the item itself, I made the decision to log the warning. If a history entry is missing, it should not be a huge deal.

If we raised an exception, the user would receive an error in the front-end that the request failed, even though the base item would have updated.

server/planning/history/planning.py
MarkLark86: A left-over. I used it to find the areas that were failing to send the `type`, as it failed during running tests and gave me a traceback of when this data was not sent. Not ideal, but it worked.

Based on above comment, I have updated it to log a warning and return without creating an entry

server/planning/history/assignments.py
MarkLark86: I looked over the front-end code and could only find this one occurrence you found

MarkLark86:
> is the prod-api part going to be tackle separately? because I think it is still pointing to the old resource if I'm not mistaken. 

We have a separate ticket for that one, and CAPI

eos87 (approved):

PR #2954, opened 2026-08-18
[SDBELGA-1119] - Merge Event & Planning Action Endpoints

BrianMwangi21:
### Purpose
Merge the parallel Event & Planning `spike`/`unspike`, `postpone` and `duplicate` action logic into a single `item_type`-dispatched implementation, the same way Events & Planning items/files/autocomplete were merged.

### What has changed
- New `planning/unified/actions/` package (`spike.py`, `postpone.py`, `duplicate.py`) holding the merged `process_*` entry points, dispatched on `item_type`; the five old `events_*`/`planning_*` action modules are removed.
- Both the `events/*` and `planning/*` action endpoints are kept and now call the same `process_*` functions.
- Correctness fixes for the shared `unified_planning` collection (the Eve→unified datalayer applies no `type` filter):
  - Cascades + `validate_spike_event` query the unified index via `get_related_planning_for_events` (the legacy `planning` index is empty).
  - `validate_recurring_event` / `events_utils.get_recurring_timeline` are scoped to `type=event` so planning items sharing a `recurrence_id` don't leak into the event series.
  - `process_postpone_planning_item` writes to the unified planning service, and planning postpone now records history.
  - The spike guard `event_has_planning_items` reads the unified index.
- Add `tests/unified_resource/actions_test.py`.

Resolves: [SDBELGA-1119](https://sofab.atlassian.net/browse/SDBELGA-1119)

[SDBELGA-1119]: https://sofab.atlassian.net/browse/SDBELGA-1119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/publish/events_post.py
```diff
@@ -236,7 +236,7 @@ async def post_event(self, event, new_post_state, repost, assignments_to_delete:
         return event, failed_planning_ids
 
     async def post_related_plannings(self, plannings, new_post_state):
-        from planning.planning.planning_spike import process_spike_planning_item
+        from planning.unified.actions import process_spike_planning_item
```
MarkLark86: Can this be moved to the top of the file now?

server/planning/unified/actions/postpone.py
```diff
@@ -133,3 +145,66 @@ async def process_postpone_event(updates: dict[str, Any], original: dict[str, An
...
+    planning_service = get_resource_service("planning")
+
+    # Postpone the planning_item using reason in updates
+    updates["state_reason"] = updates.get("reason")
+    updates[ITEM_STATE] = WORKFLOW_STATE.POSTPONED
+
+    updates["coverages"] = deepcopy(original.get("coverages"))
+    coverages = updates.get("coverages") or []
+
+    for coverage in coverages:
```
MarkLark86: Event's now support coverages. We should also add this to the processing of an Event

server/planning/unified/actions/postpone.py
```diff
@@ -133,3 +145,66 @@ async def process_postpone_event(updates: dict[str, Any], original: dict[str, An
...
+                    else None
+                ),
+                message="assignment_postponed_msg",
+                slugline=slugline,
+                coverage_type=get_coverage_type_name(coverage_type),
+            )
+
+
+async def process_postpone_planning_item(updates: dict[str, Any], original: dict[str, Any]) -> dict[str, Any]:
+    planning_service = get_resource_service("planning")
```
MarkLark86: We should be using the [UnifiedPlanningResource](https://github.com/superdesk/superdesk-planning/blob/feature/unified/server/planning/types/__init__.py#L108) to get/update items, and using it's [on_update](https://github.com/superdesk/superdesk-planning/blob/feature/unified/server/planning/unified/service.py#L193) etc methods, as the Eve service's methods are now gutted/removed of logic

server/planning/unified/common.py
```diff
@@ -151,6 +151,19 @@ async def get_related_planning_for_events(
...
 
+async def event_has_planning_items(
+    event_id: str, link_type: RelatedEventLinkType | None = RelatedEventLinkType.PRIMARY
+) -> bool:
+    """Whether an Event has related Planning items, querying the unified index.
+
+    Replaces the legacy ``planning.utils.event_has_planning_items`` which reads the
+    (now empty) legacy ``planning`` elastic index.
+    """
+    async for _ in await get_related_planning_for_events([event_id], link_type):
```
MarkLark86: Elasticsearch would return the entire documents for all matched items (unlike MongoDB, Elastic returns all items, doesn't have a proper async cursor), which adds network and processing resource requirements for Elasticsearch.

Ideally what we should use here is the `count` function, as elasticsearch only performs a count items

MarkLark86 (commented):
Mostly looks fine, but we need to use the UnifiedPlanningResource and also support Converages in Events

server/planning/unified/actions/postpone.py
```diff
@@ -120,14 +139,15 @@ async def process_postpone_event(updates: dict[str, Any], original: dict[str, An
...
-    updates.pop("skip_on_update", None)
 
-    # Update the original event in the database
+    # Update the original event in the database. Go through the service's on_update,
+    # but skip the (still unimplemented) recurring date-update logic via
+    # `skip_on_update`; on_update strips the flag before persisting.
+    updates["skip_on_update"] = True
     event_id = original[ID_FIELD]
-    await events_service.update_async(event_id, updates, original, skip_signals=True)
+    updated = await service.update(event_id, updates)
```
MarkLark86: The original Eve service's `update_async` doesn't run `on_update/d` functions. In the Pydantic service we have the [skip_signals](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/core/resources/service.py#L556) param which basically does the same thing.

Should we use `skip_signals=True` here, otherwise the `on_updated` might have side effects not desired here?

Any other parts of the code where we used the Eve's `update_async`, we should probably use the same param.

Anywhere Eve's `patch_async` method is used, we should use the `update` method without the `skip_signals` param

server/planning/unified/actions/spike.py
```diff
@@ -147,25 +147,29 @@ async def validate_spike_event(event: dict[str, Any]) -> None:
...
-    # Scope series lookups by `type` (shared collection, no datalayer filter)
-    async for event in await events_service.find_async(
-        {"recurrence_id": recurrence_id, "type": PlanningItemType.EVENT.value}
-    ):
+    # Scope series lookups by `type` (shared collection, no datalayer filter).
+    # `max_results=0` disables the per-page cap so the whole series is checked.
+    events = await service.find(
+        {"recurrence_id": recurrence_id, "type": PlanningItemType.EVENT.value}, max_results=0, use_mongo=True
+    )
+    for event_obj in await events.to_list():
```
MarkLark86: If we use `async for event_obj in events` it will iterate over the items, placing 1 item in memory at a time.
`to_list` places all items in memory at once

server/planning/unified/actions/spike.py
```diff
@@ -147,25 +147,29 @@ async def validate_spike_event(event: dict[str, Any]) -> None:
...
         if event.get(LOCK_USER) or event.get(LOCK_SESSION):
             raise SuperdeskApiError.forbiddenError(message=_("Spike failed. An event in the series is locked."))
 
-    async for planning in await planning_service.find_async(
-        {"recurrence_id": recurrence_id, "type": PlanningItemType.PLANNING.value}
-    ):
+    plannings = await service.find(
+        {"recurrence_id": recurrence_id, "type": PlanningItemType.PLANNING.value}, max_results=0, use_mongo=True
+    )
+    for planning_obj in await plannings.to_list():
```
MarkLark86: Same here, use `async for` iterator

server/planning/unified/actions/spike.py
```diff
@@ -402,7 +407,7 @@ async def post_planning_item_spike_actions(updates: dict[str, Any], original: di
     first_event_id = get_first_related_event_id_for_planning(original, "primary")
 
     if first_event_id:
-        event = await events_service.find_one_async(req=None, _id=first_event_id)
+        event = await service.find_by_id_raw(first_event_id)
```
MarkLark86: If we use the Pydantic model instances, it provides type checking and better developer experience (with autocomplete etc).
In this case:
```python
event = await service.find_by_id(first_event_id)
notify_user_on_failed_assignment_deletes = not event or event.state != WorkflowState.SPIKED
```

Also notice the enum change `WorkflowState` (new one using enum) instead of `WORKFLOW_STATE` (old one using NamedTuple)

server/planning/unified/actions/spike.py
```diff
@@ -500,23 +505,22 @@ async def process_unspike_planning_item(updates: dict[str, Any], original: dict[
...
     :return: The updated planning document.
     """
-    planning_service = get_resource_service("planning")
-    events_service = get_resource_service("events")
+    service = UnifiedPlanningResource.get_service()
 
     first_event_id = get_first_related_event_id_for_planning(original, "primary")
     if first_event_id:
-        event = await events_service.find_one_async(req=None, _id=first_event_id)
+        event = await service.find_by_id_raw(first_event_id)
```
MarkLark86: Same here, using the Pydantic model instance would be preferable

MarkLark86 (commented):
Looks good, just a couple smaller things better replication of current logic (regarding using `update` method on Pydantic based services)

server/planning/unified/actions/postpone.py
BrianMwangi21: Makes sense. 

I'd used `skip_on_update` so it'd still go through `on_update` and only skip the `_update_recurring_events` TODO that crashes, rather than skipping the hooks entirely.

I've made the changes.

MarkLark86 (approved):

PR #2956, opened 2026-08-19
[SDESK-7993] Reflect related event changes in the open planning preview

eos87:
### Purpose
When a related event is edited (e.g. renamed through the embedded editor inside the planning editor), an open planning preview kept showing the old values until it was closed and reopened. The list rows and the editor cards updated fine, only the preview lagged behind.

### What has changed
- The preview's Related Events section now renders through `WithLiveResources` instead of the redux events store. The store copy is not refreshed on `events:updated` notifications outside the Events view, which is the root cause of the staleness; the nested list rows already use `WithLiveResources`, so this brings the preview in line with them.
- New e2e spec covering the flow end to end: rename the related event in the embedded form, save, and both the preview card and the nested list row update without reopening anything. Comes with a new `EmbeddedEventEditor` page object.
- Two small console noise fixes: `IconButton`'s `refNode` prop type now accepts callback refs, and the planning loader is unmounted at the container it was rendered into.
- `@playwright/test` and `@superdesk/build-tools` bumped in e2e.

### Steps to test
1. In the Planning view, have a planning item linked to an event.
2. Open the planning item in the editor, then click it in the list so the preview opens next to it.
3. Expand the related event in the editor ("Show more"), change its name and save.
4. The Related Events card in the preview and the nested event row in the list should both show the new name within a second or so, without closing or reopening anything.

### Front-end checklist

### DIFF
--- a/client/components/Planning/PlanningPreviewContent.tsx
+++ b/client/components/Planning/PlanningPreviewContent.tsx
@@ -33,7 +33,7 @@ import {EventMetadata} from '../Events';
 import {FeatureLabel} from './FeaturedPlanning';
 import {renderProfileGroupedFields} from '../fields';
 import {PreviewFieldFiles} from '../fields/preview/Files';
-import {getRelatedEventIdsForPlanning} from '../../utils/planning';
+import {getRelatedEventIdsForPlanning, pickRelatedEventIdsForPlanning} from '../../utils/planning';
 import {coverageProfiles} from '../../selectors/coverageProfiles';
 import {getCoverageFields} from '../../api/editor/item_planning';
 import {appConfig} from 'appConfig';
@@ -130,11 +130,33 @@ export class PlanningPreviewContentComponent extends React.PureComponent<IProps>
     }
 
     render() {
+        const eventIds = pickRelatedEventIdsForPlanning(this.props.item, 'display');
+
+        if (eventIds.length < 1) {
+            return this.renderContent(null);
+        }
+
+        const {WithLiveResources} = superdeskApi.components;
+
+        // Related events are rendered from live resources rather than the redux store: the store
+        // copy is not refreshed on `events:updated` notifications outside the Events view, so it
+        // goes stale when an event changes while the preview is open
+        return (
+            <WithLiveResources resources={[{resource: 'events', ids: eventIds}]}>
+                {([res]) => this.renderContent(
+                    (res._items as Array<IEventItem>)
+                        .filter((event) => event != null)
+                        .map((event) => eventUtils.modifyForClient(event))
+                )}
+            </WithLiveResources>
+        );
+    }
+
+    renderContent(relatedEvents: Array<IEventItem> | null) {
         const {gettext} = superdeskApi.localization;
         const {item,
             users,
             formProfile,
-            relatedEvents,
             onEditEvent,
             noPadding,
             hideRelatedItems,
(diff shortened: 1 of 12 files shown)


client/components/Planning/PlanningPreviewContent.tsx
```diff
@@ -130,11 +130,33 @@ export class PlanningPreviewContentComponent extends React.PureComponent<IProps>
     }
 
     render() {
+        const eventIds = pickRelatedEventIdsForPlanning(this.props.item, 'display');
+
+        if (eventIds.length < 1) {
+            return this.renderContent(null);
+        }
+
+        const {WithLiveResources} = superdeskApi.components;
```
MarkLark86, line 133: Under the hood the `WithLiveResources` uses the [fetchChangedResources](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/helpers/CrudManager.tsx#L142C17-L142C38) function.

This might need to be looked at for performance reasons. Seems like that function uses [httpRequestJsonLocal](https://github.com/superdesk/superdesk-client-core/blob/develop/scripts/core/helpers/CrudManager.tsx#L184) for every instance of this react component. Meaning every component listening for a resource change, each one will send a request to the server.

Is this correct, or is there a client-side cache involved somewhere I am missing?
I just don't want to increase the server load (which is why we used `ws notification` -> `api request` -> `redux state` -> `listeners`, so we only send 1 request to the server, and update locally all those listeners)

eos87: yeah I'm aware of it, and you're right, it does fetch per instance from the server. The addition here is almost negligible though: it's one more instance of a pattern we already use (the embedded editor card headers and the expanded list rows), we only have one preview open at a time, it loads all the related events in a single query, and the fetcher is stopped once the component is unmounted. I will definitely revisit WithLiveResources for performance (there are a couple of improvements I think I can add).

The other option was to keep the store in sync (it goes stale outside the Events view), but that opened a different can of worms, so I went with the approach that's consistent with the rest of the code.

MarkLark86: `expanded list rows`, how is the performance here using something like this in a list (where we should be rendering as fast as possible? We have a config to to auto-expand then rows.
If there are 100 items in the list expanded, does that mean 100 instances of the `WithLiveResources` each sending unique queries to the server?

eos87: This PR doesn't affect that at all, it only touches the planning preview, so whatever the performance of expanded list rows is today, it's exactly the same after this is merged. Also, the auto-expand config (I assume  `planning_expand_related_plannings`) applies to related plannings under events, and that path renders from the redux store, not `WithLiveResources`. The live-resources rows (events under plannings) are collapsed by default with no auto-expand, so instances only exist for rows someone manually opened. I haven't measured the many-rows-expanded case, but that's pre-existing behaviour worth its own ticket if it's a concern.

client/components/Planning/PlanningPreviewContent.tsx
```diff
@@ -130,11 +130,33 @@ export class PlanningPreviewContentComponent extends React.PureComponent<IProps>
...
+
+        // Related events are rendered from live resources rather than the redux store: the store
+        // copy is not refreshed on `events:updated` notifications outside the Events view, so it
+        // goes stale when an event changes while the preview is open
+        return (
+            <WithLiveResources resources={[{resource: 'events', ids: eventIds}]}>
+                {([res]) => this.renderContent(
+                    (res._items as Array<IEventItem>)
+                        .filter((event) => event != null)
+                        .map((event) => eventUtils.modifyForClient(event))
```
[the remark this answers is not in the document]
eos87: this is a good catch, I've improved it.

eos87:
Behave tests are broken for unrelated changes to this PR. It seems we need to fix them from superdesk-core.

petrjasek (approved):

PR #2960, opened 2026-08-24
[SDBELGA-1120] - Update Resource Action endpoints to use unified resource

BrianMwangi21:
### Purpose
Repoint the remaining Event & Planning action endpoints — `update_time`, `cancel`, `reschedule`, `update_repetitions` — off the old Eve resources and onto the unified `unified_planning` resource, folding in the two still-Eve `planning_cancel`/`planning_reschedule` services.

### What has changed
- Relocate the four action modules into `planning/unified/actions/` (`cancel.py`, `reschedule.py`, `update_time.py`, `update_repetitions.py`); fold the Planning side of cancel & reschedule (the Eve `PlanningCancelService`/`PlanningRescheduleService`) into them and remove those Eve resources.
- Endpoints kept unchanged (names/URLs/privileges); `planning_cancel` becomes an async endpoint, and cancel/reschedule history + notifications now fire inside the unified functions.
- Correctness fixes for the shared `unified_planning` collection (the Eve→unified datalayer applies no `type` filter):
  - Related-planning cascades query the unified index via `get_related_planning_for_events`, and `event_has_planning_items` reads the unified index (the legacy `planning` index is empty).
  - `update_repetitions.get_internal_series` is scoped to `type=event` so planning items sharing a `recurrence_id` don't leak into the event series.
- Writes go through the unified Pydantic service (`update`/`system_update`, with `skip_signals` where hooks shouldn't run) instead of the Eve proxy.
- Add `tests/unified_resource/actions_lifecycle_test.py`.

Resolves: [SDBELGA-1120](https://sofab.atlassian.net/browse/SDBELGA-1120)

> ⚠️ Stacked on top of #2954 (SDBELGA-1119).

[SDBELGA-1120]: https://sofab.atlassian.net/browse/SDBELGA-1120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/planning/unified/actions/cancel.py
```diff
@@ -0,0 +1,335 @@
...
+    coverage_cancel_state = get_coverage_status_from_cv("ncostat:notint")
+    coverage_cancel_state.pop("is_active", None)
+
+    planning_item_id = original[ID_FIELD]
+    ids = []
+    updates["coverages"] = deepcopy(original.get("coverages"))
+    coverages = updates.get("coverages") or []
+    reason = updates.pop("reason", None)
+
+    for coverage in coverages:
```
MarkLark86: Event's can also have coverages now, will need to handle this as well when cancelling an Event

server/planning/unified/actions/cancel.py
```diff
@@ -0,0 +1,335 @@
...
+        if coverage["workflow_status"] not in [
+            WORKFLOW_STATE.CANCELLED,
+            ASSIGNMENT_WORKFLOW_STATE.COMPLETED,
+        ]:
+            ids.append(coverage.get("coverage_id"))
+            await planning_service.cancel_coverage(
```
MarkLark86: This is now in the [unified services](https://github.com/superdesk/superdesk-planning/blob/feature/unified/server/planning/unified/coverages.py#L568), which will need you to pass in the UnifiedPlanningResource instance

server/planning/planning/views.py
```diff
@@ -82,3 +88,21 @@ async def postpone_planning_item(args: PlanningArgs, params: None, request: Requ
...
+
+
+@planning_endpoint_group.endpoint(
+    "planning/cancel/<string:planning_id>",
+    name="planning_cancel",
+    methods=["PATCH"],
+    auth=[required_privilege_rule("planning_planning_management")],
+)
+async def cancel_planning_item(args: PlanningArgs, params: None, request: Request) -> Response:
+    original = await get_resource_service("planning").find_one_async(req=None, _id=args.planning_id)
```
MarkLark86: It would be preferable to use the newer UnifiedPlanningResource when interacting with these resources where possible

MarkLark86 (approved):

PR #2962, opened 2026-08-24
fix edit coverages modal opening

petrjasek:
SDESK-8012

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/components/Planning/PlanningItem.tsx
+++ b/client/components/Planning/PlanningItem.tsx
@@ -61,6 +61,8 @@ class PlanningItemComponent extends React.Component<IProps, IState> {
         this.onAddCoverageButtonClick = this.onAddCoverageButtonClick.bind(this);
         this.onItemHoverOn = this.onItemHoverOn.bind(this);
         this.onItemHoverOff = this.onItemHoverOff.bind(this);
+        this.openCoverageModal = this.openCoverageModal.bind(this);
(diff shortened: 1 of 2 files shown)


client/components/Planning/PlanningItem.tsx
```diff
@@ -61,6 +61,8 @@ class PlanningItemComponent extends React.Component<IProps, IState> {
         this.onAddCoverageButtonClick = this.onAddCoverageButtonClick.bind(this);
         this.onItemHoverOn = this.onItemHoverOn.bind(this);
         this.onItemHoverOff = this.onItemHoverOff.bind(this);
+        this.openCoverageModal = this.openCoverageModal.bind(this);
+        this.closeCoverageModal = this.closeCoverageModal.bind(this);
```
eos87, line 65: if we use arrow functions for both `openCoverageModal` and `closeCoverageModal` we do not need this bind as it is done automatically :)

eos87 (approved):

client/components/Planning/PlanningItem.tsx
petrjasek: 🤯

petrjasek: we should probably define preferred style and use it consistently, will keep it for now

PR #2966, opened 2026-08-27
[SDBELGA-1123] - Update Agenda services to use unified resource

BrianMwangi21:
### Purpose
Repoint all Agenda server-side access to Planning data off the Eve `get_resource_service("planning")` proxy (and the equally-wrong native `PlanningAsyncService`, which reads the empty legacy `planning` collection) and onto the unified `unified_planning` Pydantic service, so no server-side `get_resource_service` for events/planning remains in the agenda code. This also collapses the dual-registered Agenda resource onto its native Pydantic service and removes the old Eve one.

### What has changed
- Add type-scoped `get_planning_by_agenda_id` / `agenda_has_planning_items` helpers to `unified/common.py`; `AgendasAsyncService` now reads related Planning items from `unified_planning`, fixing the `plannings` enrichment and the delete-guard that silently read the empty legacy `planning` collection.
- Relocate the Agenda service into `unified/agenda/` (mirrors the Templates merge) and enable its native Pydantic REST endpoint with privileges; remove the Eve `agendas.py` resource.
- Repoint the remaining `get_resource_service("agenda")` call-sites to the native service, and convert the ingest routing-rule `add_planning_agendas` to async.
- Remove the now-dead `get_planning_by_agenda_id` from the Eve/native Planning services.
- Client-facing `agenda` endpoint unchanged (URL, methods, privileges, notifications).

Resolves: [SDBELGA-1123](https://sofab.atlassian.net/browse/SDBELGA-1123)

[SDBELGA-1123]: https://sofab.atlassian.net/browse/SDBELGA-1123


server/planning/io/ingest_rule_handler.py
```diff
@@ -165,7 +166,7 @@ def add_planning_agendas(self, ingest_item: Dict[str, Any], attributes: Dict[str
 
         # Append Agenda IDs found onto the item
         updates = {"agendas": ingest_item["agendas"] + new_agenda_ids}
-        updated_item = get_resource_service("planning").patch(ingest_item.get(ID_FIELD), updates)
+        updated_item = await get_resource_service("planning").patch_async(ingest_item.get(ID_FIELD), updates)
```
MarkLark86: It would be good to update this to the UnifiedPlanningResource

server/planning/io/ingest_rule_handler_test.py
```diff
@@ -146,7 +147,7 @@ async def test_skips_disabled_and_existing_calendars(self):
         self.assertEqual(calendars[0], "sports")
 
     async def test_adds_planning_agendas(self):
-        self.app.data.insert("agenda", self.agendas)
+        await AgendasResourceModel.get_service().mongo_async.insert_many(self.agendas)
```
MarkLark86: This is also `test_utils.post_items('agenda', self.agendas)` that you can use (it automatically uses Eve or Pydantic, whichever is available)

server/planning/unified/common.py
```diff
@@ -179,6 +179,29 @@ async def event_has_planning_items(
     return await UnifiedPlanningResource.get_service().count(query) > 0
 
 
+def _planning_by_agenda_query(agenda_id) -> dict:
+    return {
+        "query": {
+            "bool": {
+                "filter": [
+                    {"term": {"agendas": str(agenda_id)}},
+                    {"term": {"type": PlanningItemType.PLANNING.value}},
```
MarkLark86: The `type` filter won't be needed, as a unified item (event or planning) can now have an Agenda as well

MarkLark86 (commented):
Looks good, just a couple small improvements

MarkLark86 (approved):

PR #2967, opened 2026-08-31
[SDBELGA-1126] Merge PlanningProfiles and CoverageProfiles

MarkLark86:
### Purpose
Merge `planning_types` and `coverage_profiles` into the 1 managed resource, and allow any supported field on either Event or Planning (such as Coverages, Agendas etc)

### What has changed
**Back-end:**
* Update Pydantic models
  * new `PlanningProfileType` enum
  * use `ObjectId` for coverage `profile` field
  * Add validation to PlanningProfileResource
  * Re-use the `AuditInformation` mixin
* Merge PlanningProfile and CoverageProfile back-end resources:
  * Remove old eve resources
  * Fix newer Pydantic REST API endpoint (fixes previous implementation)
  * Migrate and merge logic from old eve resources
  * Move module initialisation into a `module.py` file, so `__init__.py` file is empty
  * Update profile defaults to use Pydantic models
  * Use an internally recognised `DEFAULT_PROFILE_ID` so model instances are valid, and we can determine if it's a default or a DB profile
  * Merge fields for Events & Planning so all most of the fields are available on both Events & Planning
* Add a data migration script
  * Moves CoveageProfiles into the PlanningProfiles resource
  * Migrates `description_text` of Planning resource to use `definition_long`
  * Add `type` field to all profiles stored in the DB
* Use Pydantic resource everywhere (eve resource is no longer available)
* Fix/add tests

**Front-end:**
* Merge PlanningProfiles and CoverageProfiles API functionality into 1 (they were basically identical anyway)
* Replace ICoverageContentProfile with IPlanningContentProfile

Resolves: [SDBELGA-1126](https://sofab.atlassian.net/browse/SDBELGA-1126)

[SDBELGA-1126]: https://sofab.atlassian.net/browse/SDBELGA-1126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/selectors/coverageProfiles.ts
```diff
@@ -1,32 +1,32 @@
...
 import {get} from 'lodash';
 import {createSelector} from 'reselect';
 
-export const coverageProfiles: (state: any) => Array<ICoverageContentProfile>
+export const coverageProfiles: (state: any) => Array<IPlanningContentProfile>
     = (state) => (state?.coverageProfiles?.profiles ?? []);
 
-export const oldProfile: (state: any) => ICoverageContentProfile
+export const oldProfile: (state: any) => IPlanningContentProfile
     = (state) => get(state, 'forms.profiles.coverage', {});
```
eos87: I think this needs to be updated to take the generic profile from `coverageProfiles.profiles` instead, since nothing writes forms.profiles.coverage any more. Right now it silently returns `{}`, which then could throw in `getProfileFields`. 

Also, should we rename it to `genericProfile`? `oldProfile` is a bit misleading now. And the type probably would be `IPlanningContentProfile | undefined` instead.

server/planning/content_profiles/service.py
```diff
@@ -1,195 +1,219 @@
...
+    - If the field exists in both, they're merged with database values taking precedence
+
+    Args:
+        content_profile (dict): The content profile from the database to be updated
+        default_coverage_profile (dict): The default coverage profile to merge from
+    """
+
+    # Update schema fields with database schema fields
+    updated_profile = deepcopy(default_profile)
+    updated_profile.setdefault("groups", {})
```
eos87, line 198: I think we also need a `setdefault` for `editor` and `schema` here too. The action profile defaults only pass `id`, `name` and `item_type`, and `to_dict()` serialises with `exclude_unset=True`, so neither key makes it into the dict. Or would you rather pass `editor={}`, `schema={}` in the constructors instead?

server/planning/data_updates/00039_20260828-152220_planning_types.py
```diff
@@ -0,0 +1,136 @@
...
+    ):
+        original_id = profile["_id"]
+        profile["_id"] = self._get_profile_id(original_id)
+
+        # Migrate the audit information to a common format
+        profile.update(
+            {
+                "type": PlanningProfileType.COVERAGE,
+                "original_creator": profile.pop("created_by", None),
+                "version_creator": profile.pop("updated_by", None),
```
eos87, line 116: Should we add a distinct name here too? I'm just wondering if it would fail because of the unique index.

server/planning/validate/planning_validate.py
```diff
@@ -134,14 +132,14 @@ def get_filtered_validator_schema(validator, validate_on_post: bool) -> dict:
...
-        profile = get_resource_service("coverage_profiles").find_one(req=None, _id=item["profile"])
-        if profile:
-            return profile
+    service = PlanningProfileResource.get_service()
 
-    validator = await PlanningTypesAsyncService().find_one(req=None, name=item_type)
+    if item_type == "coverage" and item.get("profile"):
+        profile = await service.find_by_id(item["profile"])
+    else:
+        profile = await service.find_one(name=item_type)
```
eos87: this should probably be `type=` now

```suggestion
        profile = await service.find_one(type=item_type)
```

eos87 (changes requested):

client/selectors/coverageProfiles.ts
MarkLark86: You raised a valid point here, and a bug that I introduced. This `oldProfile` is actually describing the default coverage profile to use where a newer content-specific profile is not available.

The bug I introduced was not handling this system/DB defined default coverage profile at all. I have fixed this up in my latest commit, and renamed this function to `defaultCoverageProfile` and added tests to cover this.

`defaultCoverageProfile` should always have a value, so I did not add `| undefined` to the return type here.

Please have a look over these changes, and let me know if something doesn't look right.

eos87: Thanks for the changes. I think the initial load is still off though. In https://github.com/MarkLark86/superdesk-planning/blob/aa6c0e6ca1c0e5589dde52a03a48cac98671618b/client/services/PlanningStoreService.ts#L277-L280 all the coverage profiles get pushed into `coverageProfiles.profiles`, including the default one, so `forms.profiles.coverage` is never set on page load. It only gets filled once a `planning_types` notification comes in and `updateProfilesInStore` runs.

Probably just needs the same content_type == null split you did in `updateProfilesInStore`. Might be worth a small client test for the initial state as well.

server/planning/validate/planning_validate.py
```diff
@@ -134,14 +132,14 @@ def get_filtered_validator_schema(validator, validate_on_post: bool) -> dict:
...
+    service = PlanningProfileResource.get_service()
 
-    validator = await PlanningTypesAsyncService().find_one(req=None, name=item_type)
+    if item_type == "coverage" and item.get("profile"):
+        profile = await service.find_by_id(item["profile"])
+    else:
+        profile = await service.find_one(type=item_type)
 
-    return validator.to_dict() if validator else None
+    return profile.to_dict() if profile else None
```
eos87, line 144: there is a chance that the coverage skips validation entirely if `find_by_id` would return `None`. The old code fell back to the default coverage profile in that case (see below)

```python
validator = await PlanningTypesAsyncService().find_one(req=None, name=item_type)
```

 What are the odds of hitting this?

server/planning/content_profiles/service.py
```diff
@@ -1,195 +1,257 @@
...
+            _merge_planning_type(profile, default_profile)
+        return profile, profile_type
+    else:
+        logger.error(
+            "Failed to merge PlanningProfile with system defaults",
+            extra={
+                "profile": profile,
+                "item_type": item_type,
+            },
+        )
```
eos87: This logs at error for any type that isn't in `PlanningProfileType` and it could fire on normal actions - I think. `is_valid_event_planning_reason` in `common.py` looks up `{item_type}_{lock_action}` so update time ends up asking for `event_update_time`, spike for `event_spiked`, etc. None of those are in the enum, so we'd log an error every time someone does one of those.

client/selectors/coverageProfiles.ts
MarkLark86: I updated the implementation to merge CoverageProfiles into all the forms functionality, this should fix the issue and remove redundant code.

eos87 (approved):
looks good. CI is broken but I assume it will be solved later in another PR.

PR #2969, opened 2026-09-01
[SDBELGA-1125] - Update Production API to use unified resources as datasources

BrianMwangi21:
### Purpose
This PR repoints the Production API Events & Planning to the unified collections

### What has changed
- `prod_api/events` and `prod_api/planning` datasources point at `unified_planning`, filtered by `type` (Mongo `filter` + `elastic_filter`).
- `events_history` points at the merged `planning_history` (`item_type=event`); `events_files` and `assignments` are unchanged.
- The `planning_source` filter reads the event link from `related_events` instead of the legacy `event_item`.
- Planning output is converted back to the legacy field shape (`planning_date`, `description_text`), so the API contract is unchanged.
- Tests seed the unified collections in the `prod_api` conftest and update the fixtures to the unified shape.

Depends on superdesk/superdesk-core#3308; without it the repointed resources fall back to Mongo and search does not filter.

Resolves: [SDBELGA-1125](https://sofab.atlassian.net/browse/SDBELGA-1125)

[SDBELGA-1125]: https://sofab.atlassian.net/browse/SDBELGA-1125

### DIFF
--- a/server/planning/prod_api/events/resource.py
+++ b/server/planning/prod_api/events/resource.py
@@ -21,9 +21,11 @@ class EventsResource(Resource):
     resource_methods = ["GET"]
     allow_unknown = True
     datasource = {
-        "source": "events",
+        "source": "unified_planning",
         "search_backend": "elastic",
         "default_sort": [("dates.start", 1)],
+        "filter": {"type": "event"},
+        "elastic_filter": {"term": {"type": "event"}},
     }
     privileges = {"GET": Scope.EVENTS_READ.name}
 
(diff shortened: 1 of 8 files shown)


MarkLark86 (commented):
Looks good. There are a couple tests failing for ProdAPI though

BrianMwangi21:
@MarkLark86 they are failing cause it needed the other PR in core to be merged. I have re-run the CI now and prodapi is ok

MarkLark86 (approved):

PR #2971, opened 2026-09-02
[SDBELGA-1130] - Update Export as Article to use unified resource

BrianMwangi21:
### Purpose
Move Export as Article off the Eve `planning_article_export` resource onto the unified resource and a native async endpoint.

### What has changed
- `planning_article_export` is a native async POST endpoint (same name, URL and `planning` privilege; the client needs no change). The request body is validated by a Pydantic model; unknown `desk` / `article_template` ids return 400 as the Eve schema did.
- Export logic is now module functions (`export_items_to_article`, `export_events_to_text`) shared by the endpoint, the events download endpoint and the scheduled-filters export. No `get_resource_service("planning_article_export")` remains.
- `get_items` resolves related Events/Planning via the `unified.common` helpers; the legacy `planning.utils` Eve reads are gone. This also fixes combined exports rendering Events with no coverages (the old helper searched the now-empty legacy planning index).
- `get_related_planning_for_events` gained `max_results`; the export passes its page size instead of hitting Elasticsearch's default 10-hit cap.
- Templates: unified planning `definition_long` is exposed as `description_text`; planning items linked to an Event keep using the Event's dates for `schedule`.
- Blocking `find_one` calls in the module made async.
- Fixtures for `planning_export.feature` / `combined_export.feature` gained the required `news_coverage_status`; 9 scenarios that died at setup on `feature/unified` now run. New `tests/unified_resource/article_export_test.py`.

### Notes for review
- `test_combined_export_template.html` switched from `item._type == 'events'` to `item.type == 'event'`; customer export templates using `_type` would need the same change.
- Core's `fields.ObjectId` raises bson `InvalidId` rather than a pydantic error on malformed ids; the endpoint maps it to a 400. Probably worth a small core fix.
- The per-item lookups in `get_items` (one fetch per planning, one search per event) are pre-existing and left for a follow-up.

Resolves: [SDBELGA-1130](https://sofab.atlassian.net/browse/SDBELGA-1130)

[SDBELGA-1130]: https://sofab.atlassian.net/browse/SDBELGA-1130

### DIFF
--- a/server/features/combined_export.feature
+++ b/server/features/combined_export.feature
@@ -64,12 +64,14 @@ Feature: Export combined Planning and Event items with default template
             "ednote": "Ed. note 1",
             "coverages": [{
                 "coverage_id": "123",
+                "news_coverage_status": {"qcode": "ncostat:int", "name": "coverage intended", "label": "Planned"},
                 "planning": {
                     "g2_content_type": "text"
                 }
             },
             {
                 "coverage_id": "456",
+                "news_coverage_status": {"qcode": "ncostat:int", "name": "coverage intended", "label": "Planned"},
                 "planning": {
                     "g2_content_type": "photo"
                 }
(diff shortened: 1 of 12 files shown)


MarkLark86 (approved):
Great work

PR #2976, opened 2026-09-08
[SDESK-8014] - Show Event coverage information in all views

BrianMwangi21:
### Purpose
Events on the unified resource can carry `coverages`, but the client only rendered coverages for Planning items. This PR shows Event coverages in every view where Planning coverages already appear: list row, preview, editor, history and article export.

### What has changed
- `IEventItem` gains `coverages`; the Event form profile interface gains the `coverages` editor field, schema entry and `groups`.
- List row: the `coverages` list field is no longer gated on Planning items and is added to the default event second line (`config.ts`). The date/language filtering previously inlined in `PlanningItem` is extracted to `planningUtils.filterCoveragesForList` and used by both `PlanningItem` and `EventItem`. Event rows show all their coverages on every day group; the language filter still applies.
- Preview: new `CoveragesPreview` component extracted from `PlanningPreviewContent` and reused in `EventPreviewContent`, rendered at the profile `coverages` group position (after the other sections when the profile has no such group). `EventPreviewContent` now also fetches planning files for coverage attachments.
- Editor: the existing `coverages` editor field is wired into `EventEditor` (change handler with autosave flush, files, popups, field refs) together with the add-coverage and coverage-links bookmarks, enabled when `profile.editor.coverages.enabled` is set. `validators.event` gains `validateCoverages`; `eventUtils.modifyForClient` / `modifyForServer` normalise coverages via the shared `modifyCoverageForClient` / `modifyCoverageForServer` helpers. `defaultCoverageValues` and `getDefaultCoverageDueDate` accept an Event as the parent item, defaulting the due date from the event end.
- History: `EventHistory` splits item and coverage operations the same way as `PlanningHistory` and renders `CoverageHistory` cards.
- Default Event content profile enables `coverages` in a new `Coverages` group.
- Article export: `get_items` merges an Event's own coverages with its related Planning coverages, deduplicated by `coverage_id`, instead of overwriting them; `generate_text_item` enhances the Event's own coverages too.
- Tests: karma for `EventItem`, `EventPreviewContent`, `EventHistory`, `filterCoveragesForList`, `defaultCoverageValues` and event coverage normalisation; pytest for the export merge; `content_profiles_event.feature` updated for the new default profile.

### How to test
1. Create an Event and add coverages in the Event editor (the Coverages group and bookmarks appear when the field is enabled in the Event profile).
2. Events list / combined list: the event row shows the coverage icons on every day it appears under; switch the language filter and confirm only matching coverages remain.
3. Preview the Event: a Coverages section lists each coverage with the same card as a Planning item.
4. History tab of the Event: coverage cards appear for coverage operations.
5. Export the Event as an article: the body lists the Event's own coverages plus those of linked Planning items, without duplicates.

### Notes for review
- Instances that override `planning.event_list_item.secondLine` in `superdesk.config.js` need to add `{fieldId: 'coverages', position: 'end'}` to see coverages on event rows.

Resolves: [SDESK-8014](https://sofab.atlassian.net/browse/SDESK-8014)

[SDESK-8014]: https://sofab.atlassian.net/browse/SDESK-8014

### DIFF
--- a/client/api/editor/item_events.ts
+++ b/client/api/editor/item_events.ts
@@ -19,7 +19,12 @@ import {planningApi, superdeskApi} from '../../superdeskApi';
 import {generateTempId, isTemporaryId} from '../../utils';
 import {getBookmarksFromFormGroups, getEditorFormGroupsFromProfile} from '../../utils/contentProfiles';
 
-import {AddPlanningBookmark, AssociatedPlanningsBookmark} from '../../components/Editor/bookmarks';
+import {
+    AddCoverageBookmark,
+    AddPlanningBookmark,
+    AssociatedPlanningsBookmark,
+    CoveragesBookmark,
+} from '../../components/Editor/bookmarks';
 import {RelatedPlanningItem} from '../../components/fields/editor/EventRelatedPlannings/RelatedPlanningItem';
 import {convertEventToPlanningItem} from '../../actions';
 import {addRelatedPlannings} from '../../utils/planning';
@@ -44,6 +49,18 @@ export function getEventsInstance(type: EDITOR_TYPE): IEditorAPI['item']['events
         const bookmarks = getBookmarksFromFormGroups(groups);
         let index = bookmarks.length;
 
+        const coverageBookmarks: Array<IEditorBookmark> = profile.editor.coverages?.enabled !== true ? [] : [{
+            id: 'add_coverage',
+            type: BOOKMARK_TYPE.custom,
+            index: index++,
+            component: AddCoverageBookmark,
+        }, {
+            id: 'coverage_links',
+            type: BOOKMARK_TYPE.custom,
+            index: index++,
+            component: CoveragesBookmark,
+        }];
+
         return {
             bookmarks: bookmarks.concat([{
                 id: 'divider-1',
@@ -61,7 +78,7 @@ export function getEventsInstance(type: EDITOR_TYPE): IEditorAPI['item']['events
                 index: index++,
                 disabled: !canCreatePlanningItems,
                 component: AssociatedPlanningsBookmark,
-            }]),
+            }], coverageBookmarks),
             groups: Object.values(groups),
         };
     }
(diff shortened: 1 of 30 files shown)


client/api/editor/item_events.ts
```diff
@@ -61,7 +78,7 @@ export function getEventsInstance(type: EDITOR_TYPE): IEditorAPI['item']['events
                 index: index++,
                 disabled: !canCreatePlanningItems,
                 component: AssociatedPlanningsBookmark,
-            }]),
+            }], coverageBookmarks),
```
MarkLark86: As the Coverage(s) are associated directly with the Event, it might be good to have them before related Planning items (make use of the `BOOKMARK_TYPE.divider` entry to separate it from other bookmarks

client/components/Coverages/CoverageEditor/index.tsx
```diff
@@ -81,18 +81,22 @@ function duplicateCoverage({
...
     const state: IPlanningAppState = planningApi.redux.store.getState();
 
-    // TAG: MULTIPLE_PRIMARY_EVENTS
-    const relatedEventId = getRelatedEventIdsForPlanning(planning, 'primary')[0];
+    let relatedEvent: IEventItem | undefined;
 
-    const relatedEvent: IEventItem | undefined = relatedEventId == null
-        ? undefined
-        : state.events.events[relatedEventId];
+    if (isEvent(planning)) {
```
MarkLark86: The `isEvent` function was around before we used typescript.
Using the type `IEventOrPlanningItem` you should be able to test against `type` and typescript will know which one based on the if/else branch.

client/utils/planning.tsx
```diff
@@ -455,6 +459,30 @@ export function mapCoverageByDate(coverages: Array<IPlanningCoverageItem> = []):
...
+ * Coverages to show on a list row: filtered by the list language filter, and for
+ * ad-hoc planning items by the day group the row is rendered under
+ */
+export function filterCoveragesForList(
+    item: IEventOrPlanningItem,
+    coverages: Array<IPlanningCoverageItem>,
+    options: {date?: string; activeFilter: PLANNING_VIEW; filterLanguage?: string},
+): Array<IPlanningCoverageItem> {
+    const {date, activeFilter, filterLanguage} = options;
+    const showAllDates = isEvent(item) || (
```
MarkLark86: I think we can remove the `isEvent` clause here, as we want the combined view to stay the same but the Events view should now take the Coverage dates as to on what day to display the item/Coverage

MarkLark86 (approved):

PR #2977, opened 2026-09-09
[SDBELGA-1126] fix data_update and tests

MarkLark86:
### What has changed
Fix lint issues, commented out command and fixes profiles state tests

Resolves: SDBELGA-1126
Depends on: https://github.com/superdesk/superdesk-core/pull/3315

### DIFF
--- a/server/planning/content_profiles/module.py
+++ b/server/planning/content_profiles/module.py
@@ -1,7 +1,13 @@
 from quart_babel import lazy_gettext
 
 import superdesk
-from superdesk.core.resources import ResourceConfig, RestEndpointConfig, MongoResourceConfig, MongoIndexOptions
+from superdesk.core.resources import (
+    ResourceConfig,
+    UpdateStrategy,
+    RestEndpointConfig,
+    MongoResourceConfig,
+    MongoIndexOptions,
+)
 from superdesk.core.auth.privilege_rules import http_method_privilege_based_rules
 from superdesk.factory.app import SuperdeskApp
 
@@ -30,6 +36,7 @@ planning_types_resource_config = ResourceConfig(
     name="planning_types",
     data_class=PlanningProfileResource,
     service=PlanningTypesAsyncService,
+    update_strategy=UpdateStrategy.DEEP_MERGE,
     mongo=MongoResourceConfig(
         indexes=[
             MongoIndexOptions(
--- a/server/planning/types/unified/model.py
+++ b/server/planning/types/unified/model.py
@@ -159,7 +159,7 @@ class UnifiedPlanningResource(
 
         return self
 
-    def clone_with(self, updates: dict[str, Any]) -> "UnifiedPlanningResource":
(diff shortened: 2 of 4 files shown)


server/planning/types/unified/model.py
```diff
@@ -159,7 +159,7 @@ def validate_model_instance(self) -> "UnifiedPlanningResource":
 
         return self
 
-    def clone_with(self, updates: dict[str, Any]) -> "UnifiedPlanningResource":
+    def clone_with(self, updates: dict[str, Any], **kwargs) -> "UnifiedPlanningResource":
```
eos87, line 162: Not blocking for this PR since it comes from core #3315 rather than anything here, but I think the `clone_with` change affects more than the signature. `unified_planning` doesn't set `update_strategy`, so it now falls back to `SHALLOW_MERGE`. `on_update` (unified/service.py:220) relies on `original.clone_with(updates)` to turn a partial PATCH into the full doc, so a PATCH with a coverage that only has `coverage_id` + `planning` + `assigned_to` now fails with `news_coverage_status: Field is required`. It used to fill that in from the original coverage.

That seems to match 7 behave failures here that weren't on the last `feature/unified` run (that one ran before core #3315 merged), all with the same error:

- `planning.feature:416`, `:1328`, `:1529`
- `assignments_delete.feature:247`, `:305`, `:1152`
- `content_rewrite.feature:2523`

Could you please verify it? Maybe worth a separate ticket.

server/planning/content_profiles/module.py
```diff
@@ -30,6 +36,7 @@ def init_app(app: SuperdeskApp) -> None:
     name="planning_types",
     data_class=PlanningProfileResource,
     service=PlanningTypesAsyncService,
+    update_strategy=UpdateStrategy.DEEP_MERGE,
```
eos87, line 39: Could you double-check how this behaves with list fields in the profiles? I think `DEEP_MERGE` might not match what Eve's `merge_nested_documents` used to do here.

From what I can see, Eve's `resolve_nested_documents` only merged dicts and replaced lists as a whole. Core's `merge_dicts_deep` merges lists by index instead. When the new list is shorter, it looks like it keeps the extra items from the original. The client sends the full array when it changes (json-merge-patch), so if I'm reading it right, removing items from a list in a profile wouldn't stick.

I tried `merge_dicts_deep` locally with a couple of inputs:

- `["en","fr","de"]` merged with `["en","fr"]` gives `["en","fr","de"]`
- `["a","b","c"]` merged with `["c","a"]` gives `["c","a","c"]`

So removing a language from the `language` field, or a format option or vocabulary from another field, might not get saved. I haven't tested it end to end through the UI, so I could be missing something upstream. Is this something you've already run into, or worth a quick check?

eos87 (approved):
Looks good to me. Comments below were AI findings, I tried to verify them and they seem worth taking a look.

MarkLark86:
Thanks @eos87 I fixed the deep merge in core (https://github.com/superdesk/superdesk-core/pull/3323), and set the UnifiedResource to use deep merge

PR #2979, opened 2026-09-09
reimplement belga quick coverage creation UI

petrjasek:
there is a server setting PLANNING_ADVANCED_COVERAGE_MODE to enable it

SDESK-7501

### Front-end checklist

Resolves: #[issue-number]

### DIFF
--- a/client/actions/users.ts
+++ b/client/actions/users.ts
@@ -145,10 +145,10 @@ const self = {
     fetchAndRegisterUserPreferences,
     receiveUserPreferences,
     setCoverageDefaultDesk,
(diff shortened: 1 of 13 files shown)


eos87:
Superseded by #2983. It cherry-picks the two commits from here and builds on them: the config is now PLANNING_INLINE_COVERAGE_FORM, the form only shows on the event's own coverages field and starts collapsed behind an "Add Coverages" box, and the row logic is shared with the advanced modal. superdesk-belga feature/unified already points at the new branch. Closing this one.

PR #2983, opened 2026-09-15
[SDESK-7501] Add inline coverages form to the event form behind a config

eos87:
### Purpose
Adds an inline "Coverage Types" form to the event form, so coverages can be added in one go without opening a modal per coverage. It is a rework of the form that existed on the related planning rows before STT-63 introduced the embedded planning editor, adapted to events carrying their own coverages. 

Supersedes #2979; Petr's two commits are cherry-picked here and the rest is built on top.

Off by default, turned on per instance in the Event content profile: Manage Event Profile > Coverages field > "Inline Add Coverages form" (`editor.coverages.inline_form`). Planning editors, including embedded under Related Plannings, are not affected.

### What has changed
- New `CoverageAddInlineForm`: one line per content type, tick to enable, then desk, user, language and status. Rendered by `CoverageArrayInput` when the event profile option is on and the item is an event. It starts collapsed as an "Add Coverages" box in place of the add button and the "No Coverages Yet" drop zone; existing coverages keep their collapsible editors above it.
- The row logic (create rows from content types, desk and user changes, duplicate, validation, building the coverages) is extracted from the advanced coverages modal into `coverageRows.ts` and shared by the modal and the inline form, with unit tests. The modal now shows "Desk is required" on a row that has a user but no desk, instead of letting it through.
- Rules applied by the inline form to newly added rows only, the advanced modal is not affected: status goes to Planned when a desk is picked unless `PLANNING_MANUAL_NEWS_COVERAGE_STATUS` is on; a desk is required when `PLANNING_AUTO_ASSIGN_TO_WORKFLOW` is on; slugline, headline, internal note and ednote are copied from the event's translations for the coverage language.
- The server config from #2979 (`PLANNING_ADVANCED_COVERAGE_MODE`) is gone. The toggle is a per-field option on the event profile, editable in the profile editor and merged into existing profiles as off.
- The dead `EmbeddedCoverageForm` and `CoverageRowForm` components and their stylesheet are removed.
- e2e: `events/inline_coverage_form.spec.ts` covers adding two coverage types, saving, reloading and checking desk and status, the fallback to the add button when the option is off, and turning the option on through the profile editor. The option is set per test through a `planning_types` fixture.

### Steps to test
- Open Manage Event Profile from the planning Actions menu, select the Coverages field, switch on "Inline Add Coverages form", Apply, then Save All.
- Open an event. The Coverages section shows an "Add Coverages" box instead of the add button, click it to open the form. Tick Text and Picture, pick a desk for Text, click Add Coverage(s). Two coverage editors appear and the form collapses again. Text is Planned, Picture keeps the default status.
- Save, close, reopen: the coverages, desk and status are kept.
- Open a planning item, or expand an embedded planning under Related Plannings: the add button is still there, no inline form.
- With the option off the event form is unchanged.
- The advanced coverages modal (add button, Advanced mode) still works; leaving a user without a desk now shows the error on the row.

### Screenshots
image

image

Resolves: [SDESK-7501]

[SDESK-7501]: https://sofab.atlassian.net/browse/SDESK-7501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eos87:
Hi @MarkLark86, this is still work in progress but I would appreciate if you could already take a look. I will wrap this up tomorrow and address any comments you may have. Thanks already!

client/components/Coverages/CoverageAddInlineForm.scss
```diff
@@ -0,0 +1,44 @@
+.coverage-inline-form {
+    border: 1px solid var(--sd-colour-line--light);
+    border-radius: 2px;
```
MarkLark86: Should we also use a var here: `var(--radius-xs)`

client/components/Coverages/CoverageAddInlineForm.tsx
```diff
@@ -0,0 +1,240 @@
...
+}
+
+class CoverageAddInlineFormComponent extends React.Component<IProps, IState> {
+    contentTypes: Map<string, IG2ContentType>;
+    form: React.RefObject<HTMLDivElement>;
+
+    constructor(props: IProps) {
+        super(props);
+
+        this.contentTypes = new Map(props.contentTypes.map((contentType) => [
```
MarkLark86: We could probably use momoization to get this map, and use it in the render function.
I know it won't happen often, but if the G2ContentTypes CV are changed, this would require a reload of the component

client/components/Coverages/style.scss
```diff
@@ -10,3 +10,11 @@
     margin-top: -1.5rem;
   }
 }
+
+.coverage-editable-fields--stacked {
+    display: flex;
+    flex-direction: column;
+    gap: 0.8rem;
```
MarkLark86: It would be good to use CSS variables here, for cases where the base sizing element changes

README.md
```diff
@@ -202,6 +202,12 @@ Below sections include the config options that can be defined in settings.py.
     * If False, a Coverage's `news_coverage_status` will be set to Planned (`ncostat:int`) when a `Desk` is assigned to it,
       or the Coverage is added to workflow.
     * If True, a Coverage's `news_coverage_status` will only be changed when a user changes the value.
+* PLANNING_INLINE_COVERAGE_FORM
```
MarkLark86: Thinking about this config, not sure if the `settings.py` backend config is the right place for this.
As this defines the interface to use for the `Coverages` field, to me an ideal place for this would be on the Event or Planning profile.

Do you think it would be worth exposing this config through the profiles, which allows an admin to turn this interface on/off.

MarkLark86 (commented):
Looks good, only a few minor things I noticed

README.md
eos87: yes, I thought about this. I think I will do it.

eos87: done in c1114e79c8edda50bd8b5df1a37bdded2291db5f

MarkLark86 (approved):

PR #2987, opened 2026-09-17
implement command migrate_unified_storage

petrjasek:
besides the command it contains the changes to event_contact_info mapping so it supports external contacts

SDBELGA-1137

### Front-end checklist

Resolves: #[issue-number]


server/planning/commands/migrate_unified_storage.py
```diff
@@ -0,0 +1,455 @@
...
+        while True:
+            lookup = {} if last_id is None else {"_id": {"$gt": last_id}}
+            batch = await collection.find(lookup).sort("_id").limit(self.batch_size).to_list(length=self.batch_size)
+            if not batch:
+                break
+
+            yield batch
+            last_id = batch[-1]["_id"]
+
+    async def get_existing_ids(self, collection: AsyncIOMotorCollection, ids: list[Any]) -> set[Any]:
```
MarkLark86, line 432: It might be good to use `set[str]` here instead, so the response can be type checked elsewhere in the code.

server/planning/commands/migrate_unified_storage.py
```diff
@@ -0,0 +1,455 @@
...
+        cursor = collection.find({"_id": {"$in": ids}}, {"_id": 1})
+        return {doc["_id"] async for doc in cursor}
+
+    async def find_duplicate_ids(self, source_a: str, source_b: str) -> list[Any]:
+        duplicates: list[Any] = []
+        async for batch in self.iter_batches(self.db[source_a]):
+            ids = [doc["_id"] for doc in batch]
+            duplicates.extend(await self.get_existing_ids(self.db[source_b], ids))
```
MarkLark86, line 440: Rather than converting `get_existing_ids` here to a list, what if we use a `set[str]` instead,.
i.e. `duplicates |= await self.get_existing_ids(self.db[source_b], ids))`

server/planning/commands/migrate_unified_storage.py
```diff
@@ -0,0 +1,455 @@
...
+
+            for doc in batch:
+                if doc["_id"] in existing_ids and not self.force:
+                    stats.skipped += 1
+                    continue
+
+                original = doc
+                try:
+                    # ``transform`` mutates the document, so keep the original around to print on failure
+                    original = deepcopy(doc) if self.debug else doc
```
MarkLark86: Maybe this one can replace the line before this `try..except` block. Deep copying shouldn't be an issue, and no need to re-assign a variable that was just constructed

server/planning/commands/migrate_unified_storage.py
```diff
@@ -0,0 +1,455 @@
...
+        for calendar in doc.get("calendars") or []:
+            if not isinstance(calendar.get("translations"), dict):
+                calendar.pop("translations", None)
+
+    @staticmethod
+    def clean_subject_translations(doc: dict[str, Any]) -> None:
+        for subject in doc.get("subject") or []:
+            translations = subject.get("translations")
+            if not isinstance(translations, dict):
+                continue
```
MarkLark86, line 299: Should we pop the `translations` key if it's value is not a `dict`, same as what's done with calendars?

server/planning/commands/migrate_unified_storage.py
```diff
@@ -0,0 +1,455 @@
...
+        print(f"Migrating 'event_autosave' into '{AUTOSAVE_COLLECTION}'")
+        stats = self.stats.setdefault(PHASE_AUTOSAVE, PhaseStats())
+        target = self.db[AUTOSAVE_COLLECTION]
+
+        stats.updated += await self.set_missing_field(target, "type", PlanningItemType.PLANNING.value)
+        await self.copy_documents(
+            source=self.db["event_autosave"],
+            target=target,
+            stats=stats,
+            transform=lambda doc: {**doc, "type": doc.get("type") or PlanningItemType.EVENT.value},
```
MarkLark86: Because all the items are coming from `event_autosave` are Events, we can safely force the type to `event` here rather than rely on what might be there now

server/planning/commands/migrate_unified_storage.py
```diff
@@ -0,0 +1,455 @@
...
+            source=self.db["events_history"],
+            target=target,
+            stats=stats,
+            transform=self.normalise_event_history,
+        )
+
+        print(f"  {stats}")
+        return stats
+
+    async def migrate_autosave(self) -> PhaseStats:
```
MarkLark86: This actually raises a question, are we better off getting users to close all Events & Planning, so we start with a fresh autosave (reducing risks against data integrity - otherwise we might have to run the `clean_calendar_translations` etc on autosave items too).

MarkLark86 (approved):
Looks good, just a couple minor things.

server/planning/commands/migrate_unified_storage.py
petrjasek: yep I think asking them to save everything before the update shouldn't be a problem

server/planning/commands/migrate_unified_storage.py
petrjasek: it can be both string and ObjectId for history records

